PR #8, opened 2017-11-07
Initial DASH Support

gesinger:
## Description
Initial DASH support with DRM support via videojs-contrib-eme

Will update the documentation, but wanted to put the PR up for early review.

## Requirements Checklist
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](http://jsbin.com/liwecukasi/edit?html,output))


forbesjo:
It would be nice to have `videojs-contrib-eme` as a dev dependency and included in the test pages

src/videojs-http-streaming.js
```diff
@@ -119,6 +135,26 @@ Hls.canPlaySource = function() {
                           'your player\'s techOrder.');
 };
 
+const emeOptions = (options, videoPlaylist, audioPlaylist) => {
+  if (!options.keySystems) {
+    return options;
+  }
+
+  // upsert the content types based on the selected playlist
```
mjneil: upsert

src/videojs-http-streaming.js
```diff
@@ -418,6 +455,18 @@ class HlsHandler extends Component {
     this.masterPlaylistController_.on('selectedinitialmedia', () => {
       // Add the manual rendition mix-in to HlsHandler
       renditionSelectionMixin(this);
+
+      if (this.options_.sourceType === 'dash') {
```
mjneil: It'd be nice to put this in like a `this.setupEmeOptions_()`

src/master-playlist-controller.js
```diff
@@ -1134,16 +1143,30 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+  configureLoaderMimeTypes_(mimeTypes) {
+    // If the content is demuxed, we can't start appending segments to a source buffer
+    // until both source buffers are set up, or else the browser may not let us add the
+    // second source buffer (it will assume we are playing either audio only or video
+    // only).
+    const sourceBufferEmitter =
+      // if the first mime type has muxed video and audio then we shouldn't wait on the
+      // second source buffer
+      mimeTypes.length > 1 && mimeTypes[0].indexOf(',') === -1 ?
+        new videojs.EventTarget() : null;
```
mjneil: we might want to think about expanding this emitter concept to its own controller class. When we start tearing apart flash and pulling the muxer into vhs and removing contrib-media-sources, we will still need a place for some of the contrib-media-sources logic to live e.g. buffer management logic that was added to address the edge/ie corruption issues

src/dash-playlist-loader.js
```diff
@@ -0,0 +1,173 @@
...
+    if (this.request) {
+      const oldRequest = this.request;
+
+      this.request = null;
+      oldRequest.onreadystatechange = null;
+      oldRequest.abort();
+    }
+  }
+
+  media(playlist) {
```
OshinKaramian: Would it be possible to get function comments here?  Mostly I'm asking as this has a few pieces of nuance (the playlist can be a string or object, and thus will be handled differently, as well as the fact that it is both a getter and a setter).

src/dash-playlist-loader.js
```diff
@@ -0,0 +1,173 @@
...
+    if (typeof playlist === 'string') {
+      if (!this.master.playlists[playlist]) {
+        throw new Error('Unknown playlist URI: ' + playlist);
+      }
+      playlist = this.master.playlists[playlist];
+    }
+
+    const mediaChange = !this.media_ || playlist.uri !== this.media_.uri;
+
+    this.state = 'HAVE_METADATA';
```
OshinKaramian: Are these states doc'ed anywhere? Might be nice to have them doc'ed in some form, as it might make debugging easier (particularly if there is a sequence to these events).

mjneil: these states mimic the PlaylistLoader states
https://github.com/videojs/http-streaming/blob/master/docs/arch.md

src/dash-playlist-loader.js
```diff
@@ -0,0 +1,173 @@
...
+            });
+        }
+      }
+
+      // set up phony URIs for the playlists since we won't have external URIs for DASH
+      // but reference playlists by their URI throughout the project
+      for (let i = 0; i < this.master.playlists.length; i++) {
+        const phonyUri = `placeholder-uri-${i}`;
+
+        this.master.playlists[i].uri = phonyUri;
```
mjneil: should also add `resolvedUri` here, quite a few places outside of this class use `resolvedUri` instead of `uri`

mjneil: maybe not quite a few since media-groups has source type checking now, but there is a spot in MPC and external plugins may look for it

src/dash-playlist-loader.js
OshinKaramian: Ah nice, this is what I was looking for.

src/master-playlist-controller.js
gesinger: I agree that we may end up needing that, but for now I think it's best to hold off. We may not get to it for a while, and if we do get to it sooner, it may end up having different requirements than we suppose right now.

src/dash-playlist-loader.js
gesinger: I forgot, it actually will get added in `setupMediaPlaylists`

src/sync-controller.js
```diff
@@ -516,7 +516,7 @@ export default class SyncController extends videojs.EventTarget {
         time: segment.start,
         accuracy: 0
       };
-    } else if (playlist.discontinuityStarts.length) {
+    } else if (playlist.discontinuityStarts && playlist.discontinuityStarts.length) {
```
mjneil: This will probably depend on how we plan to handle multi-period dash, and this extra check isn't problematic in anyway, but m3u8-parser always attaches an empty array for `discontinuityStarts`. If we plan to treat multi-period dash as discontinuities, then this check may be unnecessary, but still ok to be there

src/videojs-http-streaming.js
```diff
@@ -566,13 +640,21 @@ HlsSourceHandler.canPlayType = function(type, options) {
     return false;
   }
 
-  let mpegurlRE = /^(audio|video|application)\/(x-|vnd\.apple\.)?mpegurl/i;
+  const sourceType = simpleTypeFromSourceType(type);
+
+  if (sourceType === 'dash') {
+    if (!options.hls.overrideNative && Hls.supportsNativeDash) {
```
forbesjo: I think we should always override native DASH playback

src/videojs-http-streaming.js
```diff
@@ -190,6 +255,13 @@ class HlsHandler extends Component {
           }
         });
       }
+
+      // Set up a reference to the HlsHandler from player.vhs. This allows users to start
+      // migrating from player.tech_.hls... to player.vhs... for API access. Although this
+      // isn't the most appropriate form of reference for video.js (since all APIs should
+      // be provided through core video.js), it is a common pattern for plugins, and vhs
+      // will act accordingly.
+      _player.vhs = this;
```
forbesjo: We need this to map to deprecated `player.dash` and `player.hls`.

src/sync-controller.js
gesinger: Good call. I think we can just leave it in until we handle multi-period dash in the mpd-parser (so we know for sure).

src/videojs-http-streaming.js
gesinger: Did we have a `player.hls`? I though just `player.tech_.hls`

forbesjo: Uses could access `player.hls.representations`

gesinger: I forgot that it's already available (albeit deprecated). Since it still exists they should be able to continue to use it without issue.

package.json
```diff
@@ -134,6 +135,7 @@
     "shelljs": "^0.5.3",
     "sinon": "1.10.2",
     "uglify-js": "^2.5.0",
+    "videojs-contrib-eme": "http://github.com/gesinger/videojs-contrib-eme#options-rework",
```
forbesjo: Reminder that this needs to point to a version of contrib-eme with necessary changes.

package.json
```diff
@@ -92,6 +92,7 @@
     "global": "^4.3.0",
     "m3u8-parser": "2.1.0",
     "mux.js": "4.3.2",
+    "mpd-parser": "0.0.1",
```
forbesjo: mpd-parser is now on 0.2.0

package.json
forbesjo: Remember to point to a contrib-eme version with the necessary changes

forbesjo:
Looks good to me

README.md
```diff
@@ -1,12 +1,18 @@
...
+Play HLS, DASH, and future HTTP streaming protocols with video.js, even where they're not
+natively supported.
 
-Play back HLS with video.js, even where it's not natively supported.
-
-Lead Maintainer: Jon-Carlos Rivera [@imbcmdth](https://github.com/imbcmdth)
+Lead Maintainers:
+- Jon-Carlos Rivera [@imbcmdth](https://github.com/imbcmdth)
+- Joe Forbes [@forbesjo](https://github.com/forbesjo)
+- Matthew Neil [@mjneil](https://github.com/mjneil)
```
mjneil: nice

src/master-playlist-controller.js
```diff
@@ -105,6 +106,29 @@ const getCodecs = function(media) {
   return defaultCodecs;
 };
 
+const audioProfileFromDefault = (master, audioGroupId) => {
```
mjneil: Maybe this would be a good opportunity to move this and other codec related helper functions into `util/codecs.js`?

src/videojs-http-streaming.js
```diff
@@ -119,6 +135,46 @@ Hls.canPlaySource = function() {
...
+      keySystemContentTypes[keySystem].url = keySystemOptions[keySystem];
+    }
+  }
+
+  return {
+    keySystems: videojs.mergeOptions(keySystemOptions, keySystemContentTypes)
+  };
+};
+
+const setupEmeOptions = (hlsHandler) => {
```
mjneil: probably would be better to use `vhsHandler` naming

src/videojs-http-streaming.js
```diff
@@ -546,7 +622,7 @@ const HlsSourceHandler = function(mode) {
       tech.hls = new HlsHandler(source, tech, localOptions);
```
mjneil: Do we want to make this be `tech.vhs` and point `tech.hls = tech.vhs` for backwards comp?

src/videojs-http-streaming.js
mjneil: just realized, are we not making name changes for `HlsHandler` my hls->vhs name change suggestions probably shouldn't be done then

utils/manifest/dash.mpd
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<MPD xmlns="urn:mpeg:dash:schema:mpd:2011" profiles="urn:mpeg:dash:profile:full:2011" minBufferTime="1.5" mediaPresentationDuration="PT4S">
+  <Period>
+    <BaseURL>main/</BaseURL>
+    <AdaptationSet mimeType="video/mp4">
```
mjneil: Should we add an AdaptionSet for vtt since mpd-parser 0.2.0 supports vtt now?

forbesjo:
Going to merge this and open the suggested comments in a new PR.

PR #10, opened 2017-12-05
feat: extend stats object

forbesjo:
## Description
This adds a few more properties to the `hls.stats` object as well as uses the new `debug` logging level.

### TODO
- [x] Add more stats
- [x] Add and use debug logging from video.js

## Requirements Checklist

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -12,6 +12,7 @@ import { initSegmentId } from './bin-utils';
 import {mediaSegmentRequest, REQUEST_ERRORS} from './media-segment-request';
 import { TIME_FUDGE_FACTOR, timeUntilRebuffer as timeUntilRebuffer_ } from './ranges';
 import { minRebufferMaxBandwidthSelector } from './playlist-selectors';
+import logger from './util/logger';
 
 // in ms
 const CHECK_BUFFER_DELAY = 500;
@@ -177,9 +178,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     // ...for determining the fetch location
     this.fetchAtBuffer_ = false;
 
-    if (options.debug) {
-      this.logger_ = videojs.log.bind(videojs, 'segment-loader', this.loaderType_, '->');
-    }
+    this.logger_ = logger(`SegmentLoader[${this.loaderType_}]`);
   }
 
   /**
@@ -672,27 +671,17 @@ export default class SegmentLoader extends videojs.EventTarget {
       return null;
     }
 
-    this.logger_('checkBuffer_',
-      'mediaIndex:', mediaIndex,
-      'hasPlayed:', hasPlayed,
-      'currentTime:', currentTime,
-      'syncPoint:', syncPoint,
-      'fetchAtBuffer:', this.fetchAtBuffer_,
-      'bufferedTime:', bufferedTime);
-
     // When the syncPoint is null, there is no way of determining a good
     // conservative segment index to fetch from
     // The best thing to do here is to get the kind of sync-point data by
     // making a request
     if (syncPoint === null) {
       mediaIndex = this.getSyncSegmentCandidate_(playlist);
-      this.logger_('getSync', 'mediaIndex:', mediaIndex);
       return this.generateSegmentInfo_(playlist, mediaIndex, null, true);
     }
 
     // Under normal playback conditions fetching is a simple walk forward
     if (mediaIndex !== null) {
-      this.logger_('walkForward', 'mediaIndex:', mediaIndex + 1);
       let segment = playlist.segments[mediaIndex];
 
       if (segment && segment.end) {
@@ -725,9 +714,6 @@ export default class SegmentLoader extends videojs.EventTarget {
       mediaIndex = mediaSourceInfo.mediaIndex;
       startOfSegment = mediaSourceInfo.startTime;
     }
-    this.logger_('getMediaIndexForTime',
-      'mediaIndex:', mediaIndex,
-      'startOfSegment:', startOfSegment);
 
     return this.generateSegmentInfo_(playlist, mediaIndex, startOfSegment, false);
   }
@@ -1273,14 +1259,6 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.throughput.rate +=
       (segmentProcessingThroughput - rate) / (++this.throughput.count);
   }
-
-  /**
-   * A debugging logger noop that is set to console.log only if debugging
-   * is enabled globally
-   *
-   * @private
-   */
-  logger_() {}
 
   /**
    * Adds a cue to the segment-metadata track with some metadata information about the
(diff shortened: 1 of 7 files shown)


mjneil:
Do we want this as a plugin on the player? or should it be apart of the `player.vhs` api?

src/videojs-http-streaming.js
```diff
@@ -600,10 +600,46 @@ if (!videojs.use) {
...
+    hlsStats: player.tech_.hls.stats,
+    duration: player.duration(),
+    buffered: getTimeRanges(player, 'buffered'),
+    seekable: getTimeRanges(player, 'seekable'),
+    currentTime: player.currentTime(),
+    timestamp: Date.now(),
+    currentSource: player.currentSource(),
+    master: player.tech_.hls.masterPlaylistController_.masterPlaylistLoader_.master,
+    currentTech: player.tech_.name_,
+    // TODO: filter by VHS logging only
```
mjneil: How do we want to go about this? Currently videojs doesn't do any type of filtering or bucketing of the history, just a single array that all logging args get dumped into.

Some options I can think of:
* prefix all log messages we add with a VHS specific string so we can filter on that
* send only objects into the logs e.g. `{ vhs: true, message: 'log message' }` Downside to this is if the debug log is enabled, videojs will console log objects for all of our messages instead of just the message
* extend videojs.log to seperate history by type or allow a way of requesting logs in history of a specific type. Downside to this would be not being available on previous versions of videojs

forbesjo:
Then again we have not yet exposed `player.vhs` yet. Those changes are in #8

src/videojs-http-streaming.js
```diff
@@ -600,10 +600,46 @@ if (!videojs.use) {
...
+    videoPlaybackQuality: player.getVideoPlaybackQuality(),
+    playerDimensions: player.currentDimensions(),
+    hlsStats: player.tech_.hls.stats,
+    duration: player.duration(),
+    buffered: getTimeRanges(player, 'buffered'),
+    seekable: getTimeRanges(player, 'seekable'),
+    currentTime: player.currentTime(),
+    timestamp: Date.now(),
+    currentSource: player.currentSource(),
+    master: player.tech_.hls.masterPlaylistController_.masterPlaylistLoader_.master,
```
mjneil: could use the shortcut `player.tech_.hls.playlists.master`

src/videojs-http-streaming.js
```diff
@@ -600,10 +600,46 @@ if (!videojs.use) {
...
+  return timeRangesList;
+};
+
+const getPlaybackStats = function() {
+  const player = this;
+
+  return {
+    videoPlaybackQuality: player.getVideoPlaybackQuality(),
+    playerDimensions: player.currentDimensions(),
+    hlsStats: player.tech_.hls.stats,
```
mjneil: do we want to call this `hlsStats` since the stats apply to dash sources as well?

forbesjo: This should probably be renamed `vhsStats` since it would be the `vhs.stats` object

src/util/logger.js
```diff
@@ -0,0 +1,11 @@
+import videojs from 'video.js';
+
+const logger = (source) => {
+  if (videojs.log.debug) {
+    return videojs.log.debug.bind(videojs, '[VHS]', `${source} >`);
+  }
+
+  return function() {};
```
mjneil: Do we want to fall back to `videojs.log` when `videojs.log.debug` does not exist?

forbesjo: That would then print out the logging by default

mjneil: ah ya we don't want that

src/segment-loader.js
```diff
@@ -672,27 +671,17 @@ export default class SegmentLoader extends videojs.EventTarget {
       return null;
     }
 
-    this.logger_('checkBuffer_',
```
gesinger: Any specific reason for removing this and other logs that currently exist?

src/videojs-http-streaming.js
```diff
@@ -509,6 +509,36 @@ class HlsHandler extends Component {
...
+        });
+      }
+
+      return timeRangesList;
+    };
+
+    return {
+      videoPlaybackQuality: this.tech_.getVideoPlaybackQuality(),
+      playerDimensions: this.tech_.currentDimensions(),
+      vhsStats: this.stats,
```
gesinger: This seems a bit ambiguous in naming considering we have `vhsStats` as a sub property of `playbackStats`

src/segment-loader.js
forbesjo: @mjneil was removing some excessive logging

src/videojs-http-streaming.js
forbesjo: Do we want to merge with `stats` or perhaps turn `stats` into a function that has all this information?

src/util/logger.js
```diff
@@ -0,0 +1,11 @@
+import videojs from 'video.js';
+
+const logger = (source) => {
```
mjneil: This will not allow us the ability to turn debug logging on mid session via `videojs.log.level('debug')` . If we want to support that, we will need something like this

```
const logger = (source) => (...args) => {
  if (videojs.log.level() === 'debug') {
    videojs.log.debug.call(videojs, 'VHS:', `${source} >`, ...args);
  }
};
```

gkatsev: you don't even need the `.call(videojs,` in your way @mjneil.

mjneil: oh right, good call!

mjneil: oops this is checking if debug api is available, not if the mode is debug

src/util/logger.js
```diff
@@ -0,0 +1,11 @@
...
+  if (videojs.log.debug) {
```
gesinger: I know we discussed how the history object might get crowded by the debug logging once it's available. Do we know about how many logs per minute of video we'll be generating?

src/videojs-http-streaming.js
```diff
@@ -377,6 +377,20 @@ class HlsHandler extends Component {
       }
     });
 
+    const getTimeRanges = (player, type) => {
```
gesinger: Since player and type are only used once to get the time ranges, it might be simpler and clearer to only pass in the time ranges themselves. The function may also be better served as being called `timeRangesToArray` and if we put it outside of the function we can add some tests for it.

src/util/logger.js
forbesjo: Looks like around ~35 items per minute with what's being logged at the moment

gesinger: If that's pretty standard (and we can't get into a case where the log messages explode), that seems pretty good (to not have to hide behind another config flag). We should be careful to look at some of the cases in case any could spew messages though, and test some scenarios to make sure we don't kill a browser tab from memory consumptions.

PR #19, opened 2018-01-25
test: use browserstack for running tests

forbesjo:
## Description
Runs tests on http://browserstack.com when available.

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -2,6 +2,16 @@ var istanbul = require('browserify-istanbul');
 var isparta = require('isparta');
 
 module.exports = function(config) {
+  // build out a name for browserstack
+  // {TRAVIS_BUILD_NUMBER} [{TRAVIS_PULL_REQUEST} {PR_BRANCH}] {TRAVIS_BRANCH}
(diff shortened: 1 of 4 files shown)


gkatsev:
We should add a message about browserstack sponsoring like in the [Video.js README](https://github.com/videojs/video.js#contributing)

test/karma.conf.js
```diff
@@ -25,19 +30,50 @@ module.exports = function(config) {
       'dist-test/webpack-test.js',
       'dist-test/videojs-http-streaming.js'
     ],
+    browserStack: {
+      project: 'videojs-http-streaming',
+      name: 'videojs-http-streaming' + process.env.TRAVIS_BUILD_NUMBER,
```
gkatsev: I think it would be more useful to have `TRAVIS_BUILD_NUMBER + ' ' + TRAVIS_BRANCH`, testing it [here](https://github.com/videojs/video.js/pull/4903/files) currently.
We should also add in `build` so it's easier to quickly see what build is what. The project name is already easily visible so, not sure it's necessary here.

PR #21, opened 2018-01-25
Speed up Travis builds with containers

forbesjo:
The purpose of this PR is to see how the tests react to being run in a container instead of a VM. If only the playback test is having issues we can only have those tests run in browserstack (see #19)


forbesjo:
Going to use the result of @squarebracket 's investigation in this PR https://github.com/videojs/videojs-contrib-dash/pull/229

test/karma.conf.js
```diff
@@ -30,14 +20,26 @@ module.exports = function(config) {
...
-       }
+        flags: ['--no-sandbox', '--no-user-gesture-required']
+      }
+    },
+    detectBrowsers: {
+      usePhantomJS: false,
+      postDetection: function(browsers) {
+        var newBrowsers = [];
+
+        if (browsers.indexOf('Chrome') !== -1) {
```
squarebracket: Not sure if I'm the only one, but I was confused at first what this was supposed to do. Maybe a comment saying "use headless mode for detected chrome and firefox"?

test/karma.conf.js
```diff
@@ -30,14 +20,26 @@ module.exports = function(config) {
...
-      ChromeHeadlessNoSandbox: {
+      ChromeHeadlessDisableAutoplayPolicy: {
         base: 'ChromeHeadless',
-        flags: ['--no-sandbox', '--autoplay-policy=no-user-gesture-required']
-      },
-      ChromeCanaryHeadlessNoSandbox: {
-        base: 'ChromeCanaryHeadless',
-        flags: ['--no-sandbox', '--autoplay-policy=no-user-gesture-required'],
-       }
+        flags: ['--no-sandbox', '--no-user-gesture-required']
```
squarebracket: there's a `--mute-audio` flag which might be nice here. less annoying bip bops :)

test/karma.conf.js
forbesjo: I'll add a comment communicating the intent. The idea is to check if you have Chrome installed, if so then we'll use the custom browser with all the flags

PR #24, opened 2018-01-29
Don't install container deps for browserstack test runs

squarebracket:
## Description
Chrome and Firefox have certain dependencies that need to be installed when running in Travis containers. Those dependencies are not needed when testing on Browserstack.


squarebracket:
Variable is always exported :(

PR #26, opened 2018-02-05
Update babel, browserify, and other build steps

dmlap:
## Description
Start using node v8 for builds. Upgrade to the latest stable babel release and bring along browserify, babelify, and webworkify. Playback works on the test page and the tests all pass.

## Requirements Checklist
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](http://jsbin.com/liwecukasi/edit?html,output))

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,11 +91,12 @@
     "mux.js": "4.3.2",
     "url-toolkit": "^2.1.3",
     "video.js": "^6.2.0",
-    "webworkify": "1.0.2"
+    "webworkify": "^1.5.0"
   },
   "devDependencies": {
-    "babel": "^5.8.0",
-    "babelify": "^6.0.0",
+    "babel-cli": "^6.26.0",
+    "babel-preset-env": "^1.6.1",
+    "babelify": "^8.0.0",
     "bannerize": "^1.0.0",
     "browserify": "^11.0.0",
     "browserify-istanbul": "^2.0.0",
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,7 @@
+{
+  "presets": [["env", {
+    "targets": {
+      "browsers": ["last 2 versions", "ie 9-11"]
+    }
+  }]]
+}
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -91,11 +91,12 @@
...
     "url-toolkit": "^2.1.3",
     "video.js": "^6.2.0",
-    "webworkify": "1.0.2"
+    "webworkify": "^1.5.0"
   },
   "devDependencies": {
-    "babel": "^5.8.0",
-    "babelify": "^6.0.0",
+    "babel-cli": "^6.26.0",
+    "babel-preset-env": "^1.6.1",
```
gkatsev: does the default presets match our supported browser list?

package.json
```diff
@@ -32,7 +32,7 @@
     "docs:toc": "doctoc README.md",
     "lint": "vjsstandard",
     "start": "npm-run-all -p watch start:*",
-    "start:serve": "babel-node scripts/server.js",
+    "start:serve": "babel-node scripts/server.js --presets env",
```
gkatsev: also, using a `.babelrc` file would mean we can just have preset set in one place instead of sprinkled liberally everywhere :)
However, this is probably fine too.

forbesjo: See https://github.com/videojs/mpd-parser/blob/master/.babelrc for an example from another one of our projects

package.json
forbesjo: I think instead we should explicitly use the presets and plugins we need, see the .rc below

package.json
gkatsev: I think moving to babel-preset-env is a great idea, we just need to make sure we tell it to target the browsers we care about.

dmlap: I've moved the generic babel config into a .babelrc

dmlap:
@gkatsev @forbesjo updated things to use a common babel config. Let me know if you see anything else.

PR #32, opened 2018-02-09
fix: Don't reset eme options

forbesjo:
## Description
The source handler is currently resetting a player's eme options if they exist before source load, let's merge these options instead.

## Specific Changes proposed
Merge existing `player.eme()` options

## Requirements Checklist


forbesjo:
Note to self: don't edit directly in github, you'll take more comments than necessary to fix dumb things

PR #35, opened 2018-02-12
 Merge videojs-contrib-hls master into http-streaming master

gesinger:
Note that we should add cdnjs and jsDelivr CDN availability (then add them to the README)

## Requirements Checklist
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](http://jsbin.com/nofafidabo/edit?html,output))


forbesjo:
The `### IE10 and Below` note doesn't apply to VHS because it only uses MSE

PR #42, opened 2018-02-23
ignore: Don't apply browserify transform globally

forbesjo:
## Description
Previously the build would apply the `browserify-shim` transform globally so that the `videojs-contrib-media-sources` dependency would not pull in video.js. Now that we no longer have that dependency the transform only needs to be applied within the project itself.

## Requirements Checklist


forbesjo:
Closing as this change isn't really necessary at the moment.

PR #44, opened 2018-02-23
wait for both main and audio loaders for endOfStream if main starting media unknown

mjneil:
This fixes an issue introduced by #34 that can happen with very short playlists with demuxed audio on very slow connections. If the connection is slow enough such that the audio segment loader can complete requesting the entire audio playlist and trigger `ended` before the main segment loader has completed its request for the first video segment, the changes in #34 will assume the stream is audio only and just rely on the audio segment loaders ended state, ignoring the main loader, because `startingMedia_` is set after the first request has been completed. This change will force waiting for both the main and audio loaders to trigger ended if the starting media of the main loader is still unknown.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -623,10 +623,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     if (this.mediaTypes_.AUDIO.activePlaylistLoader) {
       // if the audio playlist loader exists, then alternate audio is active
-      if (this.mainSegmentLoader_.startingMedia_ &&
+      if (!this.mainSegmentLoader_.startingMedia_ ||
           this.mainSegmentLoader_.startingMedia_.containsVideo) {
-        // if the main segment loader contains video, then we need to wait for both the
-        // main and audio segment loaders to call endOfStream
+        // if we do not know if the main segment loader contains video yet or if we
+        // definitively know the main segment loader contains video, then we need to wait
+        // for both main and audio segment loaders to call endOfStream
         isEndOfStream = isEndOfStream && this.audioSegmentLoader_.ended_;
       } else {
         // otherwise just rely on the audio loader
--- a/test/master-playlist-controller.test.js
+++ b/test/master-playlist-controller.test.js
@@ -589,7 +589,7 @@ function(assert) {
 
   MPC.mediaSource.sourceBuffers[0].trigger('updateend');
 
-  assert.equal(videoEnded, 1, 'main segment loader triggered endded');
+  assert.equal(videoEnded, 1, 'main segment loader triggered ended');
   assert.equal(audioEnded, 0, 'audio segment loader did not trigger ended');
   assert.equal(MPC.mediaSource.readyState, 'open', 'Media Source not yet ended');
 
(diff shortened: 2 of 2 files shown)


src/master-playlist-controller.js
```diff
@@ -623,10 +623,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     if (this.mediaTypes_.AUDIO.activePlaylistLoader) {
       // if the audio playlist loader exists, then alternate audio is active
-      if (this.mainSegmentLoader_.startingMedia_ &&
+      if (!this.mainSegmentLoader_.startingMedia_ ||
```
gesinger: What happens in the event that we only have alternate audio, it hits the end of its playlist, and `mainSegmentLoader_.startingMedia_` is not yet set?

test/master-playlist-controller.test.js
```diff
@@ -603,6 +603,70 @@ function(assert) {
...
+  this.standardXHRResponse(this.requests.shift(), audioMedia);
+
+  // this.requests === [videoSegment, audioSegment]
+
+  // audio segment
+  this.standardXHRResponse(this.requests[1]);
+  // audio source buffer
+  MPC.mediaSource.sourceBuffers[1].trigger('updateend');
+
+  assert.equal(videoEnded, 0, 'main segment loader triggered endded');
```
gesinger: ended

test/master-playlist-controller.test.js
```diff
@@ -603,6 +603,70 @@ function(assert) {
...
+  assert.equal(audioEnded, 1, 'audio segment loader did not trigger ended');
```
gesinger: I think this comment and the one above are backwards (this one did trigger, other one didn't)

test/master-playlist-controller.test.js
```diff
@@ -603,6 +603,70 @@ function(assert) {
...
+  assert.equal(MPC.mediaSource.readyState, 'open', 'Media Source not yet ended');
+
+  // video segment
+  this.standardXHRResponse(this.requests[0]);
+
+  MPC.mainSegmentLoader_.startingMedia_ = { containsVideo: true };
+  // main source buffer
+  MPC.mediaSource.sourceBuffers[0].trigger('updateend');
+
+  assert.equal(videoEnded, 1, 'main segment loader did not trigger ended again');
```
gesinger: These descriptions may also be backwards

test/master-playlist-controller.test.js
mjneil: oops, botched copy paste

src/master-playlist-controller.js
mjneil: I do not believe this is a scenario we can get into. `STREAM-INF`s are REQUIRED to have a URI line by the spec. So there will always be a main stream that the main segment loader is loading from. In addition to that, until the way audio buffer ownership is overhauled, there isn't actually any mechanism to stop the main segment loader in an audio only stream with alternate audio activated. So even if audio only and alternate audio activated, the main segment loader will continue to request its audio segments, and then the VirtualSourceBuffer just drops them on the floor because it doesn't own the audio buffer. (Yes this is wasted bandwidth, even more reason to change the way we handle audio)

gesinger:
Tested 👍

PR #45, opened 2018-02-24
Fix program date time handling

mjneil:
The previous version of m3u8-parser had a bug that would associate the last `EXT-X-PROGRAM-DATE-TIME` tag with the start of the playlist.

This could cause problems in  live streams with the following properties
* ONE of:
  1. `EXT-X-PLAYLIST-TYPE:EVENT`
  2. a playlist with a sliding window, but viewed as the initial window is being filled, i.e. the playlist is still growing and segments are not yet being removed from the front
* Every segment has its own `EXT-X-PROGRAM-DATE-TIME` tag

When these conditions are met, each playlist refresh causes the `SyncController` to believe that there is a new sync point at the start of the playlist shifted by an amount that is equal to the amount of time added by new segments at the end. Since the playlist window is not actually sliding yet, the sync point for the start of the playlist should not be changed. This could cause issues when seeking backwards and then forwards. This is because seeking backwards causes a new `timestampOffset` to be set (because the segment that is being seeked to starts before the currently set `timestampOffset`). When `timestampOffset` changes, the `SyncController` also updates its PTS to Display time mapping. Since the `SyncController` has an incorrect sync point as described above, the mapping that gets set is incorrect. Seeking forward again after this can cause the player to select a segment that is beyond the playhead and stall indefinitely.

This PR updates the m3u8-parser version, which attaches date time info on each segment for each tag, instead of just on the playlist object. The `ProgramDateTime` strategy in `SyncController` will now look for the nearest segment to `currentTime` with date time information for an accurate sync point.

### DIFF
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -53,13 +53,11 @@ export const syncPointStrategies = [
             break;
           }
 
-          if (!syncPoint || lastDistance === null || lastDistance >= distance) {
-            lastDistance = distance;
-            syncPoint = {
-              time: segmentStart,
-              segmentIndex: i
-            };
-          }
+          lastDistance = distance;
+          syncPoint = {
+            time: segmentStart,
+            segmentIndex: i
+          };
         }
       }
       return syncPoint;
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -29,17 +29,40 @@ export const syncPointStrategies = [
   {
     name: 'ProgramDateTime',
     run: (syncController, playlist, duration, currentTimeline, currentTime) => {
-      if (syncController.datetimeToDisplayTime && playlist.dateTimeObject) {
-        let playlistTime = playlist.dateTimeObject.getTime() / 1000;
-        let playlistStart = playlistTime + syncController.datetimeToDisplayTime;
-        let syncPoint = {
-          time: playlistStart,
-          segmentIndex: 0
-        };
-
-        return syncPoint;
+      if (!syncController.datetimeToDisplayTime) {
+        return null;
       }
-      return null;
+
+      let segments = playlist.segments || [];
+      let syncPoint = null;
+      let lastDistance = null;
+
+      currentTime = currentTime || 0;
+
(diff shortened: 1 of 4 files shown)


src/sync-controller.js
```diff
@@ -336,8 +359,11 @@ export default class SyncController extends videojs.EventTarget {
    * @param {Playlist} playlist - The currently active playlist
    */
   setDateTimeMapping(playlist) {
-    if (!this.datetimeToDisplayTime && playlist.dateTimeObject) {
-      let playlistTimestamp = playlist.dateTimeObject.getTime() / 1000;
+    if (!this.datetimeToDisplayTime &&
+        playlist.segments &&
+        playlist.segments.length &&
+        playlist.segments[0].dateTimeObject) {
```
gesinger: Do we only want to consider the first segment?

src/sync-controller.js
```diff
@@ -29,17 +29,40 @@ export const syncPointStrategies = [
...
+          let segmentStart = segmentTime + syncController.datetimeToDisplayTime;
+          let distance = Math.abs(currentTime - segmentStart);
+
+          // Once the distance begins to increase, we have passed
+          // currentTime and can stop looking for better candidates
+          if (lastDistance !== null && lastDistance < distance) {
+            break;
+          }
+
+          if (!syncPoint || lastDistance === null || lastDistance >= distance) {
```
gesinger: Do we need this check here?

mjneil: Copied this logic from the Segment strategy below, but looking it over again, it doesn't look like it.

src/sync-controller.js
mjneil: I think yes. The intention of this date time mapping that is being stored is supposed to be the offset of real time to display time for the start of stream (at time of join, i.e. first segment in first playlist). If a segment other than the first is used, you'd have to account for the duration of the segments that came before it in the playlist when trying to calculate sync points from it. This has the potential for inaccuracies

PR #48, opened 2018-03-05
Live support for DASH

mjneil:
## Description
This PR adds live support for DASH manifests. Since dash doesn't always need to update the playlists to get new segments (e.g. when template is in use, all future segments are known), playlists are "refreshed" by just re-parsing the manifest. The mpd-parser uses `Date.now` to calculate the current live window, which VHS can operate on like it does with HLS. Dash still has the concept of manifest refreshes that is separate from the live window, so the `masterPlaylistLoader` handles refreshing the xml, while the other loaders just ask the master loader for a freshly parsed xml whenever a "refresh" needs to happen.

NOTE: This will require the next version of mpd-parser that includes https://github.com/videojs/mpd-parser/pull/22

edit: merged https://github.com/mjneil/http-streaming/pull/1

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -1,15 +1,48 @@
-import { EventTarget } from 'video.js';
+import { EventTarget, mergeOptions } from 'video.js';
 import mpdParser from 'mpd-parser';
 import {
+  refreshDelay,
   setupMediaPlaylists,
-  resolveMediaGroupUris
+  resolveMediaGroupUris,
+  updateMaster as updatePlaylist,
+  forEachMediaGroup
 } from './playlist-loader';
+
+export const updateMaster = (oldMaster, newMaster) => {
+  let update = mergeOptions(oldMaster, {
+    // These are top level properties that can be updated
+    duration: newMaster.duration,
+    minimumUpdatePeriod: newMaster.minimumUpdatePeriod
+  });
+
+  // First update the playlists in playlist list
+  for (let i = 0; i < newMaster.playlists.length; i++) {
+    const playlistUpdate = updatePlaylist(update, newMaster.playlists[i]);
+
+    if (playlistUpdate) {
+      update = playlistUpdate;
+    }
+  }
+
+  // Then update media group playlists
+  forEachMediaGroup(newMaster, (properties) => {
+    if (properties.playlists && properties.playlists.length) {
+      const playlistUpdate = updatePlaylist(update, properties.playlists[0]);
+
+      if (playlistUpdate) {
+        update = playlistUpdate;
+      }
+    }
+  });
+
+  return update;
+};
 
 export default class DashPlaylistLoader extends EventTarget {
   // DashPlaylistLoader must accept either a src url or a playlist because subsequent
   // playlist loader setups from media groups will expect to be able to pass a playlist
   // (since there aren't external URLs to media playlists with DASH)
-  constructor(srcUrlOrPlaylist, hls, withCredentials) {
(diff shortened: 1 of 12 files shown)


src/dash-playlist-loader.js
```diff
@@ -134,48 +228,166 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    // trigger loadedmetadata to resolve setup of media groups
+    // trigger async to mimic behavior of HLS, where it must request a playlist
+    setTimeout(() => {
+      this.trigger('loadedmetadata');
+    }, 0);
+
+    if (this.master.minimumUpdatePeriod) {
+      setTimeout(() => {
+        this.trigger('minimumUpdatePeriod');
+      }, this.master.minimumUpdatePeriod);
```
OshinKaramian: There is one special case here, where the minimumUpdatePeriod = 0.  Which effectively means "update this as often as you want".  Is this case covered?

mjneil: That case is not covered by this, I can add a TODO making note of that? I feel like for the scope of this PR we can ignore that case for now until it comes up. I think the best thing to do in this scenario would be to refresh the xml at the same rate we "refresh" the media playlist, i.e. targetDuration

OshinKaramian: TODO seems good.

src/playlist-loader.js
```diff
@@ -10,6 +10,27 @@ import { mergeOptions, EventTarget, log } from 'video.js';
 import m3u8 from 'm3u8-parser';
```
OshinKaramian: I think it makes sense, at some point, to rename some files here.  Given this is now a DASH/HLS project it might make sense to rename this to something like `hls-playlist-loader` (or something similar) to make it clear that there isn't one playlist to rule them all (or that one is not a subset of the other).

src/dash-playlist-loader.js
```diff
@@ -1,15 +1,64 @@
-import { EventTarget } from 'video.js';
+import { EventTarget, mergeOptions } from 'video.js';
 import mpdParser from 'mpd-parser';
 import {
+  refreshDelay,
```
OshinKaramian: Might make sense to extract some of these common functions into their own module at some point (doesn't have to be in this PR).  I think it will make how things are composed a little bit clearer.

test/dash-playlist-loader.test.js
```diff
@@ -184,18 +192,208 @@ QUnit.test('triggers an event when the active media changes', function(assert) {
...
+  loader.on('mediaupdatetimeout', () => refreshes++);
+
+  loader.parseMasterXml = () => {
+    const result = parseMasterXml_();
+
+    // add segment to segment list for proper refresh delay functionality
+    result.playlists[0].segments.push({ duration: 2, uri: 'segment-0' });
+
+    return result;
+  }
```
forbesjo: Missing semicolon

forbesjo:
Looks like there are some problems when a source has been changed, looking into it

src/dash-playlist-loader.js
```diff
@@ -134,48 +250,191 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    this.trigger('loadedplaylist');
+
+    if (!this.media_) {
+      // no media playlist was specifically selected so start
+      // from the first listed one
+      this.media(this.master.playlists[0]);
+    }
+    // trigger loadedmetadata to resolve setup of media groups
+    // trigger async to mimic behavior of HLS, where it must request a playlist
+    setTimeout(() => {
```
forbesjo: It might be best to do explicitly `window.setTimeout` for all occurrences of `setTimeout`

PR #61, opened 2018-03-20
WIP: transmux before append

gesinger:
Paired on work: @gesinger @mjneil 

## Description
NOTE: tests are currently failing (and new tests are missing) as this PR is a WIP.
NOTE: requires https://github.com/videojs/mux.js/pull/186

In order to provide more detailed content info and flexibility before we append a segment, move the transmuxing of TS segments to before we append. This allows us to: remove TS probe code, remove virtual source buffer code, look at program information before setting up source buffers, and have precise timing info about segments we will append.

## Requirements Checklist
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](http://jsbin.com/nofafidabo/edit?html,output))


src/segment-loader.js
```diff
@@ -128,6 +149,13 @@ const segmentInfoString = (segmentInfo) => {
   ].join(' ');
 };
 
+const isFmp4 = (segment) => {
+  // TODO we need a better way if determining if a segment is an mp4 or not. We rely on
+  // this method at the moment because the spec requires a map tag on each fmp4
+  // segment), however, ts segments may have the map tag as well
+  return !!segment.map;
```
imbcmdth: The easiest way here is to look at the first 8 bytes of the segment which will have 4 bytes for the size of the first atom (that you can ignore) followed by the four bytes: `[0x66, 0x74, 0x79, 0x70]` for the `ftyp` atom.

src/segment-loader.js
```diff
@@ -3,18 +3,39 @@
  */
 import Playlist from './playlist';
 import videojs from 'video.js';
-import SourceUpdater from './source-updater';
+import work from 'webwackify';
```
imbcmdth: I am a little saddened that we ended up putting this in `segment loader` and making it more complicated. :(

src/segment-loader.js
```diff
@@ -624,6 +661,8 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @private
    */
   fillBuffer_() {
+    // TODO since the source buffer maintains a queue, and we shouldn't call this function
+    // except when we're ready for the next segment, this check can most likely be removed
     if (this.sourceUpdater_.updating()) {
```
imbcmdth: This exists for a very important reason...  the particulars of while elude me now. 🎗

src/segment-loader.js
mjneil: Ya this is something that we want to cleanup and move to its own module

squarebracket:
Do I understand correctly that you're trying to do all segment parsing on MP4s (and thus after transmux for HLS TS)?

I ask because, in terms of pulling out the caption parser, I was imagining something like this:

```javascript
var captionParser = new CaptionParser(vttjs.VTTCue);

// We can do this now, yay
captionParser.on('newTrack', (event) => {
  this.inbandTextTracks_[event.track] = player.addRemoteTextTrack({}).track;
});

captionParser.on('data', (event) => {
  this.inbandTextTracks_[event.track].addCue(event.cue);
});

if (this.ts) {
  // let the transmuxer push data to the caption parser
  this.transmuxer_ = new Transmuxer({captionParser});
} else {
  // we need to parse out the NALs ourself
  captionParser.push(extractCaptionNals(segment.bytes));
}
```

Does that still make sense if we'll always be probing the resulting MP4? I'm pretty sure we could just keep the caption NALs on the transmuxed MP4. [Parsing out the specific boxes](https://github.com/Dash-Industry-Forum/dash.js/blob/development/src/dash/utils/FragmentedTextBoxParser.js#L47) is a fair bit heavier than [just parsing out the timing info](https://github.com/videojs/mux.js/blob/master/lib/mp4/probe.js), though it would probably still be beneficial to parse both at the same time in the case of fMP4 HLS/DASH.

forbesjo:
Closing since this is being broken into smaller PRs with updates

PR #64, opened 2018-03-21
fix: correct parser usage

forbesjo:
## Description
Update the parser imports to use the exposed APIs.

## Requirements Checklist


src/dash-playlist-loader.js
```diff
@@ -1,5 +1,5 @@
 import { EventTarget, mergeOptions } from 'video.js';
-import mpdParser from 'mpd-parser';
+import { parse as mpdParser, parseUTCTiming } from 'mpd-parser';
```
gesinger: Minor, but in the interest of a verb for the function, maybe `parse as parseMpd`?

PR #66, opened 2018-03-26
fix: videojs doesn't export individual items but an object

gkatsev:
Video.js exports an object and you can't do destructuring in imports, it's a special construct for named exports.
Also, modify the default export. Though, I can move that to a separate PR.

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -11,6 +11,7 @@ import { Parser as M3u8Parser } from 'm3u8-parser';
 import window from 'global/window';
 
 const { mergeOptions, EventTarget, log } = videojs;
+
 /**
  * Loops through all supported media groups in master and calls the provided
  * callback for each group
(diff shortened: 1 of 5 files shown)


src/videojs-http-streaming.js
```diff
@@ -698,7 +698,7 @@ if (videojs.registerPlugin) {
   videojs.plugin('reloadSourceOnError', reloadSourceOnError);
 }
 
-module.exports = {
+export default {
```
gkatsev: this should be equivalent but can be moved to a separate PR.

src/playlist-loader.js
```diff
@@ -6,10 +6,11 @@
  *
  */
 import resolveUrl from './resolve-url';
-import { mergeOptions, EventTarget, log } from 'video.js';
+import videojs from 'video.js';
 import { Parser as M3u8Parser } from 'm3u8-parser';
 import window from 'global/window';
 
+const { mergeOptions, EventTarget, log } = videojs;
```
forbesjo: Needs a new line

src/videojs-http-streaming.js
forbesjo: I think it's fine to have this change in this PR

PR #69, opened 2018-03-27
refactor: use rollup for build

forbesjo:
This switches the build system from browserify to rollup to work around using webwackify and allows for easier integration into video.js.

This PR also removes the browserify/webpack tests as well as code coverage. Code coverage was causing the tests to fail, we'll revisit and investigate the problem.


.babelrc
```diff
@@ -1,7 +1,4 @@
 {
-  "presets": [["env", {
-    "targets": {
-      "browsers": ["last 2 versions", "ie 9-11"]
-    }
-  }]]
+  presets: ['es3', ['es2015', { loose: true, modules: false }]],
+  plugins: ['external-helpers', 'transform-object-assign']
 }
```
thecotne: why not preset env?

gkatsev: we probably could but since we're targeting IE11 and up, es2015 preset is kind of easier.

.babelrc
```diff
@@ -1,7 +1,4 @@
 {
-  "presets": [["env", {
-    "targets": {
-      "browsers": ["last 2 versions", "ie 9-11"]
-    }
-  }]]
+  presets: ['es3', ['es2015', { loose: true, modules: false }]],
```
gkatsev: also, we don't need es3 preset anymore.

gkatsev: we can drop the es3 preset.

.babelrc
gkatsev: either one is fine. The babel team would probably prefer us to use the babel-preset-env, though.

thecotne: easier for what?

especially because we are targeting IE11 and up if we use env and target browsers properly we can end up with less compiled features and build size will be smaller and work better (on supported browsers)

also env preset can decide if we need es3 stuff automatically by target browsers

scripts/rollup.config.js
```diff
@@ -0,0 +1,109 @@
...
+    plugins: [
+      json({
+        preferConst: true
+      }),
+      worker()
+    ],
+    output: [{
+      name: 'videojs-http-streaming',
+      file: 'dist/videojs-http-streaming.es.js',
+      format: 'es',
```
gkatsev: we should probably add an onwarn method here to ignore the unresolved dep warnings, see: 
https://github.com/videojs/video.js/blob/378d98ee87b6ac48536d0b8f263bee1f76c68aab/build/rollup.js#L65-L73

package.json
```diff
@@ -2,117 +2,86 @@
...
       "test/test-manifests.js"
     ]
   },
   "files": [
-    "dist/",
-    "es5/"
+    "dist/"
   ],
   "dependencies": {
     "aes-decrypter": "1.0.3",
```
gkatsev: we should upgrade to aes-decrypter 3 at some point, it is rolluped. I think/hope that 3 has the correct decrypter stuff and not the broken version that 2.0 had.

forbesjo:
Requires a new version of the web worker plugin that has this change https://github.com/gkatsev/rollup-plugin-bundle-worker/pull/1

PR #75, opened 2018-04-02
chore: sync videojs-contrib-hls updates

forbesjo:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+sudo: false
 dist: trusty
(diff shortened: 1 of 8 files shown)


gesinger:
This seems to have copies of commits from https://github.com/videojs/http-streaming/pull/35

forbesjo:
That's because their histories have diverged, this PR will still get squashed like https://github.com/videojs/http-streaming/commit/3b52b2945d92b9c57b878a17692d304fcdd460ae

PR #102, opened 2018-05-31
chore: update tests and pages for video.js 7

forbesjo:
## Description
This PR updates all references to the video.js script to video.core.js so that VHS doesn't double up. This isn't totally necessary since VHS should be prepended to the source handler list but it's better to be safe.

This PR requires https://github.com/videojs/video.js/pull/5220

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "mpd-parser": "0.6.1",
     "mux.js": "4.4.1",
     "url-toolkit": "^2.1.3",
-    "video.js": "^6.8.0 || ^7.0.0"
+    "video.js": "^6.8.0 || ^7.0.3"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -65,7 +65,7 @@
     "mpd-parser": "0.6.1",
     "mux.js": "4.4.1",
     "url-toolkit": "^2.1.3",
-    "video.js": "^6.8.0 || ^7.0.0"
+    "video.js": "^6.8.0 || ^7.0.3"
```
OshinKaramian: Just need to update this to 7.0.4 (or whatever version the changes in `video.core.js` are in) when the time comes.

forbesjo: Well this really indicates what versions of video.js are compatible with VHS, I should actually bump this down to `^7.0.0` and update the lock file to 7.0.4 when that's out

PR #105, opened 2018-06-07
fix: Japanese text in IE

forbesjo:
## Description
Fixes https://github.com/videojs/video.js/issues/5070

## Specific Changes proposed
Removes text decoding from the VTT segment parsing by requesting the segments as `text` instead of `arraybuffer`

## Requirements Checklist

### DIFF
--- a/src/vtt-segment-loader.js
+++ b/src/vtt-segment-loader.js
@@ -7,12 +7,7 @@ import window from 'global/window';
 import { removeCuesFromTrack } from './mse/remove-cues-from-track';
 import { initSegmentId } from './bin-utils';
 
-const VTT_LINE_TERMINATORS =
-  new Uint8Array('\n\n'.split('').map(char => char.charCodeAt(0)));
-
-const uintToString = function(uintArray) {
-  return String.fromCharCode.apply(null, uintArray);
-};
+const VTT_LINE_TERMINATORS = '\n\n';
 
 /**
  * An object that manages segment loading and appending.
@@ -73,16 +68,9 @@ export default class VTTSegmentLoader extends SegmentLoader {
       // to follow the WebVTT spec (https://w3c.github.io/webvtt/#file-structure) that
       // requires two or more WebVTT line terminators between the WebVTT header and the
       // rest of the file
-      const combinedByteLength = VTT_LINE_TERMINATORS.byteLength + map.bytes.byteLength;
-      const combinedSegment = new Uint8Array(combinedByteLength);
-
-      combinedSegment.set(map.bytes);
-      combinedSegment.set(VTT_LINE_TERMINATORS, map.bytes.byteLength);
-
       this.initSegments_[id] = storedMap = {
         resolvedUri: map.resolvedUri,
-        byterange: map.byterange,
-        bytes: combinedSegment
+        bytes: map.bytes + VTT_LINE_TERMINATORS
       };
     }
 
@@ -282,8 +270,6 @@ export default class VTTSegmentLoader extends SegmentLoader {
       return;
     }
 
-    segmentInfo.byteLength = segmentInfo.bytes.byteLength;
-
     this.mediaSecondsLoaded += segment.duration;
 
     if (segmentInfo.cues.length) {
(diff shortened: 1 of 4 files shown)


src/vtt-segment-loader.js
```diff
@@ -73,16 +66,9 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
-      const combinedSegment = new Uint8Array(combinedByteLength);
-
-      combinedSegment.set(map.bytes);
-      combinedSegment.set(VTT_LINE_TERMINATORS, map.bytes.byteLength);
-
       this.initSegments_[id] = storedMap = {
         resolvedUri: map.resolvedUri,
-        byterange: map.byterange,
-        bytes: combinedSegment
+        bytes: map.bytes + '\n\n'
```
gesinger: Might be worth retaining a constant for `VTT_LINE_TERMINATORS`

src/vtt-segment-loader.js
```diff
@@ -307,19 +291,12 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
-    let decodeBytesToString = false;
-
-    if (typeof window.TextDecoder === 'function') {
-      decoder = new window.TextDecoder('utf8');
-    } else {
-      decoder = window.WebVTT.StringDecoder();
-      decodeBytesToString = true;
-    }
-
+    const fakeDecoder = {
```
gesinger: This might make more sense as an inline lambda, and we should probably have a comment around why it is used.

src/vtt-segment-loader.js
```diff
@@ -307,19 +291,14 @@ export default class VTTSegmentLoader extends SegmentLoader {
    * @private
    */
   parseVTTCues_(segmentInfo) {
-    let decoder;
-    let decodeBytesToString = false;
-
-    if (typeof window.TextDecoder === 'function') {
```
gkatsev: why not keep this? It'll make it less likely that we'll end up with garbled text.

forbesjo: `TextDecoder.decode()` only takes an `ArrayBuffer` so we'd have to convert from text to a buffer back to text. Is this what we want?
https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode

gkatsev: Oh, I see. In which case, maybe we should just use the StringDecoder to match what we do in Video.js?

gkatsev: Also, I wonder if that means that what's happening in vtt.js is not exactly right?
https://github.com/videojs/vtt.js/blob/1f41da80cbd0bb08543d210fbfdbdb3f3d06ba3b/lib/vtt.js#L1014
https://github.com/videojs/vtt.js/blob/1f41da80cbd0bb08543d210fbfdbdb3f3d06ba3b/lib/vtt.js#L1028-L1037

forbesjo: We originally used StringDecoder and that was causing issues

gkatsev: it was an issue because the source itself was already garbled and we weren't decoding things properly. Given that non-manifest captions worked, it can't be an issue since that's what we use in Video.js.

forbesjo:
Looks like all these changes can be avoided by making a change in vtt.js, will open a PR soon and link to it here

forbesjo:
https://github.com/videojs/vtt.js/pull/25

forbesjo:
Instead of making a change in vtt.js we will instead go forward with this PR and avoid unnecessary data transformations

src/vtt-segment-loader.js
```diff
@@ -307,19 +293,16 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
+    // we'll pass in a fake decoder
+    const fakeDecoder = {
+      decode(d) {
+        return d;
+      }
+    };
     const parser = new window.WebVTT.Parser(window,
                                             window.vttjs,
-                                            decoder);
+                                            fakeDecoder);
```
gkatsev: so, should we just be using the StringDecoder? Don't think it makes a huge difference, though.

forbesjo: I'll use that instead since the original issue seems to stem from transforming the array buffer to text.

forbesjo:
- update the vttsegmentloader to request as text and not set this option in media-groups
- store the vtt text in a `text` property and leave `bytes` alone

forbesjo:
Closing in favor of a simpler solution (#157)

PR #108, opened 2018-06-12
feat(captions): write in-band captions from DASH fmp4 segments to the textTrack API

ldayananda:
## Description
Depends on https://github.com/videojs/mux.js/pull/197. 

This will parse captions from DASH fmp4 segments as each segment request is finished and add them to a corresponding text track(i.e CC1, CC2, CC3, CC4).

## Specific Changes proposed
- requires an update to the mux.js dependency

## Requirements Checklist


ldayananda:
This should be ready for review, the tests fail because the Mux.js PR hasn't been merged and released yet. 

I don't think there is anything needed for documentation here?

PR #113, opened 2018-06-22
Remove buffered data on fast quality switches

alex-barstow:
## Description
This adds logic to remove all buffered data when a new playlist is enabled via `Representation.enabled`, such as with manual quality selection. This change comes following discussion with @gesinger, @forbesjo, and @mjneil. There was also some talk about modifying [this conditional](https://github.com/videojs/http-streaming/blob/master/src/segment-loader.js#L707) to check `fetchAtBuffer_` to reduce delay before fetching new segments at the playhead. Let me know if you would like me to add that to this PR.

## Specific Changes proposed
* Rename the existing `fastQualityChange_` function to `smoothQualityChange_`
* Create new function called `fastQualityChange_` which calls `resetEverything()` instead of just `resetLoader()`
* modify existing tests for the old `fastQualityChange_` method to reflect the new function name
* add new test for the new `fastQualityChange_` method


gkatsev:
This would also fix https://github.com/videojs/video.js/issues/5044, I believe.

mjneil:
Did we want to put this behind a feature flag (either enabled or disabled by default) so that users can keep the smooth manual switch if they want?

forbesjo:
In testing this PR it looks like there's a bit of a stutter when manually switching renditions. I'm expecting to see a loading spinner while rebuffering and smooth playback. Perhaps the stuttering happens when near the end of a segment?

Here's a gif of the behavior (BC skin)
fast-switch

alex-barstow:
@forbesjo Hm I'll into that. Interestingly, Firefox handles the switch smoothly

alex-barstow:
I believe the solution for the stutter is to seek in place after clearing the buffer. Chrome seems to cache a few frames from the previous rendition, and seeking in place gives the browser the "kick" it needs to clear them.

test/master-playlist-controller.test.js
```diff
@@ -323,6 +323,56 @@ QUnit.test('fast quality change resyncs audio segment loader', function(assert)
...
+
+  this.masterPlaylistController.mediaSource.trigger('sourceopen');
+  // master
+  this.standardXHRResponse(this.requests.shift());
+  // media
+  this.standardXHRResponse(this.requests.shift());
+
+  let segmentLoader = this.masterPlaylistController.mainSegmentLoader_;
+
+  segmentLoader.resyncLoader = function() {
```
gesinger: minor: but for this (and below) we can save some lines using arrows, e.g., `() => resyncs++`

test/master-playlist-controller.test.js
```diff
@@ -323,6 +323,56 @@ QUnit.test('fast quality change resyncs audio segment loader', function(assert)
...
+
+  assert.equal(segmentLoader.ended_, false, 'segment loader ended property is false');
+
+  assert.equal(resyncs, 1, 'resynced segment loader');
+
+  assert.equal(resets, 1, 'reseteverything event triggered successfully');
+
+  assert.deepEqual(removeFuncArgs, {start: 0, end: 60}, 'remove() called with correct arguments');
+
+  // verify stats
```
gesinger: This kinda hung on in too many tests, but I think it's OK to remove here.

test/master-playlist-controller.test.js
```diff
@@ -323,6 +323,56 @@ QUnit.test('fast quality change resyncs audio segment loader', function(assert)
...
+  segmentLoader.duration_ = function() {
+    return 60;
+  };
+
+  segmentLoader.on('reseteverything', function() {
+    resets++;
+  });
+
+  this.masterPlaylistController.selectPlaylist = () => {
+    return this.masterPlaylistController.master().playlists[0];
```
gesinger: We may also want a test for not resetting if the media remains the same.

src/master-playlist-controller.js
```diff
@@ -506,6 +506,42 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      const currentTime = this.tech_.currentTime();
+
+      this.masterPlaylistLoader_.media(media);
+
+      // delete all buffered data to allow an immediate quality switch
+      this.mainSegmentLoader_.resetEverything();
+
+      // Wait for the buffer to finish clearing, then seek in place to give the
+      // browser a kick to remove any cached frames from the previous rendition
+      sourceBuffer.one('updateend', function() {
```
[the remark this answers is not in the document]
alex-barstow: Same goes for the currentTime-- I think it can just be `this.tech_.setCurrentTime(this.tech_.currentTime());`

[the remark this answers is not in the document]
alex-barstow: @gesinger I suppose if we are in the middle of an append operation when the quality switch is initiated it might be possible? When we get the `updateend` we  could check the segment loader state to make sure it isn't `'APPENDING'`. Thoughts?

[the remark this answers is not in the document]
alex-barstow: @gesinger Makes sense. Went ahead and made those changes so let me know how they look.

src/master-playlist-controller.js
```diff
@@ -510,6 +510,36 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+
+    this.masterPlaylistLoader_.media(media);
+
+    // delete all buffered data to allow an immediate quality switch, then seek
+    // in place to give the browser a kick to remove any cached frames from the
+    // previous rendition
+    this.mainSegmentLoader_.resetEverything(() => {
+      // we want SegmentLoader.load() to be called only once the new playlist
+      // has finished loading. This prevents it from being called too soon.
+      this.hls_.ignoreNextSeekingEvent_ = true;
```
gesinger: With the merge of https://github.com/videojs/http-streaming/pull/161 `ignoreNextSeekingEvent_` was removed, and a new method of seeking was introduced. It's possible `seekTo` may be used directly and `ignoreNextSeekingEvent_` removed.

src/segment-loader.js
```diff
@@ -582,10 +584,12 @@ export default class SegmentLoader extends videojs.EventTarget {
    * Remove any data in the source buffer between start and end times
    * @param {Number} start - the start time of the region to remove from the buffer
    * @param {Number} end - the end time of the region to remove from the buffer
+   * @param {Function} done - an optional callback to be executed when the remove
```
gesinger: For JSDoc optionals we can use `[done]`

src/source-updater.js
```diff
@@ -131,14 +131,16 @@ export default class SourceUpdater {
    *
    * @param {Number} start where to start the removal
    * @param {Number} end where to end the removal
+   * @param {Function} done optional callback to be executed when the remove
```
gesinger: For JSDoc optionals we can use `[done]`, or even `[done=noop]` here

test/master-playlist-controller.test.js
```diff
@@ -324,6 +324,59 @@ QUnit.test('fast quality change resyncs audio segment loader', function(assert)
...
+  assert.deepEqual(removeFuncArgs, {}, 'remove() not called if media is unchanged');
+
+  // media is changed
+  this.masterPlaylistController.selectPlaylist = () => {
+    return this.masterPlaylistController.master().playlists[0];
+  };
+
+  this.masterPlaylistController.fastQualityChange_();
+
+  assert.equal(segmentLoader.ended_, false, 'segment loader ended property is false');
```
gesinger: Was there a specific reason for the ended check?

alex-barstow: I included it since `ended_` gets set to `false` as a side effect of calling `resetEverything()`. If you think it's superfluous I can remove it.

src/master-playlist-controller.js
alex-barstow: @gesinger Hm just calling `seekTo` with the flag removed is problematic since the logic in `MasterPlaylistController.setCurrentTime()` results in `SegmentLoader.load()` getting called before the new playlist has finished loading, so we append another segment from the previous rendition before appending the new quality segments. Perhaps adding an another `return` in `MPC.setCurrentTime()` if the PlaylistLoader state is `'SWITCHING_MEDIA`` would be a possible solution to this? Off the top of my head I'm not sure what other consequences that could have though.

gesinger: If we can avoid looking at the PlaylistLoader state, that would be best. One thing that could be done here is use the normal `this.tech_.setCurrentTime` call directly, if we want to avoid the `MasterPlaylistController.setCurrentTime()` call, since `seekTo` is meant to add exactly that to the `setCurrentTime` calls. If we do use `this.tech_.setCurrentTime` directly though, it would be good to add a comment about why we can't go through the normal `seekTo` method (in this case to avoid resetting the loader, since it's a seek meant to kick things instead of actually performing a seek with buffer clearing).

test/master-playlist-controller.test.js
gesinger: Probably don't need it here. While it's a decent test, since these tests are meant for MasterPlaylistController we can probably leave checking that state change within the SegmentLoader tests.

alex-barstow:
Had to rebase to get the netlify deploy to succeed. @gesinger Ready for another look.

PR #157, opened 2018-07-17
fix: convert non-latin characters in IE

forbesjo:
## Description
Fixes videojs/video.js#5070

## Specific Changes proposed
Fixes the fallback text decoder. (https://stackoverflow.com/a/17192845)

## Requirements Checklist

### DIFF
--- a/src/util/string.js
+++ b/src/util/string.js
@@ -1,0 +1,2 @@
+export const uintToString = (uintArray) =>
+  decodeURIComponent(escape(String.fromCharCode.apply(null, uintArray)));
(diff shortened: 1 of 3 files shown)


src/util/string.js
```diff
@@ -0,0 +1,2 @@
+export const uintToString = (uintArray) =>
```
gesinger: Minor, but this function is doing a bit more than just converting the typed array to a string. We may want to either rename the function or just call decodeURIComponent and escape within vtt-segment-loader.

test/util/string.test.js
```diff
@@ -0,0 +1,10 @@
+import QUnit from 'qunit';
+import { uintToString } from '../../src/util/string';
+
+QUnit.module('uintToString');
```
gesinger: Minor, but may want a blank line here

PR #164, opened 2018-07-24
Added skipped fmp4 test

mrocajr:
## Description
Added skipped fmp4 test against playlist with both ts and fmp4 segments which will be enabled once we've added the feature.

### DIFF
--- a/test/playback.test.js
+++ b/test/playback.test.js
@@ -59,7 +59,7 @@ QUnit.test('Advanced Bip Bop', function(assert) {
   });
 });
 
-QUnit.skip('Fmp4', function(assert) {
+QUnit.skip('playlist with fmp4 and ts segments', function(assert) {
   let done = assert.async();
 
   assert.expect(2);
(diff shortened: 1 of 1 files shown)


test/playback.test.js
```diff
@@ -38,7 +38,7 @@ QUnit.module('Playback', {
   }
 });
 
-QUnit.test('Advanced Bip Bop', function(assert) {
+QUnit.skip('Fmp4', function(assert) {
```
gesinger: I think this test and advanced bip bop got flipped around.

test/playback.test.js
```diff
@@ -59,6 +59,27 @@ QUnit.test('Advanced Bip Bop', function(assert) {
   });
 });
 
+QUnit.skip('Fmp4', function(assert) {
```
gesinger: I think it would be worth specifying that this is a playlist with both fmp4 and ts segments, as we do support fmp4 alone right now.

test/playback.test.js
```diff
@@ -59,6 +59,27 @@ QUnit.test('Advanced Bip Bop', function(assert) {
...
+  let done = assert.async();
+
+  assert.expect(2);
+  let player = this.player;
+
+  player.autoplay(true);
+
+  playFor(player, 6, function() {
+    assert.ok(true, 'played for at least six seconds order over to hit the change in codec');
```
gesinger: I think "order over" can be removed. And it should just be a change in container format, the codec should remain the same.

Issue #165, opened 2018-07-24
Troubleshooting Guide/FAQ


forbesjo:
Create a guide to help users with problems in content or player setup.

forbesjo:
__Content plays on Mac but not Windows__
Some browsers may not be able to play audio sample rate rate is higher than 44.1 or 48
__Re-encode with a different audio sample rate rate__

forbesjo:
Link to the mux.js troubleshooting guide https://github.com/videojs/mux.js/blob/master/docs/troubleshooting.md

forbesjo:
__"No compatible source was found" on IE11 Win 7__
videojs-http-streaming does not support Flash HLS playback like the videojs-contrib-hls plugin
__Please use the flashls source handler https://github.com/brightcove/videojs-flashls-source-handler__

PR #169, opened 2018-07-25
Troubleshooting doc

gesinger:
https://github.com/videojs/http-streaming/issues/165

## Requirements Checklist
  - [X] Docs/guides updated

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -7,6 +7,8 @@
 ## Table of Contents
 - [Content plays on Mac but not on Windows](#content-plays-on-mac-but-not-windows)
 - ["No compatible source was found" on IE11 Win 7](#no-compatible-source-was-found-on-ie11-win-7)
+- [CORS: No Access-Control-Allow-Origin header](#cors-no-access-control-allow-origin-header)
+- [Desktop Safari/iOS Safari/Android exhibit different behavior from other browsers](#desktop-safari-ios-safari-android-exhibit-different-behavior-from-other-browsers)
 
 ## Content plays on Mac but not Windows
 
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
+# Troubleshooting Guide
+
+## Other troubleshooting guides
```
ldayananda: 💯

docs/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Content plays on Mac but not Windows
+
+Some browsers may not be able to play audio sample rates higher than 48 kHz. See https://docs.microsoft.com/en-gb/windows/desktop/medfound/aac-decoder#format-constraints
+
+Potential solution: re-encode with a Windows supported audio sample rate
+
+## "No compatible source was found" on IE11 Win 7
+
+videojs-http-streaming does not support Flash HLS playback (like the videojs-contrib-hls plugin did)
```
ldayananda: >(like the videojs-contrib-hls plugin did)
does?

docs/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
...
+
+Solution: please use the FlasHLS source handler https://github.com/brightcove/videojs-flashls-source-handler
```
ldayananda: might want to also mention the flash tech?

forbesjo:
Would be worth mentioning CORS issues https://github.com/videojs/videojs-contrib-hls#hosting-considerations

As well as `overrideNative`

docs/troubleshooting.md
```diff
@@ -16,6 +18,22 @@ Potential solution: re-encode with a Windows supported audio sample rate
...
+
+If you see an error along the lines of
+
+```
+XMLHttpRequest cannot load ... No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin ... is therefore not allowed access.
+```
+
+you need to properly configure CORS on your server: https://github.com/videojs/http-streaming#hosting-considerations
+
+## Desktop Safari/iOS Safari/Android exhibit different behavior from other browsers
```
forbesjo: I think most users will see a stream work on desktop Chrome and not on mobile Chrome or Edge. I think we should mention those browsers specifically.

docs/troubleshooting.md
OshinKaramian: Wonder if it might be useful to highlight what kinds of issues the docs here solve.  As an example noting generally what kinds of items might be in the mux doc vs the one here.

PR #174, opened 2018-07-30
docs: add note for Safari captions error

forbesjo:
Update documentation to handle a Safari decode error scenario

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -9,7 +9,7 @@ For issues around data embedded into media segments (e.g., 608 captions), see th
 - ["No compatible source was found" on IE11 Win 7](#no-compatible-source-was-found-on-ie11-win-7)
 - [CORS: No Access-Control-Allow-Origin header](#cors-no-access-control-allow-origin-header)
 - [Desktop Safari/iOS Safari/Android Chrome/Edge exhibit different behavior from other browsers](#desktop-safariios-safariandroid-chromeedge-exhibit-different-behavior-from-other-browsers)
-- [media_err_decode error on Desktop Safari](#media_err_decode-error-on-desktop-safari)
+- [MEDIA_ERR_DECODE error on Desktop Safari](#media_err_decode-error-on-desktop-safari)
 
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -37,3 +38,10 @@ you need to properly configure CORS on your server: https://github.com/videojs/h
 Some browsers support native playback of certain streaming formats. By default, we defer to the native players. However, this means that features specific to videojs-http-streaming will not be available.
 
 Solution: use videojs-http-streaming based playback on those devices: https://github.com/videojs/http-streaming#overridenative
+
+## media_err_decode error on Desktop Safari
```
gesinger: Probably worth using all caps since the error comes back in all caps

docs/troubleshooting.md
```diff
@@ -37,3 +38,10 @@ you need to properly configure CORS on your server: https://github.com/videojs/h
...
+
+This error may occur if the source manifest has `CLOSED-CAPTIONS=NONE` and an external text track is loaded into the player. Safari does not allow the inclusion any captions if the manifest indicates that captions will not be provided.
```
gesinger: We may also want to mention that MEDIA_ERR_DECODE can happen for a number of reasons, as it is particularly common for misconfigured content, the captions being just one of them.

docs/troubleshooting.md
```diff
@@ -9,6 +9,7 @@ For issues around data embedded into media segments (e.g., 608 captions), see th
 - ["No compatible source was found" on IE11 Win 7](#no-compatible-source-was-found-on-ie11-win-7)
 - [CORS: No Access-Control-Allow-Origin header](#cors-no-access-control-allow-origin-header)
 - [Desktop Safari/iOS Safari/Android Chrome/Edge exhibit different behavior from other browsers](#desktop-safariios-safariandroid-chromeedge-exhibit-different-behavior-from-other-browsers)
+- [media_err_decode error on Desktop Safari](#media_err_decode-error-on-desktop-safari)
```
gesinger: One more spot for all caps

PR #194, opened 2018-08-09
ignore: use currentSrc instead of currentSource().src

forbesjo:
## Description
Updating code to use a shorter API


forbesjo:
this isn't totally necessary, closing

PR #213, opened 2018-08-28
fix: stutter after fast quality change in IE/Edge

alex-barstow:
## Description

This tweaks the changes in https://github.com/videojs/http-streaming/pull/113 to fix a stutter in IE/Edge after the buffer is cleared. In the initial PR, seeking in place was sufficient to address this issue in Chrome, but we need to seek further behind/ahead of the current time in order to also fix the issue in IE/Edge.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -523,21 +523,30 @@ export class MasterPlaylistController extends videojs.EventTarget {
    */
   fastQualityChange_() {
     const media = this.selectPlaylist();
+    const isPaused = this.tech_.paused();
 
     if (media === this.masterPlaylistLoader_.media()) {
       return;
     }
 
+    this.tech_.pause();
     this.masterPlaylistLoader_.media(media);
 
-    // delete all buffered data to allow an immediate quality switch, then seek
-    // in place to give the browser a kick to remove any cached frames from the
-    // previous rendition
+    // Delete all buffered data to allow an immediate quality switch, then seek
+    // forward slightly to give the browser a kick to remove any cached frames from the
+    // previous rendition (.02 seconds ahead is roughly the minimum that will accomplish this
+    // in IE and Edge)
+    // Edge/IE bug: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14600375/
+    // Chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=651904
     this.mainSegmentLoader_.resetEverything(() => {
       // Since this is not a typical seek, we avoid the seekTo method which can cause
       // segments from the previously enabled rendition to load before the new playlist
       // has finished loading
-      this.tech_.setCurrentTime(this.tech_.currentTime());
+      this.tech_.setCurrentTime(this.tech_.currentTime() + .02);
+
+      if (!isPaused) {
+        this.tech_.play();
+      }
     });
 
     // don't need to reset audio as it is reset when media changes
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -534,15 +534,15 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // Delete all buffered data to allow an immediate quality switch, then seek
     // forward slightly to give the browser a kick to remove any cached frames from the
-    // previous rendition (.02 seconds ahead is roughly the minimum that will accomplish this
-    // in IE and Edge)
+    // previous rendition (.04 seconds ahead is roughly the minimum that will accomplish this
+    // across a variety of content in IE and Edge)
     // Edge/IE bug: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14600375/
     // Chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=651904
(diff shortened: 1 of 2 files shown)


src/master-playlist-controller.js
```diff
@@ -523,21 +523,30 @@ export class MasterPlaylistController extends videojs.EventTarget {
    */
   fastQualityChange_() {
     const media = this.selectPlaylist();
+    const isPaused = this.tech_.paused();
 
     if (media === this.masterPlaylistLoader_.media()) {
       return;
     }
 
+    this.tech_.pause();
```
gesinger: Is it necessary for us to pause? I think that may have inadvertent side effects.

src/master-playlist-controller.js
```diff
@@ -523,21 +523,30 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    // previous rendition (.04 seconds ahead is roughly the minimum that will accomplish this
+    // across a variety of content in IE and Edge)
+    // Edge/IE bug: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14600375/
+    // Chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=651904
     this.mainSegmentLoader_.resetEverything(() => {
       // Since this is not a typical seek, we avoid the seekTo method which can cause
       // segments from the previously enabled rendition to load before the new playlist
       // has finished loading
-      this.tech_.setCurrentTime(this.tech_.currentTime());
+      this.tech_.setCurrentTime(this.tech_.currentTime() + .04);
```
gesinger: Should we be seeking forwards for all browsers, or just those that are problematic?

alex-barstow: I was on the fence about whether it's better to minimize the scope of the forward seeking, or to omit any user agent checks and have uniform quality switch behavior across browsers. I'll defer to your judgement on this.

src/master-playlist-controller.js
alex-barstow: The rendition switch occurs without a stutter even without pausing/playing, but for some reason `timeupdate` events continue to fire even while playback has stopped and the loading spinner isn't shown. This only happens in IE/Edge.

src/master-playlist-controller.js
gesinger: I'd be interested to hear others' thoughts as well, but I think minimizing the scope may be better, just to limit potential impact, especially when it's only one set of browsers (IE/Edge).

alex-barstow:
Added the IE/Edge browser checks, removed the pause/play, and added additional tests

test/master-playlist-controller.test.js
```diff
@@ -376,81 +376,102 @@ QUnit.test('resets everything for a fast quality change', function(assert) {
...
-  this.masterPlaylistController.mainSegmentLoader_.sourceUpdater_.processedAppend_ = true;
+  this.player.tech_.on('seeking', function() {
+    seeks++;
+  });
 
   const timeBeforeSwitch = this.player.currentTime();
 
   this.masterPlaylistController.fastQualityChange_();
-
   this.masterPlaylistController.mediaSource.sourceBuffers[0].trigger('updateend');
```
gesinger: Is this responding to an actual "append?" I think if the fastQualityChange happened, there would still have to be a segment request before an append would happen.

test/master-playlist-controller.test.js
```diff
@@ -376,81 +376,102 @@ QUnit.test('resets everything for a fast quality change', function(assert) {
...
+  });
 
   const timeBeforeSwitch = this.player.currentTime();
 
   videojs.browser.IE_VERSION = null;
   videojs.browser.IS_EDGE = true;
 
   this.masterPlaylistController.fastQualityChange_();
-
   this.masterPlaylistController.mediaSource.sourceBuffers[0].trigger('updateend');
```
gesinger: Same as above.

test/master-playlist-controller.test.js
```diff
@@ -376,81 +376,102 @@ QUnit.test('resets everything for a fast quality change', function(assert) {
...
+  });
 
   const timeBeforeSwitch = this.player.currentTime();
 
   videojs.browser.IE_VERSION = 11;
   videojs.browser.IS_EDGE = false;
 
   this.masterPlaylistController.fastQualityChange_();
-
   this.masterPlaylistController.mediaSource.sourceBuffers[0].trigger('updateend');
```
gesinger: Same as above

test/master-playlist-controller.test.js
```diff
@@ -375,6 +375,111 @@ QUnit.test('resets everything for a fast quality change', function(assert) {
...
+    return this.masterPlaylistController.master().playlists[0];
+  };
+
+  this.player.tech_.on('seeking', function() {
+    seeks++;
+  });
+
+  const timeBeforeSwitch = this.player.currentTime();
+
+  videojs.browser.IE_VERSION = null;
```
gesinger: Here and the test below, we just have to reset the value to what it was originally at the end of the test.

alex-barstow: Yep good catch

PR #229, opened 2018-09-13
Fix: remove player props on dispose to stop middleware 

BrandonOCasey:
After the tech/vhs is disposed the middleware can still end up running for the next source because `player.vhs` still exists and that is what the middleware looks for.  See: https://github.com/videojs/http-streaming/blob/master/src/middleware-set-current-time.js

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -975,9 +975,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.mainSegmentLoader_.dispose();
 
     ['AUDIO', 'SUBTITLES'].forEach((type) => {
-      if (!this.mediaTypes_[type]) {
-        return;
-      }
       const groups = this.mediaTypes_[type].groups;
 
       for (let id in groups) {
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -975,6 +975,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.mainSegmentLoader_.dispose();
 
     ['AUDIO', 'SUBTITLES'].forEach((type) => {
+      if (!this.mediaTypes_[type]) {
+        return;
+      }
       const groups = this.mediaTypes_[type].groups;
 
       for (let id in groups) {
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -739,8 +739,10 @@ class HlsHandler extends Component {
       this.qualityLevels_.dispose();
     }
 
-    this.player_.vhs = null;
-    this.player_.dash = null;
+    if (this.player_) {
+      this.player_.vhs = null;
+      this.player_.dash = null;
+    }
 
     if (this.tech_ && this.tech_.hls) {
       this.tech_.hls = null;
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -325,7 +325,8 @@ class HlsHandler extends Component {
             videojs.log.warn('player.hls is deprecated. Use player.tech().hls instead.');
             tech.trigger({ type: 'usage', name: 'hls-player-access' });
             return this;
-          }
+          },
+          configurable: true
         });
       }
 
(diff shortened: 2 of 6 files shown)


test/test-helpers.js
```diff
@@ -310,15 +310,16 @@ export const openMediaSource = function(player, clock) {
...
-  player.tech_.hls.xhr = xhrFactory();
-
-  // simulate the sourceopen event
-  player.tech_.hls.mediaSource.readyState = 'open';
-  player.tech_.hls.mediaSource.dispatchEvent({
-    type: 'sourceopen',
-    swfId: player.tech_.el().id
-  });
-  clock.tick(1);
+  if (player.tech_.hls) {
```
BrandonOCasey: This helper assumed that hls would always be on tech, which this pr gets rid of.

test/karma.conf.js
```diff
@@ -25,10 +25,10 @@ module.exports = function(config) {
       'node_modules/video.js/dist/video-js.css',
       'dist-test/videojs-http-streaming.test.js'
     ],
-    browserConsoleLogOptions: {
+    /*browserConsoleLogOptions: {
```
gesinger: Was this left in?

test/test-helpers.js
gesinger: Won't it only get rid of it on dispose though?

test/karma.conf.js
forbesjo: commented code

test/test-helpers.js
BrandonOCasey: Yeah it should only get rid of it on dispose, but that will happen between sources as well.

src/master-playlist-controller.js
```diff
@@ -969,6 +969,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.mainSegmentLoader_.dispose();
 
     ['AUDIO', 'SUBTITLES'].forEach((type) => {
+      if (!this.mediaTypes_[type]) {
```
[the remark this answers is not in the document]
gesinger: That's odd. Do you know what the error was? It seems like both of those media types should exist if an instance of MasterPlaylistController exists.

test/test-helpers.js
forbesjo: Let's prefer to return early

src/videojs-http-streaming.js
```diff
@@ -643,6 +645,16 @@ class HlsHandler extends Component {
     if (this.qualityLevels_) {
       this.qualityLevels_.dispose();
     }
+
+    if (this.player_) {
+      this.player_.vhs = null;
+      this.player_.dash = null;
```
forbesjo: There's also `player.hls`

src/videojs-http-streaming.js
```diff
@@ -325,7 +325,8 @@ class HlsHandler extends Component {
             videojs.log.warn('player.hls is deprecated. Use player.tech().hls instead.');
             tech.trigger({ type: 'usage', name: 'hls-player-access' });
             return this;
-          }
+          },
+          configurable: true
```
gesinger: What's this property for?

src/master-playlist-controller.js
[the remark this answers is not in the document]
gesinger: Did we ever discover what might have been causing it? I'm always a bit hesitant to add code to fix a test error.

src/videojs-http-streaming.js
gkatsev: By default configurable is false, means you can't change the value once set.

gkatsev: change including delete.

src/master-playlist-controller.js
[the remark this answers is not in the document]
forbesjo: Perhaps this https://github.com/videojs/http-streaming/blob/master/test/videojs-http-streaming.test.js#L2884 ? I'll investigate and update the test if that's the problem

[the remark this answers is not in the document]
forbesjo: Confirmed, updating

PR #235, opened 2018-09-19
add back manual smooth quality changes via smoothQualityChange flag

squarebracket:
## Description
#113 unfortunately _completely_ removed smooth quality switching for manual quality switches. This PR adds it back via a configuration option.

## Specific Changes proposed
- Add `smoothQualityChange` config option
- Merge it with a possible source option of the same name
- Select `fastQualityChange` or `smoothQualityChange` using a ternary that checks the aforementioned config option

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ Video.js Compatibility: 6.0, 7.0
       - [blacklistDuration](#blacklistduration)
       - [bandwidth](#bandwidth)
       - [enableLowInitialPlaylist](#enablelowinitialplaylist)
+      - [limitRenditionByPlayerDimensions](#limitrenditionbyplayerdimensions)
+      - [smoothQualityChange](#smoothqualitychange)
   - [Runtime Properties](#runtime-properties)
     - [hls.playlists.master](#hlsplaylistsmaster)
     - [hls.playlists.media](#hlsplaylistsmedia)
@@ -341,6 +343,22 @@ When `limitRenditionByPlayerDimensions` is set to true, rendition
 selection logic will take into account the player size and rendition
 resolutions when making a decision.
 This setting is `true` by default.
+
+##### smoothQualityChange
+* Type: `boolean`
+* can be used as a source option
+* can be used as an initialization option
+
+When the `smoothQualityChange` property is set to `true`, a manual quality
+change triggered via the [representations API](#hlsrepresentations) will use
+smooth quality switching rather than the default fast (buffer-ejecting)
+quality switching. Using smooth quality switching will mean no loading spinner
+will appear during quality switches, but will cause quality switches to only
+be visible after a few seconds.
+
+Note that this _only_ affects quality changes triggered via the representations
+API; automatic quality switches based on available bandwidth will always be
+smooth switches.
 
 ### Runtime Properties
 Runtime properties are attached to the tech object when HLS is in
--- a/src/rendition-mixin.js
+++ b/src/rendition-mixin.js
@@ -48,11 +48,13 @@ const enableFunction = (loader, playlistUri, changePlaylistFn) => (enable) => {
  */
 class Representation {
   constructor(hlsHandler, playlist, id) {
-    // Get a reference to a bound version of fastQualityChange_
-    let fastChangeFunction = hlsHandler
-                              .masterPlaylistController_
-                              .fastQualityChange_
-                              .bind(hlsHandler.masterPlaylistController_);
+    const {
+      masterPlaylistController_: mpc,
+      options_: { smoothQualityChange }
+    } = hlsHandler;
+    // Get a reference to a bound version of the quality change function
+    const changeType = smoothQualityChange ? 'smooth' : 'fast';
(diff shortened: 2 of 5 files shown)


src/rendition-mixin.js
```diff
@@ -48,11 +48,11 @@ const enableFunction = (loader, playlistUri, changePlaylistFn) => (enable) => {
  */
 class Representation {
   constructor(hlsHandler, playlist, id) {
-    // Get a reference to a bound version of fastQualityChange_
-    let fastChangeFunction = hlsHandler
-                              .masterPlaylistController_
-                              .fastQualityChange_
-                              .bind(hlsHandler.masterPlaylistController_);
+    // Get a reference to a bound version of the quality change function
+    let qualityChangeFunction = (hlsHandler.options_.smoothQualityChange ?
```
mjneil: Personally not a fan of large or multi line ternaries, but feel free to ignore this suggestion as its just my preference.
```
const {
  masterPlaylistController_: mpc,
  options_: { smoothQualityChange }
} = hlsHandler;
const changeType = smoothQualityChange ? 'smooth' : 'fast';
const qualityChangeFunction = mpc[`${changeType}QualityChange_`].bind(mpc);
```

test/configuration.test.js
```diff
@@ -28,6 +28,11 @@ const options = [{
   default: 4194304,
   test: 5,
   alt: 555
+}, {
+  name: 'smoothQualityChange',
+  default: undefined,
```
mjneil: should the default be `false` instead of `undefined`?

README.md
```diff
@@ -333,6 +334,22 @@ When `enableLowInitialPlaylist` is set to true, it will be used to select
...
+* can be used as an initialization option
+
+When the `smoothQualityChange` property is set to `true`, a manual quality
+change triggered via the `qualityLevels` API will use smooth quality switching
+rather than the default fast (buffer-ejecting) quality switching. Using smooth
+quality switching will mean no loading spinner will appear during quality
+switches, but will cause quality switches to only be visible after a few
+seconds.
+
+Note that this _only_ affects quality changes triggered via the `qualityLevels`
```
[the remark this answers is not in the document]
squarebracket: Yeah, I was wondering about this myself since it's the representations API from the point of view of VHS, but the `qualityLevels` API from the point of view of the programmer. I will do as you suggest.

src/rendition-mixin.js
squarebracket: Dang, that's much saner. Thanks for the suggestion.

test/configuration.test.js
squarebracket: Good question. [Technically speaking, I don't set the default value to `false`, I just let it be undefined if it's not defined in the options.](https://github.com/videojs/http-streaming/pull/235/files#diff-24fe5b9b63a0e1c36b46a09177d8027fR388) So I chose `undefined` here since that's what the default would be in absence of an explicit value. Do you think I should explicitly set the option to be `false` if a value isn't given, or just change the default here to be `false`?

README.md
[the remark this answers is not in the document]
mjneil: correct, but ONLY if the quality levels plugin is included, which is optional. Manual rendition selection is still possible from the programmer point of view using the representations API. I suggested linking to the readme section since it mentions both the quality levels api (recommended approach) and the representations api

src/videojs-http-streaming.js
```diff
@@ -384,6 +384,9 @@ class HlsHandler extends Component {
       this.tech_.setCurrentTime(time);
       this.setCurrentTime(time);
     };
+    if (this.source_.smoothQualityChange !== undefined) {
+      this.options_.smoothQualityChange = this.source_.smoothQualityChange;
```
mjneil: is this supposed to also be a source level option, or just a player level option?
If source level option is also valid, you should add `smoothQualityChange` to [hereinstead](https://github.com/videojs/http-streaming/blob/master/src/videojs-http-streaming.js#L356)

test/configuration.test.js
mjneil: I think it would be good to set it to false if not set like we do with `withCredentials` [here](https://github.com/videojs/http-streaming/blob/master/src/videojs-http-streaming.js#L337) The explicit setting of the false default should be done here as well.

src/videojs-http-streaming.js
squarebracket: Ah yes thanks, in an older version I was setting the quality change function hereish which required the MPC, and thus couldn't go in `setOptions`. But it can now, great catch.

test/configuration.test.js
squarebracket: Yup, fair enough, explicit is always good.

README.md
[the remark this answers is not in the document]
squarebracket: Yep, covers all bases nicely.

squarebracket:
Made the review changes, thanks for the great comments!! let me know if you think README blob is sufficient.

squarebracket:
Rebased to master and squashed into a single commit.

PR #236, opened 2018-09-20
Adding missing manifest information on to segments (EXT-X-PROGRAM-DATE-TIME)

rart:
## Description
The segment metadata is currently missing the information that comes from the `EXT-X-PROGRAM-DATE-TIME`. 

## Specific Changes proposed
I've added the `dateTimeObject` & `dateTimeString` to the segment data object. The information as already coming through from the buffer, only missing to be added to the segment data object that's _published_.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1371,6 +1371,8 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     const Cue = window.WebKitDataCue || window.VTTCue;
     const value = {
+      dateTimeObject: segment.dateTimeObject || null,
+      dateTimeString: segment.dateTimeString || null,
       bandwidth: segmentInfo.playlist.attributes.BANDWIDTH,
       resolution: segmentInfo.playlist.attributes.RESOLUTION,
       codecs: segmentInfo.playlist.attributes.CODECS,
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -1371,6 +1371,8 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     const Cue = window.WebKitDataCue || window.VTTCue;
     const value = {
+      dateTimeObject: (segment && segment.dateTimeObject) || null,
+      dateTimeString: (segment && segment.dateTimeString) || null,
```
gkatsev: this should probably default to the empty string.

rart: Didn't seem accurate to me. I felt null was a better representation of the value not being there. I can change if `''` is preferred, though.

gkatsev: I think it should either be a string to match the type of that property when a value is present, or, maybe better yet, not have the property at all if a value is absent.

rart: I disagree; but ok :) I'll make the update.

gkatsev: Why do you think having null values would be better?

rart: so you're ok with dateTimeObject being null? You just want dateTimeString to be `''`?

or are we sticking with removing them all together when not present?

forbesjo: @rart I believe what you have is fine, if the program date time tag has no value for some reason (not sure if this is allowed in the HLS spec) I don't think we should have the property be present with an empty string but rather use null.

PR #238, opened 2018-09-21
add an option to ignore player size in selection logic

theRealWardo:
## Description
Some modern encoding techniques target the best quality at a given bitrate, to some extent irrespective of resolution. In such scenarios, users should be able to tell the player that the player size should be ignored during rendition selection in order to favor a pure bandwidth based decision which will maximize quality for the playback. 

## Specific Changes proposed
- add a `limitRenditionByPlayerDimensions ` configuration option which can be set to false to ignore resolutions when doing rendition selection.

## Requirements Checklist
- [x] Feature implemented / Bug fixed - awww yea
  - [x] Unit Tests updated or fixed - wrote some unit tests!
  - [x] Docs/guides updated - wrote docs in the README
  - [x] Example created - https://jsbin.com/gejugat/4/edit?html,output - I mean this is a pretty boring example as I'm just setting the config variable in it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,16 @@ When `enableLowInitialPlaylist` is set to true, it will be used to select
 the lowest bitrate playlist initially.  This helps to decrease playback start time.
 This setting is `false` by default.
 
+##### ignorePlayerSize
+* Type: `boolean`
+* can be used as an initialization option
+
+When `ignorePlayerSize` is set to true, selection logic will ignore the
+player size and rendition resolutions when making a decision.  This should
+be used when doing per-title encoding which makes non-traditional
+rendition resolution decisions.
+This setting is `false` by default.
+
 ### Runtime Properties
 Runtime properties are attached to the tech object when HLS is in
 use. You can get a reference to the HLS source handler like this:
--- a/README.md
+++ b/README.md
@@ -333,13 +333,14 @@ When `enableLowInitialPlaylist` is set to true, it will be used to select
 the lowest bitrate playlist initially.  This helps to decrease playback start time.
 This setting is `false` by default.
 
-##### ignorePlayerSize
+##### limitRenditionByPlayerDimensions
 * Type: `boolean`
 * can be used as an initialization option
 
-When `ignorePlayerSize` is set to true, rendition selection logic will ignore
-the player size and rendition resolutions when making a decision.
-This setting is `false` by default.
+When `limitRenditionByPlayerDimensions` is set to true, rendition
+selection logic will take into account the player size and rendition
+resolutions when making a decision.
+This setting is `true` by default.
 
 ### Runtime Properties
 Runtime properties are attached to the tech object when HLS is in
--- a/README.md
+++ b/README.md
@@ -337,10 +337,8 @@ This setting is `false` by default.
 * Type: `boolean`
 * can be used as an initialization option
 
-When `ignorePlayerSize` is set to true, selection logic will ignore the
-player size and rendition resolutions when making a decision.  This should
-be used when doing per-title encoding which makes non-traditional
-rendition resolution decisions.
+When `ignorePlayerSize` is set to true, rendition selection logic will ignore
+the player size and rendition resolutions when making a decision.
 This setting is `false` by default.
 
 ### Runtime Properties
--- a/src/playlist-selectors.js
+++ b/src/playlist-selectors.js
@@ -124,6 +124,8 @@ export const comparePlaylistResolution = function(left, right) {
  *        Current width of the player element
  * @param {Number} playerHeight
  *        Current height of the player element
+ * @param {Boolean} ignorePlayerSize
+ *        True if the player width and height should be ignored during the selection, false otherwise
  * @return {Playlist} the highest bitrate playlist less than the
  * currently detected bandwidth, accounting for some amount of
  * bandwidth variance
@@ -131,7 +133,8 @@ export const comparePlaylistResolution = function(left, right) {
 export const simpleSelector = function(master,
                                        playerBandwidth,
                                        playerWidth,
-                                       playerHeight) {
+                                       playerHeight,
+                                       ignorePlayerSize) {
   // convert the playlists to an intermediary representation to make comparisons easier
   let sortedPlaylistReps = master.playlists.map((playlist) => {
     let width;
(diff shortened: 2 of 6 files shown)


forbesjo:
Seems reasonable to me, @gesinger thoughts?

src/playlist-selectors.js
```diff
@@ -190,6 +193,16 @@ export const simpleSelector = function(master,
     (rep) => rep.bandwidth === highestRemainingBandwidthRep.bandwidth
   )[0];
 
+  // if we are ignoring the player size, make an early decision
+  if (ignorePlayerSize) {
+    let chosenRep = (
+      bandwidthBestRep ||
+      enabledPlaylistReps[0] ||
+      sortedPlaylistReps[0]
+    );
```
forbesjo: You'll need a newline here to fix linting errors

README.md
```diff
@@ -333,6 +333,16 @@ When `enableLowInitialPlaylist` is set to true, it will be used to select
 the lowest bitrate playlist initially.  This helps to decrease playback start time.
 This setting is `false` by default.
 
+##### ignorePlayerSize
+* Type: `boolean`
+* can be used as an initialization option
+
+When `ignorePlayerSize` is set to true, selection logic will ignore the
```
gesinger: selection logic => rendition selection logic

README.md
```diff
@@ -333,6 +333,16 @@ When `enableLowInitialPlaylist` is set to true, it will be used to select
...
+player size and rendition resolutions when making a decision.  This should
```
gesinger: The should part here is a bit unclear to me, but I may not fully understand the scenario. Any more details on that part?

README.md
```diff
@@ -333,6 +333,16 @@ When `enableLowInitialPlaylist` is set to true, it will be used to select
 the lowest bitrate playlist initially.  This helps to decrease playback start time.
 This setting is `false` by default.
 
+##### ignorePlayerSize
```
gesinger: In terms of naming, I think it might be a bit easier to use a positive and have it be true by default. Also maybe a bit more precise for rendition selection. Maybe something like `limitRenditionSelectionByPlayerDimensions`, though that name could be too long.

src/playlist-selectors.js
theRealWardo: yea so like I got a *ton* of linting errors relating to `exceeds the maximum line length of 90`, `Don't make functions within a loop` and `Unexpected "todo" comment` so I admittedly ignored all of them and just opened a PR... sorry

README.md
theRealWardo: done

README.md
theRealWardo: I am opening this PR because I had a customer using our per-title encoding solution which is marketed at https://mux.com/per-title-encoding/ and described in all kinds of fun technical detail at - https://mux.com/blog/per-title-encoding-scale/ - a specific video was seen to never play anything "reasonably good looking" given the generated rendition set which I put in the unit tests. most people in the industry would look at that ladder and say "what?" so I wrote this comment. but since the comment had a "what?" response, I guess I'll delete it.  😄

src/playlist-selectors.js
forbesjo: No worries, when we update our linter I think it should be more clear which of those is a linter error and just a warning

README.md
forbesjo: Let's go with `limitRenditionByPlayerDimensions`, default to `true`

theRealWardo: I believe I have updated all of the things!

src/playlist-selectors.js
```diff
@@ -190,6 +193,17 @@ export const simpleSelector = function(master,
     (rep) => rep.bandwidth === highestRemainingBandwidthRep.bandwidth
   )[0];
 
+  // if we not going to limit renditions by player size, make an early decision.
```
gesinger: we to we're

theRealWardo: doh. done!

theRealWardo:
oo there seems to be something wrong according to the tests... taking a look...

theRealWardo:
maybe something is wrong in the defaults here, despite the unit tests saying they work, I saw a bunch of behavior that indicated that `limitRenditionByPlayerDimensions` was `undefined`. not sure what to do here as the tests that were failing in CI never pass on my machine even on master.

theRealWardo:
@forbesjo @gesinger - mind taking one last look and merging this if you are comfortable with it please?

src/videojs-http-streaming.js
```diff
@@ -335,6 +332,7 @@ class HlsHandler extends Component {
   setOptions_() {
     // defaults
     this.options_.withCredentials = this.options_.withCredentials || false;
+    this.options_.limitRenditionByPlayerDimensions = this.options_.limitRenditionByPlayerDimensions || true;
```
forbesjo: Wont this always be true?

theRealWardo:
patched the silly mistake that @forbesjo found, fix the test that I misconfigured making be believe it was working correctly when it wasn't, and now I think its ready for another look. thanks!!

src/playlist-selectors.js
```diff
@@ -261,7 +275,8 @@ export const lastBandwidthSelector = function() {
   return simpleSelector(this.playlists.master,
                         this.systemBandwidth,
                         parseInt(safeGetComputedStyle(this.tech_.el(), 'width'), 10),
-                        parseInt(safeGetComputedStyle(this.tech_.el(), 'height'), 10));
+                        parseInt(safeGetComputedStyle(this.tech_.el(), 'height'), 10),
+                        this.limitRenditionByPlayerDimensions);
```
gesinger: Here and below, I don't think this value will ever be set, since `limitRenditionByPlayerDimensions` is set within the `options_` object property and not on the `HlsHandler` instance directly.

theRealWardo: I really should have spent some more time running the code via the debugger to make sure things were getting passed around. I did some more testing using this m3u8 which I created with [Mux per title encoding](https://docs.mux.com/docs/per-title-encoding) - https://stream.mux.com/d3dYsgwvtxk101D6ExK3HuoKe3aps9IIA.m3u8 - you can pretty clearly spot the difference on the test page when setting `limitRenditionByPlayerDimensions: false` if you use this manifest.

this is my PR for this repo so I was all like...



I'm going to update your contribution guidelines with some learnings from this process in another PR

PR #239, opened 2018-09-24
fix(subtitles): set default property if default and autoselect are both enabled

gkatsev:


gkatsev:
Looking at https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.4.1, seems like `propreties.default && properties.autoselect` is the correct way to hint whether something should be chosen.

forbesjo:
LGTM, @gesinger ?

gkatsev:
I can try but I'm not sure I have the time to add tests to this.

forbesjo:
@gkatsev I'm going to add some tests and also make the change for `captions`, unless `default` doesn't apply?

gkatsev:
We can actually tell whether it should be captions or subtitles by which characteristics are enabled. If there are characteristics, particularly the value `public.accessibility.describes-music-and-sound`, it means it's captions and not just subtitles.

forbesjo:
I'm specifically referring to https://github.com/videojs/http-streaming/blob/master/src/media-groups.js#L571

gkatsev:
oh, I see. Yeah, we probably should do that there as well.

PR #257, opened 2018-10-11
feat: add convertToStreamTime method

ldayananda:
## Description
Please describe the change as necessary.
If it's a feature or enhancement please be as detailed as possible.
If it's a bug fix, please link the issue that it fixes or describe the bug in as much detail.

## Specific Changes proposed
Please list the specific changes involved in this pull request.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/util/time.js
+++ b/src/util/time.js
@@ -1,0 +1,76 @@
+/**
+ * @file time.js
+ */
+
+const findSegmentForTime = (time, timeline) => {
+  const matchingSegments = timeline.filter((x) => {
+    return x.start <= time && time <= x.end;
+  });
+
+  if (matchingSegments.length === 0) {
+    return null;
+  }
+
+  return matchingSegments[0].segment;
+};
+
+const buildMediaTimeline = (playlist) => {
+  const timeline = [];
+
+  if (!playlist.segments || playlist.segments.length === 0) {
+    return;
+  }
+
+  // Assumptions:
+  // - there will always be a segment.duration
+  // - we can start from zero
+  // - segments are in time order
+  // - segment.start and segment.end only come
+  //    from syncController(ignored for now)
+
+  let time = 0;
+
+  playlist.segments.forEach((segment) => {
+    const end = time + segment.duration;
+
+    timeline.push({
+      start: time,
+      end,
+      segment
+    });
+
+    time = end;
+  });
+
+  return timeline;
+};
+
+export const getStreamTime = (player, masterPlaylistController) => {
+  return function({
+    time = player.currentTime(),
+    callback
+  } = {}) {
+    const streamTime = {
+      mediaSeconds: time,
+      programDateTime: null
+    };
+
+    const media = masterPlaylistController.media();
+    const timeline = buildMediaTimeline(media);
+
+    if (timeline) {
+      const segment = findSegmentForTime(time, timeline);
+
+      if (segment.dateTimeObject) {
+        // TODO confirm this is YYYY-MM-DDThh:mm:ss.SSSZ ISO_8601 format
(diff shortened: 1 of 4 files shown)


src/util/time.js
```diff
@@ -0,0 +1,76 @@
...
+      segment
+    });
+
+    time = end;
+  });
+
+  return timeline;
+};
+
+export const getStreamTime = (player, masterPlaylistController) => {
```
gesinger: Probably should just pass the playlist here instead of MPC.

src/util/time.js
```diff
@@ -0,0 +1,76 @@
...
+
+  if (!playlist.segments || playlist.segments.length === 0) {
+    return;
+  }
+
+  // Assumptions:
+  // - there will always be a segment.duration
+  // - we can start from zero
+  // - segments are in time order
+  // - segment.start and segment.end only come
```
gesinger: I know this may just be temporary (hence the "ignored for now"), but we are going to have to be careful here. If we haven't seen a segment, thus don't have start and end times, we are making an inaccurate guess, and we may need to surface that info somehow to allow a user to reject the returned time.

src/videojs-http-streaming.js
```diff
@@ -644,6 +645,11 @@ class HlsHandler extends Component {
     }
     super.dispose();
   }
+
+  convertToStreamTime(args) {
+    return getStreamTime(this.player_, this.masterPlaylistController_)
```
gesinger: Any specific reason this is called in this fashion?

src/util/time.js
```diff
@@ -0,0 +1,76 @@
...
+
+    const media = masterPlaylistController.media();
+    const timeline = buildMediaTimeline(media);
+
+    if (timeline) {
+      const segment = findSegmentForTime(time, timeline);
+
+      if (segment.dateTimeObject) {
+        // TODO confirm this is YYYY-MM-DDThh:mm:ss.SSSZ ISO_8601 format
+        streamTime.programDateTime = segment.dateTimeObject.toISOString();
```
gesinger: May want to add a TODO that this will eventually get modified to be adjusted based on time requested

src/util/time.js
ldayananda: I was not sure if the playlist could potentially change? But I guess it's unlikely that that would happen while calling the method, right?

src/util/time.js
ldayananda: Maybe I'm not entirely clear on what you mean here.

The segment.start and segment.end are appended to the playlist after running through sync-controller. I believe we have a later story to take the timestampOffset and discontinuities into account, which I think would take precedent over the segment.start and segment.end times, right?

ldayananda: Ah, yeah. That's a good point too

src/util/time.js
```diff
@@ -0,0 +1,87 @@
...
+      };
+    }
+
+    manifestTime = estimatedEnd;
+  }
+
+  return null;
+};
+
+export const getStreamTime = ({
```
keisans: Is there a reason we are passing `onsuccess` and `onreject` as callbacks instead of returning a promise?

src/videojs-http-streaming.js
```diff
@@ -644,6 +645,16 @@ class HlsHandler extends Component {
     }
     super.dispose();
   }
+
+  convertToStreamTime(time, onsuccess, onreject) {
```
forbesjo: Do we want this to be a node style callback instead?

src/util/time.js
forbesjo: We want this to be compatible with IE and we don't polyfill Promises so we tend not to use them in video.js projects

PR #259, opened 2018-10-13
fix: ensure content loops

### LINKED ISSUE #231: 7.2.0 to 7.2.4 playback does not restart with loop
## Description
Video fails to play after its reaches the end. This also breaks looping.
Briefly describe the issue.

Repro:
http://jsbin.com/kamavaduru/1/edit?html,output

## Sources
Is a certain source or a certain segment affected? please provide a public (accesible over the internet) link to it below.

## Steps to reproduce
1. Start video
1. Let video end(seeking to near end is fine); play button should change to replay

### Expected
Video continues playback at start

### Error output
Spinner appears and does not go away

## Additional Information
A related issue where playback could not be manually restarted by clicking `play` was solved in 1.2.5
https://github.com/videojs/video.js/issues/5436

videojs-http-streaming 1.2.5 (Pinned version for video.js 7.2.3; latest as of 9/14/2018)

### videojs version
7.2.3

### Browsers
* chrome
Others may be affected but chrome is used for baseline sanity checking

### Platforms
* Ubuntu 18.04
* Windows 10

### Other Plugins
None

### Other JavaScript
None(yes in production, no in repro above)

mister-ben:
## Description
Fixes #231, playback stalls if looping

## Specific Changes proposed
On a seek to 0 when loop enabled, call `setCurrentTime(0)` on vhs.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -321,6 +321,7 @@ class HlsHandler extends Component {
       }
     });
 
+    // Handle seeking when looping - middleware doesn't handle this seek event from the tech
     this.on(this.tech_, 'seeking', function() {
       if (this.tech_.seeking() && this.tech_.currentTime() === 0 && this.tech_.player_.loop()) {
         this.setCurrentTime(0);
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -320,6 +320,13 @@ class HlsHandler extends Component {
         this.masterPlaylistController_.smoothQualityChange_();
       }
     });
+
+    this.on(this.tech_, 'seeking', function() {
+      if (this.tech_.seeking() && this.tech_.currentTime() === 0 && this.tech_.player_.loop()) {
+        this.setCurrentTime(0);
+      }
+    });
+
     this.on(this.tech_, 'error', function() {
       if (this.masterPlaylistController_) {
         this.masterPlaylistController_.pauseLoading();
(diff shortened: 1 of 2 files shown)


src/videojs-http-streaming.js
```diff
@@ -320,6 +320,13 @@ class HlsHandler extends Component {
         this.masterPlaylistController_.smoothQualityChange_();
       }
     });
+
+    this.on(this.tech_, 'seeking', function() {
```
gesinger: It may be worth us adding a comment essentially saying that the middleware can't catch looping behavior because the tech fires a seeking event and we don't get that through the middleware, thus we listen for that specific case.

test/playback.test.js
```diff
@@ -165,3 +165,24 @@ QUnit.test('Live DASH', function(assert) {
...
+  let done = assert.async();
+  let player = this.player;
+
+  player.loop(true);
+  player.src({
+    src: 'http://d2zihajmogu5jn.cloudfront.net/bipbop-advanced/bipbop_16x9_variant.m3u8',
+    type: 'application/x-mpegURL'
+  });
+  player.one('playing', function() {
+    player.vhs.mediaSource.on('sourceopen', () => {
```
gesinger: Is it possible for the playing event to fire before the media source is opened? If so, it may be best for us to add one more listener (for ended), then check for sourceopen after that.

PR #263, opened 2018-10-17
fix: immediately setup EME if available

forbesjo:
## Description
Uses a new videojs-contrib-eme API to set up EME immediately and not wait for `encrypted`
https://github.com/videojs/videojs-contrib-eme/pull/61

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -180,6 +180,10 @@ const setupEmeOptions = (hlsHandler) => {
 
     if (sourceOptions) {
       player.currentSource().keySystems = sourceOptions;
+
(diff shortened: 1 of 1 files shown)


src/videojs-http-streaming.js
```diff
@@ -180,6 +180,10 @@ const setupEmeOptions = (hlsHandler) => {
 
     if (sourceOptions) {
       player.currentSource().keySystems = sourceOptions;
+
+      if (player.eme.initializeMediaKeys) {
```
gesinger: Might be worth it to add a comment pointing to the issue that we noticed.

Issue #112, opened 2018-06-22
Update package.json description


forbesjo:
## Description
Update https://github.com/videojs/http-streaming/blob/master/package.json#L4 so that it mentions DASH and future HTTP segmented content

PR #267, opened 2018-10-22
docs: add dash to package.json description

mjneil:
address #112


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@videojs/http-streaming",
   "version": "1.4.0",
-  "description": "Play back HLS with video.js, even where it's not natively supported",
+  "description": "Play back HLS and DASH with video.js, even where it's not natively supported",
```
gkatsev: we should capitalize Video.js

PR #270, opened 2018-10-24
Add unit tests and closed-caption autoselect to fix(subtitles) PR

mrocajr:


src/media-groups.js
```diff
@@ -568,7 +568,7 @@ export const initialize = {
           const track = tech.addRemoteTextTrack({
             id: properties.instreamId,
             kind: 'captions',
-            enabled: false,
+            enabled: properties.default && properties.autoselect,
```
gkatsev: the property name should be `default`

test/media-groups.test.js
```diff
@@ -726,6 +726,7 @@ function(assert) {
   assert.ok(this.mediaTypes[type].groups.sub2[1].playlistLoader,
     'playlistLoader created');
   assert.ok(this.mediaTypes[type].tracks.en, 'created text track');
+  assert.equal(this.mediaTypes[type].tracks.en.default, true, 'en track auto selected');
```
gesinger: This seems to be testing that the default property is set, rather than it being auto selected.

gkatsev: I think that's fine? Actually selecting the caption by default is the job of Video.js, so, there's no need to test that behavior here (or override it). But we do need to make sure that the signal is passed to Video.js properly.

src/media-groups.js
mrocajr: properties.default as opposed to (properties.default && properties.autoselect)?

gkatsev: instead of `enabled`, see https://github.com/videojs/http-streaming/pull/239/files#diff-46e43d417013a9cfb72f6b3210979b63R515

PR #272, opened 2018-10-25
feat: Transmux Before Append and LHLS

gesinger:
Includes work by: @mjneil @ldayananda @gesinger

This PR refactors our segment handling logic by transmuxing a segment before appending it to the source buffer, and adds support for LHLS (low latency HLS) in the process.

This allows us to remove the legacy videojs-contrib-media-sources code where we faked the source buffers to allow us to append TS segments directly (and transmuxed within the fake source buffer). There are many reasons we chose to do this, and for a more detailed look into why we went this route, please see the docs dir in this branch: https://github.com/gesinger/http-streaming/blob/transmux-before-append-lhls/docs/lhls/index.md .

Most of the changes are in test files and package-lock.json, however, there is a good deal of refactoring throughout a large part of the codebase. In addition, some legacy code was cleaned up, and some bug fixes were made.

## Portions Reviewed:

Due to the size of this PR, it may make the most sense to review in portions. From individual modules to logical paths.

#### Docs, linting, and scripts
- [x] eslint changes (extending videojs' standard, overriding for tests)
- [x] docs/creating-content.md - sample ffmpeg commands for creating (and created) content
- [x] docs/lhls/index.md - overview of transmux before append and LHLS work
- [x] docs/lhls/transmux-before-append-changes.md
- [x] scripts/segments-data.js - script changes for real segments in tests
- [x] scripts/test.rollup.config.js - script changes for async/await in tests

#### Modules
_and associated tests_
- [x] src/master-playlist-controller.js
- [x] src/media-groups.js
- [x] src/media-segment-request.js
- [x] src/playback-watcher.js
- [x] src/segment-loader.js
- [x] src/segment-transmuxer.js
- [x] src/source-updater.js
- [x] src/sync-controller.js
- [x] src/transmuxer-worker.js
- [x] src/util/codecs.js
- [x] src/util/segment.js
- [x] src/util/string-to-array-buffer.js
- [x] src/util/text-tracks.js
- [x] src/videojs-http-streaming.js
- [x] src/vtt-segment-loader.js

#### Logic Paths and Areas of Focus
_and associated tests_
- [x] (real) media source and source buffer creation (largely in master-playlist-controller)
- [x] duration updates for live and VOD (largely in master-playlist-controller)
- [ ] handling partial data (largely in media-segment-request) (including captions and ID3 tags)
- [x] removed code from videojs-contrib-media-sources
- [ ] captions (largely in segment-loader)
- [ ] id3 metadata (largely in segment-loader)
- [ ] WebWorker creation, management, and cleanup (largely in segment-loader)
- [ ] discontinuities/timestamp offsets (largely in segment-loader)
- [ ] gopsToAlignWith
- [ ] segment-loader queueing
- [ ] segment start and end times (largely in segment-loader)
- [ ] source-updater queueing

## Known Issues
- Alt audio with handlePartialData set to true
- Alt audio, when offset from a 0 start time, sometimes causes video to forever buffer. This happens when the audio downloads and appends before the video has time to set the timestamp offset on the source buffer (therefore, audio will remain offset like the content itself rather than start at 0). Keeping in mind discontinuities/timelines, the audio segment loader will need to wait for video to update the source buffer's timestamp offset (if necessary) within a timeline before appending its own content.
- Muxed fmp4 content (HLS) that doesn't provide codec info in the master manifest doesn't play. This is due to us not parsing the fmp4 segments, and assuming that content is demuxed if using fmp4. We will need to parse the fmp4 segments and not assume demuxed in order to resolve this issue.
- Doubled captions when entering fullscreen. This also happens when switching renditions. It appears the caption timestamps we get back from mux.js are different for different renditions(but the same caption text). It just so happens when entering fullscreen, because we use `keepOriginalTimestamps: true` for this branch, the caption timestamps align and are shown at the same time. Both on master and when doing rendition switches, the captions are actually added twice, but with very different timestamps so they are not displayed together. It remains to be investigated why we are getting these incorrect timestamps back from mux.js in the first place.

## Explicit Areas to Test
- [ ] Not appending video init segment for every segment. From @mjneil: "I think we tried this in the past, but had to leave it as appending video init for every segment cause of a browser issue."


docs/lhls/transmux-before-append-changes.md
```diff
@@ -0,0 +1,36 @@
+# Transmux Before Append Changes
+
+## Overview
+
+In moving our transmuxing stage from after append (to a virtual source buffer from videojs-contrib-media-sources) to before appending (to a native source buffer), some changes were required, and others made the logic simpler. What follows are some details into some of the changes made, why they were made, and what impact they will have.
+
+### Source Buffer Creation
+
+In a pre-TBA world, videojs-contrib-media-source's source buffers provided an abstraction around the native source buffers. They also required a bit more information than the native buffers. For instance, they used the full mime types instead of simply relying on the codec information, when creating the source buffers. This provided the container types, which let the virtual source buffer know whether the media needed to be transmuxed or not. In a post-TBA world, the container type is no longer required, therefore only the codec strings are passed along.
```
mjneil: Would be good to explicitly call out what `TBA` means on first use here

docs/lhls/transmux-before-append-changes.md
```diff
@@ -0,0 +1,36 @@
...
+    * after timestampOffset is set
+    * audio track events: change/addtrack/removetrack
+    * 'mediachange' event
+
+segment-loader and FMP4:
+* if segment.map is set:
+  * save (cache) the init segment after the request finished
+  * append the init segment directly to the source buffer if the segment loader's activeInitSegmentId doesn't match the segment.map generated init segment ID
+
+With the transmux before append and LHLS changes, we only append video init segments on changes as well. This is more important with LHLS, as prepending an init segment before every frame of video would be wasteful.
```
mjneil: Has there been any regression testing with this? I think we tried this in the past, but had to leave it as appending video init for every segment cause of a browser issue.

package.json
```diff
@@ -22,7 +22,7 @@
     "docs": "run-p docs:*",
     "docs:api": "jsdoc src -r -d docs/api",
     "docs:toc": "doctoc README.md",
-    "lint": "vjsstandard",
+    "lint": "eslint .",
```
mjneil: Why no longer videojs-standard dependency? you can still use .eslintrc files with videojs-standard to override things

src/media-segment-request.js
```diff
@@ -235,59 +248,235 @@ const handleSegmentResponse = (segment, captionParser, finishProcessingFn) => (e
...
+  doneFn
+}) => {
+  const bytesAsUint8Array = new Uint8Array(bytes);
+
+  segment.isFmp4 =
+    // only set the property the first time we see some bytes so that partial appends
+    // don't try to check every section of bytes (since the check should only consider the
+    // first bytes in the segment)
+    typeof segment.isFmp4 === 'boolean' ? segment.isFmp4 :
+      isLikelyFmp4Data(bytesAsUint8Array);
```
ldayananda: We should use `isLikelyFmp4Data` in all the places this is used

https://github.com/videojs/http-streaming/blob/de5aacb782f3f85754b77a7cd34c4e8a6e12ebcf/src/playlist.js#L478-L484

src/master-playlist-controller.js
```diff
@@ -93,10 +98,16 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.mediaTypes_ = createMediaTypes();
 
-    this.mediaSource = new videojs.MediaSource();
+    this.mediaSource = new window.MediaSource();
```
mjneil: 🎉

src/media-segment-request.js
```diff
@@ -235,59 +248,235 @@ const handleSegmentResponse = (segment, captionParser, finishProcessingFn) => (e
...
+  // Check to see if we are appending a full segment.
+  if (!isPartial && !segment.lastReachedChar) {
+    // In the full segment transmuxer, we don't yet have the ability to extract a "proper"
+    // start time. Meaning cached frame data may corrupt our notion of where this segment
+    // really starts. To get around this, full segment appends should probe for the info
+    // needed.
+    const probeResult = probeTsSegment(bytes, segment.baseStartTime);
+
+    if (probeResult) {
+      trackInfoFn(segment, {
```
mjneil: You could probably just pass `probeResult` here `trackInfoFn(segment, probeResult);`

src/playback-watcher.js
```diff
@@ -213,6 +216,7 @@ export default class PlaybackWatcher {
 
       this.logger_(`Stopped at ${currentTime} while inside a buffered region ` +
         `[${currentRange.start(0)} -> ${currentRange.end(0)}]. Attempting to resume ` +
+        // eslint-disable-next-line quotes
```
mjneil: why not just change the next line to use `'` since its not using any features of templates?

forbesjo:
I'm seeing duplicate 608 captions when switching to fullscreen
screen shot 2018-11-13 at 12 39 03 pm

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
+import document from 'global/document';
+import window from 'global/window';
 import QUnit from 'qunit';
 import videojs from 'video.js';
-import { useFakeMediaSource } from './test-helpers';
+import SourceUpdater from '../src/source-updater';
+import { mp4Video, mp4Audio } from './test-segments';
 
 QUnit.module('Source Updater', {
   beforeEach() {
```
ldayananda: seems like this should be a `before` instead(and we need a `after` to clean up)? or we need a `afterEach` to clean up the media source and video element created before each test

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
+    'sourceopen',
+    'listening on sourceopen');
+  assert.equal(
+    mockMediaSource.addSourceBufferCalls.length, 0, 'no add source buffer calls');
+
+  mockMediaSource.readyState = 'open';
+  mockMediaSource.addEventListenerCalls[0].callback();
+
+  assert.equal(
+    mockMediaSource.addEventListenerCalls.length, 1, 'one event listener');
```
ldayananda: do we ever clean up the `sourceopen` listener before disposing the sourceUpdater?

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
+
+  assert.equal(this.sourceUpdater.audioBuffered().length, 0, 'no buffered time range');
+
+  this.sourceUpdater.createSourceBuffers({
+    audio: 'mp4a.40.2'
+  });
+
+  assert.equal(this.sourceUpdater.audioBuffered().length, 0, 'no buffered time range');
+
+  this.sourceUpdater.appendBuffer('audio', mp4Audio(), () => {
```
ldayananda: This is awesome 🔥

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
-  // create other media source
-  this.mediaSource.addSourceBuffer('audio/mp2t');
+  let executedCallback = false;
+  let appendedAudio = false;
 
-  let updater = new SourceUpdater(this.mediaSource, 'video/mp2t', sourceBufferEmitter);
+  this.sourceUpdater.appendBuffer('audio', mp4Audio(), () => {
+    appendedAudio = true;
+    assert.notOk(executedCallback, 'haven\'t executed callback');
+    setTimeout(() => {
```
ldayananda: is there another way to ensure that the queue cleared instead of a timeout?

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
 
-  updater.appendBuffer(new Uint8Array([0, 1, 2]));
+  assert.notOk(appendedAudio, 'haven\'t appended audio before callback is queued');
 
-  sourceBuffer = this.mediaSource.sourceBuffers[1];
-  assert.equal(sourceBuffer.updates_.length, 1, 'called the source buffer once');
-  assert.deepEqual(sourceBuffer.updates_[0].append, new Uint8Array([0, 1, 2]),
-                  'appended the bytes');
+  this.sourceUpdater.audioQueueCallback(() => {
+    executedCallback = true;
```
ldayananda: could you just call `done` here?

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
-  // create other media source
-  this.mediaSource.addSourceBuffer('audio/mp2t');
-  sourceBufferEmitter.trigger('sourcebufferadded');
+  assert.notOk(appendedVideo, 'haven\'t appended video before callback is queued');
 
-  assert.equal(sourceBuffer.updates_.length, 1, 'called the source buffer once');
-  assert.deepEqual(sourceBuffer.updates_[0].append, new Uint8Array([0, 1, 2]),
-                  'appended the bytes');
+  this.sourceUpdater.videoQueueCallback(() => {
+    executedCallback = true;
```
ldayananda: and just call `done` here instead as well

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
+  const done = assert.async();
 
-  this.mediaSource.trigger('sourceopen');
-  sourceBuffer = this.mediaSource.sourceBuffers[0];
-  updater.appendBuffer(new Uint8Array([0, 1, 2]), function() {
-    updateends++;
+  // Source buffer must exist for the callback to run. This case isn't tested, as it isn't
+  // required behavior (at the moment), but is necessary to know for this test.
+  this.sourceUpdater.createSourceBuffers({
+    audio: 'mp4a.40.2'
```
ldayananda: seems like we should create a video sourceBuffer as well to really test this case

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
+    }, 0);
   });
 
-  assert.equal(updateends, 0, 'no completions yet');
-  sourceBuffer.trigger('updateend');
-  assert.equal(updateends, 1, 'ran the completion callback');
+  assert.notOk(appendedAudio, 'haven\'t appended audio before callback is queued');
+
+  this.sourceUpdater.videoQueueCallback(() => {
+    executedCallback = true;
```
ldayananda: this is what we _don't_ want to happen, right? Is the test name incorrect?

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
 
-  updater.appendBuffer(new Uint8Array([2, 3, 4]));
-  assert.equal(sourceBuffer.updates_.length, 1, 'delayed the update');
+  assert.notOk(appendedVideo, 'haven\'t appended video before callback is queued');
 
-  sourceBuffer.trigger('updateend');
-  assert.equal(sourceBuffer.updates_.length, 2, 'updated twice');
-  assert.deepEqual(sourceBuffer.updates_[1].append, new Uint8Array([2, 3, 4]),
-                  'appended the bytes');
+  this.sourceUpdater.audioQueueCallback(() => {
```
ldayananda: similar confusion as the last test

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
+
+  assert.notOk(this.sourceUpdater.updating(), 'not updating by default');
+
+  this.sourceUpdater.appendBuffer('video', mp4Video(), () => {
+    assert.notOk(this.sourceUpdater.updating(), 'not updating after append');
+    this.sourceUpdater.appendBuffer('audio', mp4Audio(), () => {
+      assert.notOk(this.sourceUpdater.updating(), 'not updating after append');
+      done();
+    });
+    assert.ok(this.sourceUpdater.updating(), 'updating during append');
```
ldayananda: maybe `updating during audio append`?

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
+  this.sourceUpdater.dispose();
+
+  assert.ok(abortedAudio, 'aborted audio');
+  assert.ok(abortedVideo, 'aborted video');
+  assert.notOk(this.sourceUpdater.audioBuffer, 'removed audioBuffer reference');
+  assert.notOk(this.sourceUpdater.videoBuffer, 'removed videoBuffer reference');
+});
+
+QUnit.test('no error passed by default in done callback', function(assert) {
+  const done = assert.async();
```
ldayananda: you don't have errors after the dispose test runs?

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
+
+  this.sourceUpdater.appendBuffer('audio', mp4Audio(), (error) => {
+    assert.notOk(error, 'no error');
+    done();
+  });
+});
+
+// Firefox allows appending video to an audio buffer, for some reason.
+// TODO find an alternative for triggering an error
+QUnit.skip('audio source buffer error passed in done callback', function(assert) {
```
ldayananda: maybe append some corrupted data instead of actual video/audio data?

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
 
-  updater.appendBuffer(new Uint8Array([0]));
+  const checkDuration = () => {
+    // duration is set to infinity if content is appended before an explicit duration is
+    // set https://w3c.github.io/media-source/#sourcebuffer-init-segment-received
+    assert.equal(this.mediaSource.duration, Infinity, 'duration not set on media source');
+  };
 
-  updater.abort();
+  this.sourceUpdater.appendBuffer('audio', mp4Audio(), checkDuration);
```
ldayananda: might be a bit easier to follow if an anonymous function was used here instead

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
-  updater.appendBuffer(new Uint8Array([0]));
+  const checkDuration = () => {
+    // duration is set to infinity if content is appended before an explicit duration is
+    // set https://w3c.github.io/media-source/#sourcebuffer-init-segment-received
+    assert.equal(this.mediaSource.duration, Infinity, 'duration not set on media source');
+  };
 
-  updater.remove(1, 14);
+  this.sourceUpdater.appendBuffer('video', mp4Video(), checkDuration);
+  this.sourceUpdater.appendBuffer('audio', mp4Audio(), checkDuration);
```
ldayananda: might want to try to setDuration while either video or audio is still updating

docs/lhls/transmux-before-append-changes.md
gesinger: Besides our general tests, I'm not sure if we covered any content that we know had issues in the past. We may have to dig some up. Added note to the description as an area to test.

src/media-segment-request.js
gesinger: I think I was avoiding passing the whole object so we could me more explicit about which parameters we are passing, but open to the alternative if you prefer it.

src/source-updater.js
```diff
@@ -13,117 +163,147 @@ import noop from './util/noop';
...
+      videojs.createTimeRange();
+  }
+
+  videoBuffered() {
+    return this.videoBuffer && this.videoBuffer.buffered ? this.videoBuffer.buffered :
+      videojs.createTimeRange();
+  }
+
+  buffered() {
+    return buffered(this.videoBuffer, this.audioBuffer);
```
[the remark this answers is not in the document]
gesinger: From utils, imported at top

src/source-updater.js
```diff
@@ -13,117 +163,147 @@ import noop from './util/noop';
...
-  abort(done) {
-    if (this.processedAppend_) {
-      this.queueCallback_(() => {
-        this.sourceBuffer_.abort();
-      }, done);
+    if (this.audioBuffer) {
+      this.onAudioUpdateEnd_ = onUpdateend('audio', this);
+      this.audioBuffer.addEventListener('updateend', this.onAudioUpdateEnd_);
+      this.onAudioError_ = (e) => {
+        this.audioError_ = e;
```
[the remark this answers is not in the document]
gesinger: Only for debugging purposes. I can add a comment, or remove them if you think they're not necessary.

src/util/codecs.js
```diff
@@ -146,58 +122,65 @@ const audioProfileFromDefault = (master, audioGroupId) => {
   return null;
 };
 
+export const isMaat = (master, media) => {
+  if (!media) {
+    // Not enough information
+    return false;
```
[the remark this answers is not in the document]
gesinger: Doesn't look like it anymore, will remove 👍

src/util/codecs.js
```diff
@@ -146,58 +122,65 @@ const audioProfileFromDefault = (master, audioGroupId) => {
...
+  const mediaAttributes = media.attributes || {};
+
+  return master.mediaGroups.AUDIO &&
+    mediaAttributes.AUDIO &&
+    master.mediaGroups.AUDIO[mediaAttributes.AUDIO];
+};
+
+export const isMuxed = (master, media) => {
+  if (!isMaat(master, media)) {
+    return true;
```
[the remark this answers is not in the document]
gesinger: Audio only wouldn't be MAAT unless it also has alternate audio tracks. MAAT is just defined as having AUDIO media groups with alt audio there.

src/util/codecs.js
```diff
@@ -212,113 +195,33 @@ export const mimeTypesForPlaylist = function(master, media) {
...
-    // mime-type for the first entry in the array even when there is only audio.
-    // This doesn't appear to be a problem and simplifies the code.
-    return [
-      bothVideoAudio,
-      justAudio
-    ];
+export const isLikelyFmp4Data = (bytes) => {
+  // not enough data to determine, in which case it is an invalid mp4 file/fragment anyway
+  if (bytes.length < 8) {
+    return false;
```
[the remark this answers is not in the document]
gesinger: Probably not, as it should be a normal (albeit rare) case for LHLS.

test/util/string-to-array-buffer.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+QUnit.test('array buffer created from string contains the correct codes',
+function(assert) {
+  const text = 'test';
+  const arrayBuffer = stringToArrayBuffer(text);
+  const view = new Uint8Array(arrayBuffer);
+
+  assert.strictEqual(
+    typeof arrayBuffer,
+    typeof new ArrayBuffer(0),
```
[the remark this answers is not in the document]
gesinger: Typed arrays are just views into ArrayBuffers.

src/vtt-segment-loader.js
```diff
@@ -220,24 +229,62 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
+      this.state = 'READY';
+      return;
+    }
+
+    this.saveTransferStats_(simpleSegment.stats);
+
+    // the request was aborted
+    if (!this.pendingSegment_) {
       this.state = 'READY';
+      this.mediaRequestsAborted += 1;
```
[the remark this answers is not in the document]
gesinger: Yeah, seems like it.

src/util/segment.js
```diff
@@ -0,0 +1,74 @@
+import mp4probe from 'mux.js/lib/mp4/probe';
+import tsInspector from 'mux.js/lib/tools/ts-inspector.js';
+
+const ONE_SECOND_IN_TS = 90000;
```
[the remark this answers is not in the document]
gesinger: We can, but I do like seeing the number here without digging through another project to find it. Open to either though.

src/util/segment.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+export const probeTsSegment = (bytes, baseStartTime) => {
+  let timeInfo = tsInspector.inspect(bytes, baseStartTime * ONE_SECOND_IN_TS);
+
+  if (!timeInfo) {
+    return null;
+  }
+
+  const result = {
+    hasVideo: timeInfo.video && timeInfo.video.length === 2,
```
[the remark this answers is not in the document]
gesinger: Start and end times. I'll add a comment to clarify.

test/util/segment.test.js
```diff
@@ -0,0 +1,28 @@
+import QUnit from 'qunit';
+import { concatSegments } from '../../src/util/segment';
+
+QUnit.module('util/segment');
```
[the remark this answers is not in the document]
gesinger: I believe that's been that way for a while. We should add tests to it, but right now it's maintaining the status quo.

test/source-updater.test.js
```diff
@@ -1,232 +1,769 @@
...
+    const video = document.createElement('video');
+
+    this.mediaSource = new window.MediaSource();
+    // need to attach the real media source to a video element for the media source to
+    // change to an open ready state
+    video.src = URL.createObjectURL(this.mediaSource);
+    this.sourceUpdater = new SourceUpdater(this.mediaSource);
+
+    // wait for the source to open (or error) before running through tests
+    return new Promise((accept, reject) => {
```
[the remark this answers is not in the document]
gesinger: Each test should wait to start until that promise resolves. And if it does reject then it's good that it'll fail (since the test probably can't run without an open media source).

test/util/codecs.test.js
```diff
@@ -0,0 +1,350 @@
...
+    {
+      audio: 'mp4a.40.2',
+      video: 'avc1.4d400d'
+    },
+    'MAAT, demuxed, codecs: none');
+
+  assert.deepEqual(
+    codecsForPlaylist.apply(null, generateMedia(true, false, true, false, isFMP4)),
+    {
+      audio: 'mp4a.40.2',
```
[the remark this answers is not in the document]
gesinger: Since it is MAAT, there should be audio. But only the video codec is included in the manifest, so we have to ensure that the audio codec is also added (we use a default one).

test/util/codecs.test.js
```diff
@@ -0,0 +1,350 @@
...
+      video: 'avc1.4d400d'
+    },
+    'MAAT, muxed, codecs: none');
+
+  assert.deepEqual(
+    codecsForPlaylist.apply(null, generateMedia(true, true, true, false, isFMP4)),
+    {
+      audio: 'mp4a.40.2',
```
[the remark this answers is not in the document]
gesinger: Same re: above.

test/util/codecs.test.js
```diff
@@ -0,0 +1,350 @@
...
+    'muxed if no URI and no playlists');
+
+  master.mediaGroups.AUDIO.test.demuxed.playlists = [{}];
+  assert.deepEqual(
+    codecsForPlaylist(master, media),
+    {
+      audio: 'mp4a.40.E',
+      video: 'avc1.deadbeef'
+    },
+    'demuxed if no URI but playlists');
```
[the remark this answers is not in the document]
gesinger: Adding comments to help clear up why the results are expected. Once up, let me know if they can be improved.

test/source-updater.test.js
gesinger: Added an `afterEach` to clean up the media source. It's a `beforeEach` instead of a `before` so we have a clean instance of source updater (and source buffers) on each test.

test/source-updater.test.js
gesinger: We're only using a timeout here to ensure we can run something right after the callback finishes and is marked as called. Unless we add another event, I think we have to use the timeout.

test/source-updater.test.js
gesinger: We just want to make sure that the queue processes the callback we set up at the end of the test after processing prior queue items.

test/source-updater.test.js
gesinger: Same re: above.

test/source-updater.test.js
gesinger: Yup, test name should be correct, though I can change the variable name to `executedVideoCallback` to be more precise.

test/source-updater.test.js
gesinger: Same re: above

test/source-updater.test.js
gesinger: Each source updater is a new instance, since it is created in `beforeEach`

test/source-updater.test.js
gesinger: I just didn't want to copy the comment and code inside of the function for both video and audio, but am open to changing if you think the copying makes it easier to read.

src/source-updater.js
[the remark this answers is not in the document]
ldayananda: We could probably debug log this

src/util/segment.js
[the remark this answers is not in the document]
ldayananda: I think it's best to keep them in sync by having the clock values in one place

test/util/string-to-array-buffer.test.js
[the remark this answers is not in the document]
ldayananda: maybe the assert name should be changed then?

test/source-updater.test.js
ldayananda: has this been pushed up already?

test/source-updater.test.js
ldayananda: Could we also use the sinon clock?

test/source-updater.test.js
ldayananda: I think it'll make it easier to read

test/util/string-to-array-buffer.test.js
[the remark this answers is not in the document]
gesinger: good call 👍

test/source-updater.test.js
gesinger: Yup. https://github.com/videojs/http-streaming/pull/272/commits/526bc0931705b6d433531c4fb0a2d588c700a548

test/source-updater.test.js
gesinger: Added in https://github.com/videojs/http-streaming/pull/272/commits/3641cf6ec52129245935561cb772747aa61c88b0

test/source-updater.test.js
ldayananda: Might want to also clean up `this.mediaSource` in the `afterEach` to make sure the old ones die

test/source-updater.test.js
gesinger: Good call, that worked.

test/source-updater.test.js
gesinger: I was looking at a different test where we reused it. Definitely makes sense as anonymous here 👍

gesinger: (Possibly there as well.)

gkatsev:
There's work to resolve conflicts in [a PR](https://github.com/gesinger/http-streaming/pull/10) but it requires a bit more work than just merging that PR, so, we should use it as inspiration. We also have plans to slowly work through this and review and then merge and fix forward any issues.

gkatsev:
further work will be happening in #466

PR #277, opened 2018-10-30
Throttle final playlist reloads when using DASH

gesinger:
## Description
HLS already throttles reloads of the final playlist, however, DASH does not. This adds the throttling behavior to the DASH playlist loader.

To see the behavior:
1) Go to https://videojs.github.io/http-streaming/
2) Enter the source as: https://dash.akamaized.net/akamai/bbb_30fps/bbb_30fps.mpd and the mime type as: application/dash+xml
3) Click play
4) Open the console and go to the network tab
5) Click "Offline"

The failed requests will shoot up as fast as can be processed (since the retries are immediate).

The fix makes it so that retries are only made at half the target duration.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/test/dash-playlist-loader.test.js
+++ b/test/dash-playlist-loader.test.js
@@ -397,3 +397,41 @@ QUnit.test('media playlists "refresh" by re-parsing master xml', function(assert
 
   assert.equal(refreshes, 1, 'refreshed playlist after last segment duration');
 });
+
+QUnit.test('delays load when on final rendition', function(assert) {
+  let loader = new DashPlaylistLoader('dash.mpd', this.fakeHls);
+  let loadedplaylistEvents = 0;
+
+  loader.on('loadedplaylist', () => loadedplaylistEvents++);
+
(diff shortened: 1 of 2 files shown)


test/dash-playlist-loader.test.js
```diff
@@ -397,3 +397,41 @@ QUnit.test('media playlists "refresh" by re-parsing master xml', function(assert
...
+  let loader = new DashPlaylistLoader('dash.mpd', this.fakeHls);
+  let loadedplaylistEvents = 0;
+
+  loader.on('loadedplaylist', () => loadedplaylistEvents++);
+
+  // do an initial load to start the loader
+  loader.load();
+  standardXHRResponse(this.requests.shift());
+
+  assert.equal(loadedplaylistEvents, 2, 'two loadedplaylist events after first load');
```
ldayananda: might be useful to have a comment here explaining where the 2 loadedplaylist events come from

PR #278, opened 2018-10-31
[WIP] chore: switch to videojs-generate-karma-config

ldayananda:


ldayananda:
looks like resolving the conflict didn't work out when I changed the source branch to be on the main remote. Replaced by #283

PR #283, opened 2018-11-01
[WIP] chore: switch to videojs-generate-karma-config

ldayananda:


package.json
```diff
@@ -89,20 +89,11 @@
...
-    "karma-qunit": "^1.2.1",
-    "karma-safari-launcher": "^1.0.0",
-    "karma-safaritechpreview-launcher": "0.0.6",
     "karma-spec-reporter": "0.0.32",
     "lodash": "^4.17.4",
-    "lodash-compat": "^3.10.0",
     "npm-run-all": "^4.1.2",
     "portscanner": "^2.1.1",
-    "qunitjs": "^2.0.1",
+    "qunit": "^2.0.1",
```
ldayananda: package name was changed after 2.4.1 https://www.npmjs.com/package/qunitjs

package.json
```diff
@@ -89,20 +89,11 @@
...
-    "karma-chrome-launcher": "^2.2.0",
-    "karma-coverage": "^1.1.1",
-    "karma-detect-browsers": "^2.2.6",
-    "karma-firefox-launcher": "^1.1.0",
-    "karma-qunit": "^1.2.1",
-    "karma-safari-launcher": "^1.0.0",
-    "karma-safaritechpreview-launcher": "0.0.6",
     "karma-spec-reporter": "0.0.32",
     "lodash": "^4.17.4",
-    "lodash-compat": "^3.10.0",
```
ldayananda: this didn't look like it was used by anything

package.json
```diff
@@ -114,10 +105,11 @@
     "semver": "^5.5.0",
     "serve-static": "^1.13.2",
     "shelljs": "^0.8.2",
-    "sinon": "^1.10.3",
+    "sinon": "1.10.3",
```
ldayananda: a sinon update broke the tests, so I locked to a specific version

test/master-playlist-controller.test.js
```diff
@@ -2285,6 +2285,9 @@ QUnit.test('switches off subtitles on subtitle errors', function(assert) {
   assert.equal(textTracks[1].kind, 'subtitles', 'kind is subtitles');
   textTracks[1].mode = 'showing';
 
+  // Wait for VTT segment to be requested
+  this.clock.tick(1);
```
ldayananda: these were the changes required for #226

package.json
```diff
@@ -89,20 +89,11 @@
...
-    "karma-qunit": "^1.2.1",
-    "karma-safari-launcher": "^1.0.0",
-    "karma-safaritechpreview-launcher": "0.0.6",
     "karma-spec-reporter": "0.0.32",
     "lodash": "^4.17.4",
-    "lodash-compat": "^3.10.0",
     "npm-run-all": "^4.1.2",
     "portscanner": "^2.1.1",
-    "qunitjs": "^2.0.1",
+    "qunit": "2.4.1",
```
forbesjo: `videojs-generate-karma-config` comes with its own `qunit` https://github.com/videojs/videojs-generate-karma-config/blob/master/package.json#L49. Instead of serving the `test/index.html` the other projects start up a karma server, see https://github.com/videojs/mpd-parser/pull/28/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R21 . cc @BrandonOCasey

ldayananda: I noticed that video.js [still has qunit](https://github.com/videojs/video.js/blob/master/package.json#L119) so I left it in. 

Also versions above 2.4.1 broke the tests in some non-obvious way so I wanted to lock the version down to 2.4.1. See:  https://travis-ci.org/videojs/http-streaming/builds/449537117. I tried ensuring that muted was on the video element, and running Chrome with the autoplay-policy set to `no-user-gesture-required` but neither made a difference. 

That said, maybe using qunit's test page rather than `test/index.html` will fix things. I'll give it a shot

gkatsev: Yeah, not sure why we still have qunit in Video.js directly since the karma-config project includes it.

It looks like the tests that failed in that travis run may be autoplaying players that weren't muted? That's what it sounds like from the message at least.

ldayananda: I think the messages were red herrings since disabling the autoplay policy via flags made no difference

ldayananda: Looks like qunit is required at the top level because of karma. Without it:

>>>> karma start scripts/karma.conf.js

module.js:550
    throw err;
    ^

Error: Cannot find module 'qunit'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.resolve (internal/module.js:18:19)
    at initQUnit (/Users/ldayananda/Documents/work/vhs/node_modules/karma-qunit/lib/index.js:14:39)
    at Array.invoke (/Users/ldayananda/Documents/work/vhs/node_modules/di/lib/injector.js:75:15)
    at Injector.get (/Users/ldayananda/Documents/work/vhs/node_modules/di/lib/injector.js:48:43)
    at /Users/ldayananda/Documents/work/vhs/node_modules/karma/lib/server.js:143:20
    at Array.forEach (<anonymous>)
    at Server._start (/Users/ldayananda/Documents/work/vhs/node_modules/karma/lib/server.js:142:21)
    at Injector.invoke (/Users/ldayananda/Documents/work/vhs/node_modules/di/lib/injector.js:75:15)
    at Server.start (/Users/ldayananda/Documents/work/vhs/node_modules/karma/lib/server.js:103:18)
    at Object.exports.run (/Users/ldayananda/Documents/work/vhs/node_modules/karma/lib/cli.js:280:26)
    at Object.<anonymous> (/Users/ldayananda/Documents/work/vhs/node_modules/karma/bin/karma:3:23)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Function.Module.runMain (module.js:694:10)
    at startup (bootstrap_node.js:204:16)
    at bootstrap_node.js:625:3

gkatsev: could be pinned to ~2.5.1 instead

package.json
```diff
@@ -112,12 +100,12 @@
...
     "rollup-watch": "^4.3.1",
     "semver": "^5.5.0",
-    "serve-static": "^1.13.2",
     "shelljs": "^0.8.2",
-    "sinon": "^1.10.3",
+    "sinon": "1.10.3",
     "uglify-es": "^3.3.9",
     "videojs-contrib-eme": "^3.2.0",
     "videojs-contrib-quality-levels": "^2.0.4",
+    "videojs-generate-karma-config": "^4.0.2",
```
gkatsev: might as well update to ^5

ldayananda:
Alright, tests are back to failing with qunit@v2.5.1, looks like for [good reason](https://github.com/qunitjs/qunit/blob/master/History.md#250--2018-01-09):

```
Core: Capture the proper stack trace in onUnhandledRejection.
Tests: Update tests for unhandled rejection scenarios.
Core: Add QUnit.onUnhandledRejection.
HTML Reporter: Ensure unhandled rejection fails.
CLI: Ensure an unhandled rejection results in a failed test.
```

To make things better, looks like the play promise rejections are coming from disposing the player and don't happen if you run the tests that fail by themselves. I haven't found a fix for them yet.

gkatsev:
Add a .catch to the play promise where-ever play is called in the tests? Can even make it a util: https://github.com/videojs/video.js/blob/ab740bcf097adb26d852d638d7221b8ffa14c95a/src/js/utils/promise.js#L24-L28

ldayananda:
Surprisingly there's only a few places where `play` is actually called. The `afterEach` for many tests call `player.dispose()` which eventually calls `el.load()`. According to the [Google docs on this](https://developers.google.com/web/updates/2017/06/play-request-was-interrupted#fix):

```Note: Calling video.pause() isn't the only way to interrupt a video. You can entirely reset the video playback state, including the buffer, with video.load() and video.src = ''.```

ldayananda:
how does video.js get around this?

gkatsev:
I don't think we have any full integration tests that play the video but I'd expect that if you can silence the promises, you won't get the error.

ldayananda:
Alright, looks like we're leaking some form of state still. Expect a couple of commits as I try to go through and identify what's actually wrong here.

ldayananda:
closing as I'm unlikely to continue this PR

PR #285, opened 2018-11-05
ignore: add stub seekToStreamTime method

ldayananda:
### functional
- videojs-http-streaming.test.js:
  - seekToStreamTime will error if live stream has not started
  - seekToStreamTime will seek to time if buffered
- time.test.js:
  - live: seeks and returns player time seeked to if buffered
  - vod: seeks and returns player time seeked to if buffered
  - live: returns error if we can only get estimates even with retries
  - vod: returns error if we can only get estimates even with retries
  -  returns error if time does not exist in live stream/requesting non-available segment
  - returns error if live stream has not started

### unit
- seekToStreamTime
  - returns error if no playlist or streamTime provided
  - throws error if no callback is provided
  - returns error if any playlist segments do not include programDateTime tags
  - setting pauseAfterSeek to false seeks without pausing
- verifyProgramDateTimeTags
  - only returns true when all segments have programDateTime tags
- findSegmentForStreamTime
  - returns nothing if a match cannot be found
  - returns estimate if segment not buffered
  - returns accurate if segment buffered
- findSegmentForPlayerTime
  - returns nothing if a match cannot be found
  - returns estimate if segment not buffered
  - returns accurate if segment buffered
- getTimestampOffset
  - returns difference in timestamps
  - returns offset in seconds
  - negative offset returned if streamTime is before comparison timestamp
- timeWithinSegment
  - streamTime: true if requestedTime is within the segment duration
  - streamTime:  false if requestedTime is outside the segment duration
  - streamTime: false if requestedTime is before segment start
  - playerTime: true if requestedTime is within the segment duration
  - playerTime:  false if requestedTime is outside the segment duration
  - playerTime: false if requestedTime is before segment start


src/util/time.js
```diff
@@ -86,3 +189,87 @@ export const getStreamTime = ({
...
+}) => {
+
+  if (!callback) {
+    throw new Error('seekToStreamTime: callback must be provided');
+  }
+
+  if (typeof streamTime === 'undefined' || !playlist || !seekTo) {
+    return callback({
+      message: 'seekToStreamTime: streamTime, seekTo and playlist must be provided',
+      newTime: null
```
forbesjo: Why provide `newTime`?

ldayananda: just to have a consistent negative result for `newTime`. This way the caller doesn't need to check for undefined

forbesjo: But the time is the second argument to the callback, the user wouldn't think there would be anything but error information in the error object.

ldayananda: ah true 👍

src/videojs-http-streaming.js
```diff
@@ -667,6 +670,19 @@ class HlsHandler extends Component {
       callback
     });
   }
+
+  // the player must be playing before calling this
+  seekToStreamTime(streamTime, callback, pauseAfterSeek = true, retryCount = 2) {
```
gesinger: I think it would be worth exposing the `playlist` property as part of the API, so that a user doesn't have to disable renditions and ensure we are on a specific one before calling the function.

forbesjo: What would be the advantage of providing a playlist over using the currently loaded one? I think the API would be too complicated.

src/util/time.js
```diff
@@ -1,38 +1,81 @@
 /**
  * @file time.js
  */
-const findSegmentForTime = (time, playlist) => {
 
+const timeWithinSegment = (requestedTime, type, segmentStart, duration) => {
```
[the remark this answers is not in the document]
ldayananda: Yes! Will do once I finish up more of the unit tests

src/videojs-http-streaming.js
gesinger: My thought was that the times can be different depending on which playlist you are in, and if you are seeking, it's possible that you will change playlists.

As it stands, in order for a user to get a time from a specific playlist, they have to disable all other renditions before calling this function. Otherwise, they may end up with the time within a playlist that differs from the one they started on.

ldayananda: I suppose we could add the playlist option and default to the current playlist if not provided

PR #303, opened 2018-11-30
feat: add support for dash manifests describing sidx boxes

### LINKED ISSUE #162: Request sidx defined segments from DASH SegmentBase streams
The following test file cannot be played in the reference player:

https://dash.akamaized.net/dash264/TestCases/10a/1/iis_forest_short_poem_multi_lang_480p_single_adapt_aaclc_sidx.mpd

ldayananda:
## Description
Attempt to fix #162.

**Requires** https://github.com/videojs/mpd-parser/pull/41

## Specific Changes proposed
Dash playlist loader requests the sidx after parsing the master manifest once for the sidx byte range information. Then, the master playlist is mutated.

## Current Status
- Need to investigate ended issue (known issue 1)

## Sources:
```
{ 
  src: "https://dash.akamaized.net/dash264/TestCases/10a/1/iis_forest_short_poem_multi_lang_480p_single_adapt_aaclc_sidx.mpd",
  type: "application/dash+xml"
}
```

### Known Issues
1. there appears to be a bug where the `ended` event is not fired sometimes. This may be an existing issues with DASH.
Repro by using https://dash.akamaized.net/dash264/TestCases/10a/1/iis_forest_short_poem_multi_lang_480p_single_adapt_aaclc_sidx.mpd, seeking to > 25s, changing audio selection, seeking near the end.
The skipped playback test `DASH sidx with alt audio should end` should also reproduce the issue

2. if sidx ranges change in between Periods in a live stream, we do not clean up sidxMapping. 
  This will **not be handled** at this point as there is further multiperiod work that is required to be able to detect a period change. There is a TODO in the code for this, but no Github issues yet

## Requirements Checklist
- [ ] Reviewed by Core Contributor(s)


ldayananda:
Things to do still:

All representations(specifically audio) can also have a SegmentBase and also need their sidx to be requested via DashPlaylistLoader. The current idea is to modify [this](https://github.com/videojs/http-streaming/pull/303/files#diff-ca1029eb342915b65f2b261194fef835R229) to include audio groups as well.

ldayananda:
Appears to be working on at least one source. More testing next week.

src/dash-playlist-loader.js
```diff
@@ -220,7 +226,20 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    for (var key in master.playlists) {
+      if (master.playlists[key].sidx) {
+        hasSidx = true;
+        break;
+      }
+    }
+
+    if (hasSidx) {
+      // Note: this calls done!
+      this.fetchMediaSegmentsFromSidx_(master.playlists, master, done);
```
mjneil: There should be a return here otherwise `done` will get called twice when `hasSidx` is true

src/dash-playlist-loader.js
```diff
@@ -409,35 +429,114 @@ export default class DashPlaylistLoader extends EventTarget {
...
+      this.masterPlaylistLoader_.parseMasterXml(updateWithMaster);
     } else {
       oldMaster = this.master;
-      newMaster = this.parseMasterXml();
+      this.parseMasterXml(updateWithMaster);
     }
+  }
 
-    const updatedMaster = updateMaster(oldMaster, newMaster);
+  addSidxInfoToPlaylists_(activeXhrs, master, doneFn) {
```
mjneil: Any reason to put the sidx helper functions (`addSidxInfoToPlaylists_`, `handleSidxResponse_`, `requestSidx_`, `fetchMediaSegmentsFromSidx_` ) onto the `DashPlaylistLoader` instance? None of them really require a reference to `this` (the only thing is a reference to xhr, which could easily be passed in as a param) and could instead be made into just functions that aren't attached to the class

ldayananda: Ah, yeah that's a good point, most of those are just utilities 👍

ldayananda:
One last bug to figure out: looks like both audio playlists are being requested and overwriting each other, potentially due to the state of the master playlist changing which sidx boxes for each playlist are being requested and parsed. Still debugging

ldayananda:
As an update: I discovered some issues with DashSegmentLoader that needs to be addressed before this PR can proceed.

ldayananda:
Now depends on #386 being merged first

ldayananda:
Here are the things that are left to do:

* unit tests for sidx cases, and probably a playback test with this source
* handle cases where saved sidx doesn't match updatedMaster
* handle source changes and period changes?
* Investigating a final bug, mentioned in known issues in the description

gkatsev:
Tests are failing because no mpd-parser update. We should review and test and get it in.

gkatsev:
Replaced by #455

PR #334, opened 2019-01-08
fix(id3): cuechange event not being triggered on audio-only HLS streams

mparisi76:
This is a proposed fix for https://github.com/videojs/http-streaming/issues/130. Instead of adding text-track data only to video segments, we are now adding to all segments.

One change bringing the addTextTrackData invocation outside of the conditional which checks for video buffer.

Sources:

With audio only
http://li720-23.members.linode.com/hls/audio.m3u8
With video and audio
http://li720-23.members.linode.com/hls/video.m3u8

## Requirements Checklist

### DIFF
--- a/src/mse/virtual-source-buffer.js
+++ b/src/mse/virtual-source-buffer.js
@@ -479,7 +479,7 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
       sortedSegments.video.segments.unshift(sortedSegments.video.initSegment);
       sortedSegments.video.bytes += sortedSegments.video.initSegment.byteLength;
       this.concatAndAppendSegments_(sortedSegments.video, this.videoBuffer_);
-      
+
     } else if (this.videoBuffer_ && (this.audioDisabled_ || !this.audioBuffer_)) {
       // The transmuxer did not return any bytes of video, meaning it was all trimmed
       // for gop alignment. Since we have a video buffer and audio is disabled, updateend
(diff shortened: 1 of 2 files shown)


src/mse/virtual-source-buffer.js
```diff
@@ -491,6 +490,9 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
       triggerUpdateend = true;
     }
 
+	//Add text-track data for all
+	addTextTrackData(this, sortedSegments.captions, sortedSegments.metadata);
```
ldayananda: Looks like the build is failing since this should be indented 2 more spaces (and the comment too)

ldayananda:
One last nice to have would be to add a test!

Here's the test for video-only case: https://github.com/videojs/http-streaming/blob/master/test/mse/html.test.js#L1162

You could add another test that is audio-only be copying that test and modifying

https://github.com/videojs/http-streaming/blob/8f4519435b62cbbe1cfd097a895f0f9b3319368f/test/mse/html.test.js#L1164

to be:

`let sourceBuffer = mediaSource.addSourceBuffer('video/mp2t; codecs=\"avc1.4d400d, mp4a.40.2\"');`

and modifying

https://github.com/videojs/http-streaming/blob/8f4519435b62cbbe1cfd097a895f0f9b3319368f/test/mse/html.test.js#L1204-L1206

to be:

``` 
sourceBuffer.transmuxer_.onmessage(createDataMessage('audio', new Uint8Array(1), {
  metadata
}));
```

Let me know if you need any help! You can run tests locally by running `npm run test`

ldayananda:
@mparisi76 I've made a [PR](https://github.com/mparisi76/http-streaming/pull/1) against your branch with the test! Once that's merged into this branch I think we should be able to merge this PR into master

PR #386, opened 2019-01-24
refactor: align DashPlaylistLoader closer to PlaylistLoader states

ldayananda:
## Description
This is a change to ensure that DashPlaylistLoader and PlaylistLoader follow a similar setup path. This should make it easier for `sidx` boxes received to result in further XHR requests in particular DASH Media Presentations, e.g `SegmentBase` with `indexRange`

## Specific Changes proposed
1. DashPlaylistLoader should go through the following states:

`HAVE_NOTHING -> HAVE_MASTER -> HAVE_METADATA`

2. DashPlaylistLoader should trigger the following events:

- `loadedplaylist` when the master playlist is setup after requesting or refreshing the mpd
- `loadedplaylist` when a video playlist has finished being selected, or a child playlist(for media groups) has finished being selected
- `loadedmetadata` the first time a media playlist has been loaded for a specific loader

3. `updateMaster` should return null if there are no manifest changes 

4. Wait until audioPlaylistLoader has set media to trigger `selectedinitalmedia`

5. We now use `hasPendingRequest` to automatically select a media playlist for the masterPlaylistLoader as a fallback in case one is not selected by MPC. The child loaders are created with a media playlist so this fallback is not necessary for them.

## Sources to use for testing:
- https://dash.akamaized.net/akamai/bbb_30fps/bbb_30fps.mpd

## Requirements Checklist
  - [x] New unit tests added
  - [x] Docs/guides updated


test/dash-playlist-loader.test.js
```diff
@@ -30,51 +30,149 @@ QUnit.test('throws if the playlist url is empty or undefined', function(assert)
...
   standardXHRResponse(this.requests.shift());
-  assert.ok(loader.master, 'the master playlist is available');
+  this.clock.tick(1);
+
   // because DASH only has one manifest, it should go through two loadedplaylists
   // and end with HAVE_METADATA because it already has the first media ready
-  assert.equal(loadedPlaylistStates.length, 2, 'triggered two loadedplaylist events');
-  assert.equal(loadedPlaylistStates[0], 'HAVE_MASTER', 'got master first');
-  assert.equal(loadedPlaylistStates[1], 'HAVE_METADATA', 'got media second');
+  assert.strictEqual(loadedPlaylist, 2,
```
mjneil: This is incorrect. The comment above is under the assumption that we are skipping the `HAVE_MASTER` state and jumping straight to `HAVE_METADATA`, but we discussed mapping the sidx to an hls child playlist conceptually. We should assert that (not using assert syntax cause I'm lazy)
```
loadedPlaylist === 1
loadedMetadata === 0
loaded.master !== undefined
loaded.media() === undefined
loader.state === HAVE_MASTER
```
this is the point at which initial playlist selection would be made, and then a call to `loader.media(initialSelectedPlaylist)` is made. After making that call, you would assert that
```
loadedPlaylist === 2
loadedMetadata === 1
loaded.media() !== undefined
loader.state === HAVE_METADATA
```
This series of assertions will be different than what is needed for the child loader below, which I'll touch on in a separate comment

test/dash-playlist-loader.test.js
```diff
@@ -30,51 +30,149 @@ QUnit.test('throws if the playlist url is empty or undefined', function(assert)
...
+  const childLoader = new DashPlaylistLoader(loader.master.playlists[0], this.fakeHls, false, loader);
+
+  childLoader.on('loadedplaylist', function() {
+    loadedPlaylist++;
+  });
+  childLoader.on('loadedmetadata', function() {
+    loadedMetadata++;
+  });
+
+  childLoader.load();
```
mjneil: Since the child loader will be "loading" a media playlist and not a master playlist, it *will* skip the `HAVE_MASTER` state and move directly to `HAVE_METADATA`, however *NOT* on construction, but after a call to `childLoader.load()`, which is *asynchronous*. This would result in a single `loadedplaylist` event and a single `loadedmetadata` event.
We should add assertions both before and after the `load` call to ensure this behavior is followed.
Before the call to load, no events should have been fired and the state should not have changed
```
loadedPlaylist === 2
loadedMetadata === 1
childLoader.state === HAVE_NOTHING
childLoader.media() === undefined
```
After the call to `load`
```
loadedPlaylist === 3
loadedMetadata === 2
childLoader.state === HAVE_METADATA
childLoader.media() !=== undefined
```

test/dash-playlist-loader.test.js
```diff
@@ -30,51 +30,149 @@ QUnit.test('throws if the playlist url is empty or undefined', function(assert)
...
+  this.clock.tick(1);
+
+  // actually, loadedplaylist is triggered 4 times at this point,
+  // but the 3rd is on childLoader's creation so it's not observable
+  assert.strictEqual(loadedPlaylist, 3, 'fired loadedplaylist thrice');
+  assert.strictEqual(loadedMetadata, 2, 'fired loadedmetadata twice');
+  assert.strictEqual(loader.state, 'HAVE_METADATA', 'the state is correct');
+});
+
+QUnit.test('jumps to HAVE_METADATA when initialized with a live media playlist',
```
mjneil: I don't understand the point of this test. This is exactly the behavior we do not want.

test/dash-playlist-loader.test.js
```diff
@@ -30,51 +30,149 @@ QUnit.test('throws if the playlist url is empty or undefined', function(assert)
...
+  });
+
+  loader.load();
+  standardXHRResponse(this.requests.shift());
+  this.clock.tick(1);
+
+  assert.strictEqual(mediaChangings, 0, 'initial selection is not a media changing');
+  assert.strictEqual(mediaChanges, 0, 'initial selection is not a media change');
+
+  loader.media(loader.master.playlists[1]);
```
mjneil: This call would be considered "initial selection" and should result in `0` `mediachanging/mediachange` events. The rest of the assertions would cascade a `-1` to account for this

test/dash-playlist-loader.test.js
```diff
@@ -30,51 +30,149 @@ QUnit.test('throws if the playlist url is empty or undefined', function(assert)
...
+  });
+  loader.on('mediachanging', function() {
+    mediaChangings++;
+  });
+
+  loader.load();
+  standardXHRResponse(this.requests.shift());
+  this.clock.tick(1);
+
+  assert.strictEqual(mediaChangings, 0, 'initial selection is not a media changing');
```
mjneil: this would be "no selection" rather than "initial selection"

test/dash-playlist-loader.test.js
ldayananda: If there is no sidx though, wouldn't we want to make the initial playlist selection as part of `loader.load()`? The current asserts are expecting that the video playlist has been selected.

test/dash-playlist-loader.test.js
ldayananda: I think this will probably get reworked or replaced with the skipped test once the code changes are in

ldayananda: I think this should probably actually be for a child playlist loader instead of how I have it right now

test/dash-playlist-loader.test.js
ldayananda: PlaylistLoader also does this https://github.com/videojs/http-streaming/blob/master/src/playlist-loader.js#L543

mjneil: Only if `this.request` is not set. You can see in the line before, we trigger `loadedplaylist`, which has a listener here https://github.com/videojs/http-streaming/blob/master/src/master-playlist-controller.js#L222-L242 This will call `masterPlaylistLoader_.media()` which will result in a request for a media playlist be made, and thus `this.request` will be set and that pathway will not be taken.

So definitely ew want to make the initial selection for the `DashPlaylistLoader` in `load`, but only if `this.request` is *not* set. The big difference here is that for the `DashPlaylistLoader` request is now referring to either the sidx request or just an asynchronous `setTimeout` in the case we don't actually have a sidx but still wait to "fake" that we are "making a request".

test/dash-playlist-loader.test.js
mjneil: I thought the point of these added tests was to write tests that test the functionality we want in this new DashPlaylistLoader world, if this test is not valid in that world, we should remove it. This test being for the child playlist loader makes sense. However, is it really testing anything different from the test above?

test/dash-playlist-loader.test.js
ldayananda: After double checking the PlaylistLoader path, it does appear that the child loader does actually go through the `HAVE_MASTER` state briefly in `start()` when the master playlist is inferred. I plan to do the same in DashPlaylistLoader to make both paths be in line with each other

test/dash-playlist-loader.test.js
ldayananda: the initial selection happens as part of `onClientServerClockSync_`, triggered from `start`

test/videojs-http-streaming.test.js
```diff
@@ -2915,6 +2915,50 @@ QUnit.test('configures eme if present on selectedinitialmedia', function(assert)
   }, 'set source eme options');
 });
 
+QUnit.test('integration: configures eme if present on selectedinitialmedia', function(assert) {
```
ldayananda: The test above this tests the same thing, but with all the playlist loaders mocked. Should we keep both?

src/playlist-loader.js
```diff
@@ -537,6 +537,8 @@ export default class PlaylistLoader extends EventTarget {
         setupMediaPlaylists(this.master);
         resolveMediaGroupUris(this.master);
 
+        // This will trigger an existing child loader to be loaded
```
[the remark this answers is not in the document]
ldayananda: when I said "loaded" here I meant `load()` is called, but I think your explanation is a bit clearer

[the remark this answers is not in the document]
ldayananda: oh, also this is meant to be about the child playlist loader's `loadedplaylist` event that calls `load()`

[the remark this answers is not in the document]
ldayananda: changing this to:

>>>If this is a child loader, this will trigger the child loader to make a request for the child manifest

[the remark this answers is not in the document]
mjneil: but this is within `if (parser.manifest.playlists)` which would only evaluate to true if it parsed a master playlist, which a child loader would never have, so it would never be a child loader at this point

src/dash-playlist-loader.js
```diff
@@ -336,35 +358,37 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    this.state = 'HAVE_MASTER';
+
+    if (!this.masterPlaylistLoader_) {
+      this.master = this.parseMasterXml();
+      // We have the master playlist at this point, so
+      // trigger this to allow MasterPlaylistController
+      // to make an initial playlist selection
+      this.trigger('loadedplaylist');
+
+    } else if (!this.media_) {
```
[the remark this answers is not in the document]
ldayananda: I see where you are coming from 🤔 . Here are the reasons I eventually decided on this path:

i. it did not appear to actually be used by the main loader of either the PlaylistLoader or DashPlaylistLoader in playback 

ii. it was being used only in the tests which made it seem like this was a common case and resulted in a lot of tests being written assuming that an initial (video) playlist would be set.

iii. `!this.media_` isn't the best equivalent of whether there is a pending request to set media since `this.media_` indicates that the media has already been set. However, this will likely change once we include support for `sidx`, which we could then use `!this.request && !this.media_` to check for which I think would cover both cases?

src/playlist-loader.js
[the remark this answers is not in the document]
ldayananda: _facepalms_

maybe this was the wrong one...

src/dash-playlist-loader.js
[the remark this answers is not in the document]
mjneil: My main concern is that this relies on the behavior that MPC will be making an initial selection after `loadedplaylist` instead of being more independent in that it will trigger the event so something can make an initial selection, but has a fallback. So any change to MPC behavior means the dev also needs to know that a change might need to be made here. It also means we have to have a lot more boilerplate in the the tests to set an initial media when we are trying to test some other behavior.

wrt iii. I agree checking `meia_` isn't really a good equivalent here. I think just checking `request_` would work just like it does in HLS. This is another reason to maintain the asynchronous nature of `media()` even when there is no sidx so it could be synchronous. By keeping async with a setTimeout, you can set `request_`  to that when no sidx is present so behavior aligns with the PlaylistLoader. Basically your `request_` is either for a sidx, or just an asnyc fn call added to the queue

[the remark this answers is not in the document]
ldayananda: >It also means we have to have a lot more boilerplate in the the tests to set an initial media when we are trying to test some other behavior.

I actually think this is a good thing as it's clear what is actually happening in the test rather than having to reference the source code to see that x, y, z methods are called automatically. 

I'll go down the path of setting `request_` to the asynchronously called methods too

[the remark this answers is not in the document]
ldayananda: One thing to note is that `this.master` doesn't exist for the child loader, right? So this would be the fallback for the master loader only (I hope)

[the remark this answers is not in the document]
mjneil: correct, the child loader shouldn't have a `this.master`

src/dash-playlist-loader.js
```diff
@@ -336,35 +387,42 @@ export default class DashPlaylistLoader extends EventTarget {
...
-    }, 0);
 
     // TODO: minimumUpdatePeriod can have a value of 0. Currently the manifest will not
     // be refreshed when this is the case. The inter-op guide says that when the
     // minimumUpdatePeriod is 0, the manifest should outline all currently available
     // segments, but future segments may require an update. I think a good solution
     // would be to update the manifest at the same rate that the media playlists
     // are "refreshed", i.e. every targetDuration.
-    if (this.master.minimumUpdatePeriod) {
+    if (this.master && this.master.minimumUpdatePeriod) {
```
mjneil: Mostly just curious was there a situation where `this.master` was `undefined` and you needed to add this check? or just adding it in the name of good practices? (which I don't disagree with)

ldayananda: At this point, it's just good practices. I think it was actually necessarily before I stopped childLoaders from calling `onClientServerClockSync_`

PR #389, opened 2019-01-25
fix: Prevent exceptions from being thrown by the MediaSource

BrandonOCasey:
See code comments

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -688,8 +688,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return;
     }
 
-    // sometimes endOfStream will fail here due to suourcebuffers updating
-    // often times it will not.
+    // on chrome calling endOfStream can sometimes cause an exception,
+    // even when the media source is in a valid state.
     try {
       this.mediaSource.endOfStream();
     } catch (e) {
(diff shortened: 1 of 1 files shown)


src/master-playlist-controller.js
```diff
@@ -684,9 +684,18 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    const isOpen = this.mediaSource.readyState === 'open';
+
+    // calling end of stream if the media source is not open
+    // or the sourcebuffers are updating causes an exception
+    // on chrome, this prevents that.
+    if (!isEndOfStream || isUpdating || !isOpen) {
+      return;
     }
+
+    this.mediaSource.endOfStream();
```
BrandonOCasey: Previously this function would cause an exception to be thrown on Chrome if the source buffers were updating or the media source was not open. This has been fixed.

src/master-playlist-controller.js
```diff
@@ -944,9 +953,15 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
     let buffered = this.tech_.buffered();
     let setDuration = () => {
-      this.mediaSource.duration = newDuration;
+      // on firefox setting the duration may sometimes cause an exception
+      // even if the media source is open and source buffers are not
+      // updating.
+      try {
+        this.mediaSource.duration = newDuration;
+      } catch (e) {
+        videojs.log.warn('Failed to set media source duration', e);
```
BrandonOCasey: I tried for a long time to determine the cause of this, it almost seems like a bug in firefox. Originally I thought it was due to `updating` sourceBuffers or the media source not being `open` but that is not the case.

src/master-playlist-controller.js
```diff
@@ -684,8 +684,16 @@ export class MasterPlaylistController extends videojs.EventTarget {
       }
     }
 
-    if (isEndOfStream) {
+    if (!isEndOfStream) {
+      return;
+    }
+
+    // on chrome calling endOfStream can sometimes cause an exception,
```
BrandonOCasey: Chrome can sometimes throw a DOM exception here when `endOfStream` is called. It seems like it almost always happens when:
1. `sourceBuffers` are still updating
2. The `mediaSource` is no longer `open`

It does seem to happen at other times though. So I think we just need to wrap this in a try catch here.

PR #392, opened 2019-01-30
fix: Prevent infinite buffering at the start of looped video on edge

BrandonOCasey:
### Problem
Right now Edge will sometimes buffer indefinitely once a video that is set to loop moves back to `currentTime` 0. 

After investigating this it appears that Edge does not set `tech.seeking()` to `true` during a `seeking` event for looping back to `currentTime` 0. Normal `seeking` events do set `tech.seeking()` to true, as expected. 

### Solution

I don't think that we need the extra check for `seeking` on the `tech` at the start of the if statement. If we just had a `seeking` event it should be fairly evident that we are seeking. I also think that the other two checks (for currentTime, and loop) should be enough to have this only happen when a video is looping.

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -378,7 +378,7 @@ class HlsHandler extends Component {
 
     // Handle seeking when looping - middleware doesn't handle this seek event from the tech
     this.on(this.tech_, 'seeking', function() {
-      if (this.tech_.seeking() && this.tech_.currentTime() === 0 && this.tech_.player_.loop()) {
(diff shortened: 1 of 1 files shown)


forbesjo:
I wonder if the problem is by the time the callback is called the video element has resolved the seek and `seeking` is no longer true. This looks safe to me.

forbesjo:
Perhaps we should consider either reverting https://github.com/videojs/http-streaming/pull/161 and only react to seek events or very clearly document the instances where the event is unreliable.

PR #396, opened 2019-02-07
fix: Clear the blacklist for other playlists if final rendition errors

forbesjo:
https://github.com/videojs/videojs-contrib-hls/pull/1083

## Description
Before the behavior is we never blacklist the final request and wait up to 5 minutes for other playlists' blacklist duration to be cleared. Change the behavior as follows:

- If there is only one rendition and it errors, never blacklist it
- if there are multiple renditions and we are down to the last one (the previous renditions are all blacklisted) and it errors, we clear the blacklist duration for the other playlists, and retry the best available playlist after a delay.

## Requirements Checklist
  - [x] Docs/guides updated


src/master-playlist-controller.js
```diff
@@ -765,15 +765,33 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+
+    if (isFinalRendition) {
+      // Since we're on the final non-blacklisted playlist, and we're about to blacklist
+      // it, instead of erring the player or retrying this playlist, clear out the current
+      // blacklist. This allows other playlists to be attempted in case any have been
+      // fixed.
+      videojs.log.warn('Removing all playlists from the blacklist because the last ' +
+                       'rendition is about to be blacklisted.');
+      playlists.forEach((playlist) => {
+        delete playlist.excludeUntil;
```
gesinger: I think this was a bug from the original PR, but we need to check to ensure we don't clear incompatible playlists from the blacklist (where `excludeUntil` === `Infinity`).

gkatsev:
Good to go once we ignore `excludeUntil === Infinity`

gkatsev:
Not sure why the test for commit f805916 is showing as passed, it failed as you can see here: https://travis-ci.org/videojs/http-streaming/builds/514306305#L2492. However, it's supposed to fail to show that the test works as expected. The build for the fix commit (b37e4fc) is here: https://travis-ci.org/videojs/http-streaming/builds/514308525.

forbesjo:
Looks good to me

PR #398, opened 2019-02-07
test: fix intermittent Firefox loop test failures

forbesjo:
Firefox won't loop if seeking directly to the duration, or to too close
to the duration. 100ms from the duration seems to work.


forbesjo:
Closing since #371 will be merged soon

PR #401, opened 2019-02-12
fix: Fix video playback freezes caused by not using tech current time

BrandonOCasey:
This bug is somewhat hard to reproduce as it only happens when another middleware reports a `currentTime` that is not the same as what vhs suggests.


gkatsev:
A potential test for this is to make a middleware that always returns a "weird" time, i.e., a time that's different from what the tech returns for a regular play. For example, on the first play request, have the middleware return 5s. Presumably, that would trigger this issue.

PR #412, opened 2019-02-21
Migrate HLS redirect logic from contrib-hls PR; Add logic for DASH

OshinKaramian:
## Description
This is partially a port of https://github.com/videojs/videojs-contrib-hls/pull/1213 with additional changes to handle redirects for DASH.

## Specific Changes proposed
Adds an option `handleManifestRedirects` to the source that, when enabled, will allow manifest redirects to function correctly.  If a manifest contains a relative route, it will utilize the redirected path (without this functionality the path will be based on the url we are being redirected from).

## Manual Testing
Sample HLS Manifest: https://dash.akamaized.net/dash264/TestCases/1c/qualcomm/2/MultiRate.mpd
Sample DASH Manifest: https://bitdash-a.akamaihd.net/content/MI201109210084_1/m3u8s/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.m3u8

I used a few chrome plugins to get the redirect to work (I attempted to use Charles with rewrite rules and remote mapping but did not have luck getting it to work correctly).  

Redirectly (to redirect the origin manifest): https://chrome.google.com/webstore/detail/requestly-redirect-url-mo/mdnleldcmiljblolnjhpnblkcekpdkpa?hl=en

Allow-Control-Access-Origin (to hijack/ignore CORS): https://chrome.google.com/webstore/detail/allow-control-allow-origi/nlfbmbojpeacfghkpbjhddihlkkiljbi/related?hl=en

Setup:
- In Redirectly -> Rules -> (+) New Rule -> Redirect Request
- Set URL match rule to something you're redirecting from (as an example http://example.com/hls.m3u8) and the destination URL to one of the above assets (as an example https://dash.akamaized.net/dash264/TestCases/1c/qualcomm/2/MultiRate.mpd)
- `npm start` the project and get the local server running.
- Edit the base `index.html` so that the source has a `handleManifestRedirect` option:

```
player.src({
  src: url.value,
  type: mimeType.options[mimeType.selectedIndex].innerText,
  handleManifestRedirects: true
});
```
- http://localhost:9999/index.html
- Add the url you put in redirectly (http://example.com/hls.m3u8) as the Video Url.

## Requirements Checklist
  - [x] Docs/guides updated


test/test-helpers.js
```diff
@@ -104,6 +105,11 @@ export class MockTextTrack {
   }
 }
 
+// return an absolute version of a page-relative URL
+export const absoluteUrl = function(relativeUrl) {
+  return URLToolkit.buildAbsoluteURL(window.location.href, relativeUrl);
```
gkatsev: is this compatible with our implementation?

test/videojs-http-streaming.test.js
```diff
@@ -1662,6 +1662,88 @@ QUnit.test('if withCredentials global option is used, withCredentials is set on
...
+  assert.ok(this.player.tech_.hls.masterPlaylistController_.masterPlaylistLoader_.handleManifestRedirects);
+
+  videojs.options.hls = hlsOptions;
+});
+
+QUnit.test('the handleManifestRedirects in DashPlaylistLoader option overrides the global default', function(assert) {
+  let hlsOptions = videojs.options.hls;
+
+  this.player.dispose();
+  videojs.options.hls = {
```
gkatsev: Is the `hls` option used to configure VHS in general still? It affects DASH playback? We probably should migrate to a `vhs` option or something instead (not as part of this PR for sure)

OshinKaramian: Looking around, both can be used.  In the codebase itself (and other tests), I'm seeing `videojs.options.hls` used (as well as in the test `index.html`).  

I'm going to leave this as is (for consistency), as I thought the `hls` and `vhs` keys are the same and I would like to keep it consistent.  I agree it would be nice to get rid of `hls` in a future PR.

gkatsev: Ok, cool, wasn't sure.
Yeah, def not something for this PR.

test/test-helpers.js
OshinKaramian: I believe yes for two reasons:

1. It was used in the contrib-hls pr, I figure if there were issues it wouldn't have been ported over
2. The tests that use it seem to still pass.

I can undo it if you would prefer though.

test/playlist-loader.test.js
```diff
@@ -952,6 +952,68 @@ QUnit.test('resolves media initialization segment URIs', function(assert) {
...
+QUnit.test('recognizes redirect, when manifest requested', function(assert) {
+  let loader = new PlaylistLoader('manifest/media.m3u8', this.fakeHls, {
+    handleManifestRedirects: true
+  });
+
+  loader.load();
+
+  const manifestRequest = this.requests.shift();
+
+  manifestRequest.responseURL = window.location.protocol + '//' +
```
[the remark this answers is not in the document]
OshinKaramian: If you look at the line that does the redirect (`resolveManifestRedirect`) it is using the `req.responseUrl` to gauge things (this was also the case in my manual testing).  I'm not sure how else to test it, this test is also within the ported over PR.

test/test-helpers.js
gkatsev: Nah, it's fine. Carry on :)

test/playlist-loader.test.js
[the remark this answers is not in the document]
gkatsev: Yeah, just asking to help my understanding

src/dash-playlist-loader.js
```diff
@@ -327,6 +329,8 @@ export default class DashPlaylistLoader extends EventTarget {
         this.masterLoaded_ = Date.now();
       }
 
+      this.srcUrl = resolveManifestRedirect(this.handleManifestRedirects, this.srcUrl, req);
```
ldayananda: is this the only place that this needs to be done? What about `refreshXml`?

src/playlist-loader.js
```diff
@@ -558,14 +587,14 @@ export default class PlaylistLoader extends EventTarget {
         uri: window.location.href,
         playlists: [{
           uri: this.srcUrl,
-          id: 0
+          id: 0,
+          resolvedUri: this.srcUrl,
```
ldayananda: should the faked `resolvedUri` for the master playlist also be run through `resolveManifestRedirect`?

test/dash-playlist-loader.test.js
```diff
@@ -1102,7 +1102,7 @@ QUnit.test('starts without any metadata', function(assert) {
 });
 
 QUnit.test('moves to HAVE_MASTER after loading a master playlist', function(assert) {
-  const loader = new DashPlaylistLoader('dash.mpd', this.fakeHls);
+  const loader = new DashPlaylistLoader('dash.mpd', this.fakeHls, { withCredentials: false });
```
ldayananda: why was this needed?

test/dash-playlist-loader.test.js
```diff
@@ -1102,7 +1102,7 @@ QUnit.test('starts without any metadata', function(assert) {
...
   const origHasPendingRequest = loader.hasPendingRequest;
```
ldayananda: Can we add a test to see if we replace the `resolvedUri` if the option is true?

test/dash-playlist-loader.test.js
ldayananda: there's also probably more tests now in this file that need to be converted from `false` to `{withCredentials: false}`

test/playback.test.js
```diff
@@ -167,8 +167,6 @@ QUnit.test('Live DASH', function(assert) {
 });
 
 QUnit.test('loops', function(assert) {
-  assert.timeout(5000);
```
ldayananda: this should be kept in since it stops the tests from getting stuck

test/playlist-loader.test.js
```diff
@@ -1005,6 +1005,68 @@ QUnit.test('resolves media initialization segment URIs', function(assert) {
...
+                                '#EXTINF:10,\n' +
+                                '/00001.ts\n' +
+                                '#EXT-X-ENDLIST\n');
+  assert.equal(loader.media().segments[0].resolvedUri,
+              window.location.protocol + '//' +
+              'foo-bar.com/00001.ts',
+              'resolved segment URI');
+});
+
+QUnit.test('recognizes redirect, when media requested', function(assert) {
```
ldayananda: is the test name correct here? How is it different from the previous test?

src/playlist-loader.js
```diff
@@ -33,6 +33,30 @@ export const forEachMediaGroup = (master, callback) => {
...
+ * on `handleManifestRedirects` option
+ *
+ * @api private
+ *
+ * @param  {String} url - an url being requested
+ * @param  {XMLHttpRequest} req - xhr request result
+ *
+ * @return {String}
+ */
+export const resolveManifestRedirect = (handleManifestRedirect, url, req) => {
```
ldayananda: maybe this makes more sense to be in the `resolve-url.js` file now that it's used for HLS and DASH?

test/videojs-http-streaming.test.js
```diff
@@ -1662,6 +1662,88 @@ QUnit.test('if withCredentials global option is used, withCredentials is set on
...
+  };
+  this.player = createPlayer();
+  this.player.src({
+    src: 'http://example.com/media.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+
+  this.clock.tick(1);
+
+  assert.ok(this.player.tech_.hls.masterPlaylistController_.masterPlaylistLoader_.handleManifestRedirects);
```
ldayananda: this should have an assert name

test/videojs-http-streaming.test.js
```diff
@@ -1662,6 +1662,88 @@ QUnit.test('if withCredentials global option is used, withCredentials is set on
...
+
+  videojs.options.hls = hlsOptions;
+});
+
+QUnit.test('the handleManifestRedirects option overrides the global default', function(assert) {
```
ldayananda: `handleManifestRedirects source option`

src/dash-playlist-loader.js
OshinKaramian: I think it wouldn't hurt (other than adding some complexity) *but* I don't think it's necessary.  You should already be on the redirected manifest by the time you're refreshing it.  The redirect is really only going to happen on an initial load of an asset.

src/playlist-loader.js
OshinKaramian: It should have already been set correctly above on line 558.

test/dash-playlist-loader.test.js
OshinKaramian: I think the mock needed it because there was some assumption with options in the test code... somewhere.  I can try and remove it to see if it blows up.

test/playback.test.js
OshinKaramian: Ah yes, I removed this way earlier because I was seeing something weird (that went away).  I'll re-add this.

test/playlist-loader.test.js
OshinKaramian: This looks like a copy/pasta error :o 

Good catch.

ldayananda:
Gave it a test, and looks good overall

src/dash-playlist-loader.js
ldayananda: I guess it's more of an issue when loading a "media" playlist which could be the sidx in a later PR. So, not a huge problem, but we'll have to deal with extra complexity at some point

OshinKaramian: I haven't committed it yet, but I'm going to add a comment to suggest that this should be resolved due to sidx.

PR #416, opened 2019-02-26
Fix/never calling eos

BrandonOCasey:
## Description
Trigger `ended` on segment loaders while the underlying sourcebuffers were updating was incorrect. This caused some videos to never call `endOfStream` and hang at the end of the video.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -665,9 +665,11 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
-    let isEndOfStream = detectEndOfStream(this.playlist_,
-                                          this.mediaSource_,
-                                          segmentInfo.mediaIndex);
+    let isEndOfStream = detectEndOfStream(
+      this.playlist_,
+      this.mediaSource_,
+      segmentInfo.mediaIndex
+    ) && !this.sourceUpdater_.updating();
 
     if (isEndOfStream) {
       this.endOfStream();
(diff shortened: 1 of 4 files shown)


src/source-updater.js
```diff
@@ -160,7 +160,10 @@ export default class SourceUpdater {
    * @return {Boolean} the updating status of the SourceBuffer
    */
   updating() {
-    return !this.sourceBuffer_ || this.sourceBuffer_.updating || this.pendingCallback_;
+    // we are updating if the sourcebuffer is updating or
+    return !this.sourceBuffer_ || this.sourceBuffer_.updating ||
+      // if we have a pending callback that is not our internal noop
+      (!!this.pendingCallback_ && this.pendingCallback_ !== noop);
```
BrandonOCasey: I noticed that this function would return the value of `pendingCallback_` rather than a boolean I also noticed that our internal noop is counted towards wether we are updating, which I think is incorrect.

src/segment-loader.js
```diff
@@ -665,9 +673,11 @@ export default class SegmentLoader extends videojs.EventTarget {
...
     }
 
-    let isEndOfStream = detectEndOfStream(this.playlist_,
-                                          this.mediaSource_,
-                                          segmentInfo.mediaIndex);
+    let isEndOfStream = detectEndOfStream(
+      this.playlist_,
+      this.mediaSource_,
+      segmentInfo.mediaIndex
+    ) && !this.sourceUpdater_.updating();
```
misteroneill: Should this condition be inside the `detectEndOfStream` function?

BrandonOCasey: I thought about it but decided against it, because it seemed weird to pass the source updater just so we could check updating in the function. Perhaps a new private function on the segment loader would make sense? `this.isEndOfStream_`?

BrandonOCasey:
sure I can add a test

PR #435, opened 2019-03-20
Synchronize the audio and video channel when timestamps are different

Llorito:
When we have a hls (with fmp4 fragments) with a separate video channel and an audio channel, if they have different segments timestamp offset value (it is possible according to the HLS specification), it does not play correctly. Both are out of sync.

Specific Changes proposed
Change in sync-controller.js in case we have async timescodes.


gkatsev:
This would be good to get in, @squarebracket can you review?
Also, this may no longer be necessary or will need to be udpated once transmux before append (#272) work is completed.

squarebracket:
@Llorito do you have some example media that shows this behavior? I have some hls-ts and dash media that exhibits non-sync behavior, but no hls-fmp4 media that does.

PR #437, opened 2019-03-20
fix(id3): ignore unsupported id3 frames

gkatsev:
When we get an unsupported id3 frame, we don't parse its time. This ends
up coming back to us as NaN in the timing for the frame. Instead of
trying to create a cue for this frame, ignore it whenever the time isn't
a number or when it isn't finite.

Fixes videojs/video.js#5823

### DIFF
--- a/src/mse/add-text-track-data.js
+++ b/src/mse/add-text-track-data.js
@@ -81,6 +81,14 @@ export const addTextTrackData = function(sourceHandler, captionArray, metadataAr
     metadataArray.forEach(function(metadata) {
       let time = metadata.cueTime + this.timestampOffset;
 
+      // if time isn't a finite number between 0 and Infinity, like NaN,
+      // ignore this bit of metadata.
+      // This likely occurs when you have an non-timed ID3 tag like TIT2,
+      // which is the "Title/Songname/Content description" frame
+      if (typeof time !== 'number' || window.isNaN(time) || time < 0 || !(time < Infinity)) {
+        return;
+      }
+
       metadata.frames.forEach(function(frame) {
         let cue = new Cue(
           time,


RafalLukawiecki:
I would be happy to test it using our media. I will get back to you before the weekend, as a little overloaded with workload at the moment. If it does not fix it, I will make example media available for your testing.

gkatsev:
@RafalLukawiecki that would be awesome!

RafalLukawiecki:
Pleasure. I just realised that the PR is on the http-streaming component only. Can I get the PR easily applied using a patch to the npm-installed most recent version of video.js or do I need to create a new testing environment for this PR?

gkatsev:
@RafalLukawiecki you can test on here https://deploy-preview-437--videojs-http-streaming.netlify.com/

RafalLukawiecki:
Thank you, I understand now. In that case it will still require me to make one of our video streams public, so that I can play against your hosted test. Will do.

gkatsev:
since it loads from your browser, as long as you can reach the resource you should be able to test without making it public to the web at large

RafalLukawiecki:
I have tested using your very helpful hosted tester. I am very happy to say that this PR fixes the bug that we have encountered. **Many thanks!** When will it appear in a version of video.js that we can install using npm? Otherwise we have to build from source, right?

gkatsev:
@RafalLukawiecki glad to hear it! If we can test it on our end, it may be out this week, otherwise maybe within a couple of weeks.

RafalLukawiecki:
@gkatsev Would it help if I made a video stream available publicly that showed the original issue and which works now? I can have that ready tomorrow.

PR #442, opened 2019-03-26
Fix: On dispose, don't call abort on SourceBuffer until after remove() has finished

squarebracket:
## Description
Per the MSE spec, [calling `abort()` during a `remove()` will throw an `InvalidStateError`](https://www.w3.org/TR/media-source/#dom-sourcebuffer-abort). During `dispose()` of the `SourceUpdater`, we call `this.sourceBuffer_.abort()` without waiting for the pending operation to finish, even if that operation is a `remove`. Thus, we can get an `InvalidStateError` on `dispose`.

## Specific Changes proposed
Track whether we're in the process of removal. If `dispose` happens before we finish the `remove`, add the dispose function as a handler for `updateend`.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/mse/virtual-source-buffer.js
+++ b/src/mse/virtual-source-buffer.js
@@ -311,7 +311,7 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
             if (t === 'audio' && this.audioDisabled_) {
               return true;
             }
-            // if the other type if updating we don't trigger
+            // if the other type is updating we don't trigger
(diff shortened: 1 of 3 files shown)


forbesjo:
Looks good to me

squarebracket:
Noticed that I was removing the `this.onUpdateendCallback_` callback after the abort, when I think I should have been doing it before, so I changed it around.

PR #446, opened 2019-03-28
feat: add option to cache encrpytion keys in the player

mjneil:
## Description
Adds an option to force the player to cache encryption keys instead of requesting for every segment. Defaults to `false` for backwards compatibility.
Addressess #140 

todo/future enhancement: Use HTTP Expires header to expire our internal cache per https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-6.2.3

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -372,23 +372,31 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @return {Object}
    *         Key object for desired key
    */
-   segmentKey(key, set = false) {
-     if (!key) {
-       return null;
-     }
+  segmentKey(key, set = false) {
+    if (!key) {
+      return null;
+    }
 
-     const id = segmentKeyId(key);
-     let storedKey = this.keyCache_[id];
+    const id = segmentKeyId(key);
+    let storedKey = this.keyCache_[id];
 
-     if (this.cacheEncryptionKeys_ && set && !storedKey && key.bytes) {
-       this.keyCache_[id] = storedKey = {
-         resolvedUri: key.resolvedUri,
-         bytes: key.bytes
-       };
-     }
+    if (this.cacheEncryptionKeys_ && set && !storedKey && key.bytes) {
+      this.keyCache_[id] = storedKey = {
+        resolvedUri: key.resolvedUri,
+        bytes: key.bytes
+      };
+    }
 
-     return storedKey || { resolvedUri: key.resolvedUri };
-   }
+    const result = {
+      resolvedUri: (storedKey || key).resolvedUri
+    };
+
+    if (storedKey) {
+      result.bytes = storedKey.bytes;
+    }
+
+    return result;
+  }
 
   /**
    * Returns true if all configuration required for loading is present, otherwise false.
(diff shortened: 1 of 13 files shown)


src/segment-loader.js
```diff
@@ -355,6 +360,34 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+   segmentKey(key, set = false) {
+     if (!key) {
+       return null;
+     }
+
+     const id = segmentKeyId(key);
+     let storedKey = this.keyCache_[id];
+
+     if (this.cacheEncryptionKeys_ && set && !storedKey && key.bytes) {
+       this.keyCache_[id] = storedKey = {
```
ldayananda: we may also want to check if the key has actually changed? I suppose this would only be if the iv value is different

src/segment-loader.js
```diff
@@ -1048,10 +1081,8 @@ export default class SegmentLoader extends videojs.EventTarget {
         0, 0, 0, segmentInfo.mediaIndex + segmentInfo.playlist.mediaSequence
       ]);
 
-      simpleSegment.key = {
-        resolvedUri: segment.key.resolvedUri,
-        iv
-      };
+      simpleSegment.key = this.segmentKey(segment.key);
+      simpleSegment.key.iv = iv;
```
ldayananda: Does this only work because the m3u8-parser repeats the last parsed `EXT-X-KEY` tag on every segment until it gets a different tag?

src/segment-loader.js
mjneil: A change in IV does not mean the key has changed. Probably the proper approach is to check for the HTTP Expires header and invalidate our cache based on that per https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-6.2.3

src/segment-loader.js
mjneil: correct

src/segment-loader.js
```diff
@@ -1048,10 +1083,8 @@ export default class SegmentLoader extends videojs.EventTarget {
...
```
ldayananda: This winds up adding the `iv` value to the cached key because `segmentKey` will return the reference

mjneil: Good catch. I don't think this will cause any issues as we don't ever use the "cached" iv value since we are always overwriting it here anyway. I can update this to return a new reference though

src/segment-loader.js
```diff
@@ -347,6 +347,8 @@ export default class SegmentLoader extends videojs.EventTarget {
     const id = initSegmentId(map);
     let storedMap = this.initSegments_[id];
 
+    // TODO: We should use the HTTP Expires header to invalidate our cache per
+    // https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-6.2.3
```
ldayananda: Whoops, this accidentally ended up in the `initSegment` method instead of `segmentKey`

hasinhayder:
Just add this to your options and it will be cached. I was suffering from the same issue with latest build, and finally figured it out

```html
html5: {
    vhs: {
      cacheEncryptionKeys: true
    }
  }
```

notice the **vhs**, not hls :)

PR #455, opened 2019-04-05
feat: add support for dash manifests describing sidx boxes

### LINKED ISSUE #162: Request sidx defined segments from DASH SegmentBase streams

ldayananda:
## Description
Attempt to fix #162.

**Requires** https://github.com/videojs/mpd-parser/pull/41

## Specific Changes proposed
Dash playlist loader requests the sidx after parsing the master manifest once for the sidx byte range information. Then, the master playlist is mutated.

## Current Status
- Need to investigate ended issue (known issue 1)

## Sources:
```
{ 
  src: "https://dash.akamaized.net/dash264/TestCases/10a/1/iis_forest_short_poem_multi_lang_480p_single_adapt_aaclc_sidx.mpd",
  type: "application/dash+xml"
}
```

### Known Issues
1. there appears to be a bug where the `ended` event is not fired sometimes. This may be an existing issues with DASH.
Repro by using https://dash.akamaized.net/dash264/TestCases/10a/1/iis_forest_short_poem_multi_lang_480p_single_adapt_aaclc_sidx.mpd, seeking to > 25s, changing audio selection, seeking near the end.
The skipped playback test `DASH sidx with alt audio should end` should also reproduce the issue

2. if sidx ranges change in between Periods in a live stream, we do not clean up sidxMapping. 
  This will **not be handled** at this point as there is further multiperiod work that is required to be able to detect a period change. There is a TODO in the code for this, but no Github issues yet

## Requirements Checklist
- [ ] Reviewed by Core Contributor(s)


gkatsev:
TODO:
- [x] review code
- [x] try it out
- [ ] review tests
- [ ] re-review code
- [ ] have someone else look through the mpd-parser code
- [ ] have someone else look through the code
- [ ] release actual mpd-parser update
- [ ] update the ["current behavior" section](https://github.com/videojs/http-streaming/blob/master/docs/dash-playlist-loader.md#current-behavior) of the DPL doc.

gkatsev:
Another test stream that works: http://download.tsi.telecom-paristech.fr/gpac/DASH_CONFORMANCE/TelecomParisTech/mp4-onDemand/mp4-onDemand-mpd-AV.mpd https://gpac.wp.imt.fr/2012/02/23/dash-sequences/

This was originally failing, but then works after doing a null check for segments when setting discontinuity. http://yt-dash-mse-test.commondatastorage.googleapis.com/media/car-20120827-manifest.mpd, from http://dash-mse-test.appspot.com/media.html.

gkatsev:
This one from the dash-mse-test is also working: http://yt-dash-mse-test.commondatastorage.googleapis.com/media/motion-20120802-manifest.mpd

gkatsev:
Also, this one: http://yt-dash-mse-test.commondatastorage.googleapis.com/media/oops-20120802-manifest.mpd

src/dash-playlist-loader.js
```diff
@@ -472,11 +651,51 @@ export default class DashPlaylistLoader extends EventTarget {
...
+        if (!this.sidxMapping_[sidxKey]) {
+          const playlist = this.media();
+
+          this.request = requestSidx_(
+            playlist.sidx,
+            playlist,
+            this.hls_.xhr,
+            { handleManifestRedirects: this.handleManifestRedirects },
+            this.handleSidxResponse_(playlist, master, this.state, (newMaster, sidx) => {
+              if (!newMaster || !sidx) {
```
BrandonOCasey: would it make more sense for this to be `getSidxFromResponse_` since it isn't actually doing all of the handling.

src/dash-playlist-loader.js
```diff
@@ -178,7 +316,48 @@ export default class DashPlaylistLoader extends EventTarget {
       this.trigger('mediachanging');
     }
 
-    // TODO: check for sidx here
+    if (playlist.sidx) {
```
BrandonOCasey: perhaps it would make more sense to more the mediaRequest below this if statement into `if (!playlist.sidx)` and then return inside of it? That would keep similar code together and get rid of a big if scope.

src/dash-playlist-loader.js
```diff
@@ -65,6 +70,103 @@ export const updateMaster = (oldMaster, newMaster) => {
...
+        newSidxMapping[key] = oldSidxMapping[key];
+      }
+    }
+  }
+
+  return newSidxMapping;
+};
+
+// exported for testing
+export const filterSidxMapping = (masterXml, srcUrl, clientOffset, oldSidxMapping) => {
```
BrandonOCasey: I don't quite understand what this function is filtering. It seems to be merging a new sidx mapping with an old one?

package.json
```diff
@@ -69,7 +69,7 @@
     "aes-decrypter": "3.0.0",
     "global": "^4.3.0",
     "m3u8-parser": "4.3.0",
-    "mpd-parser": "0.7.0",
+    "mpd-parser": "0.8.0-0",
```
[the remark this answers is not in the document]
gkatsev: yup, once mpd-parser is released, I'll update it. This is just to have the tests pass until a full release.

src/dash-playlist-loader.js
```diff
@@ -65,6 +70,103 @@ export const updateMaster = (oldMaster, newMaster) => {
...
+
+  return sidxInfo.uri + '-' +
+    sidxInfo.byterange.offset + '-' +
+    sidxByteRangeEnd;
+};
+
+const equivalentSidx = (a, b) => {
+  let equivalentMap = true;
+
+  if (a.map && b.map) {
```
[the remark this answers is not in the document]
ldayananda: yes, the `map` here is the init segment for the playlist

src/dash-playlist-loader.js
ldayananda: it's removing sidx mappings that have changed (either url or indexRange)

src/dash-playlist-loader.js
ldayananda: maybe, I based the naming off of segmentLoader. Looks like that has `segmentRequestFinished_` and `processSegmentResponse_` which might be better ones to follow for naming

ldayananda:
> I find it extremely confusing that there are so many playlists objects (are some of them arrays?).

The playlist objects are mimicking HLS playlists, where we have an array of playlists, but also put keys on them (which I agree is confusing). Changing this would be a large refactor

src/dash-playlist-loader.js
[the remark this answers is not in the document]
BrandonOCasey: So should the check be `(a.map || b.map)` so that we check if either has a map?

src/dash-playlist-loader.js
BrandonOCasey: Maybe rename it to `removeChangedMappings` for clarity?

src/dash-playlist-loader.js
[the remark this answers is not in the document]
ldayananda: I suppose the if condition is unnecessary as the value of `equivalentMap` would be falsy if either a or b did not have `map`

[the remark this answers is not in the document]
gkatsev: equivalentMap already checks that both have .map, so, I'm going to remove the if statement.

src/dash-playlist-loader.js
gkatsev: Makes sense to me

src/dash-playlist-loader.js
gkatsev: I think filter is fine there because it parallels Array#filter. However, a comment can be added to clarify.

gkatsev: Updated the name a bit as well.

src/dash-playlist-loader.js
gkatsev: went with sidxRequestFinished_ to match segment loader.

src/dash-playlist-loader.js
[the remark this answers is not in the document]
gkatsev: Improved

PR #466, opened 2019-04-11
feat: Transmux Before Append and LHLS

gesinger:
Includes work by: @mjneil @ldayananda @gesinger
Reopened from https://github.com/videojs/http-streaming/pull/272 as a branch in `http-streaming`

This PR refactors our segment handling logic by transmuxing a segment before appending it to the source buffer, and adds support for LHLS (low latency HLS) in the process.

This allows us to remove the legacy videojs-contrib-media-sources code where we faked the source buffers to allow us to append TS segments directly (and transmuxed within the fake source buffer). There are many reasons we chose to do this, and for a more detailed look into why we went this route, please see the docs dir in this branch: https://github.com/gesinger/http-streaming/blob/transmux-before-append-lhls/docs/lhls/index.md .

Most of the changes are in test files and package-lock.json, however, there is a good deal of refactoring throughout a large part of the codebase. In addition, some legacy code was cleaned up, and some bug fixes were made.

## Portions Reviewed:

Due to the size of this PR, it may make the most sense to review in portions. From individual modules to logical paths.

#### Docs, linting, and scripts
- [x] eslint changes (extending videojs' standard, overriding for tests)
- [x] docs/creating-content.md - sample ffmpeg commands for creating (and created) content
- [x] docs/lhls/index.md - overview of transmux before append and LHLS work
- [x] docs/lhls/transmux-before-append-changes.md
- [x] scripts/segments-data.js - script changes for real segments in tests
- [x] scripts/test.rollup.config.js - script changes for async/await in tests

#### Modules
_and associated tests_
- [x] src/master-playlist-controller.js
- [x] src/media-groups.js
- [x] src/media-segment-request.js
- [x] src/playback-watcher.js
- [x] src/segment-loader.js
- [x] src/segment-transmuxer.js
- [x] src/source-updater.js
- [x] src/sync-controller.js
- [x] src/transmuxer-worker.js
- [x] src/util/codecs.js
- [x] src/util/segment.js
- [x] src/util/string-to-array-buffer.js
- [x] src/util/text-tracks.js
- [x] src/videojs-http-streaming.js
- [x] src/vtt-segment-loader.js

#### Logic Paths and Areas of Focus
_and associated tests_
- [x] (real) media source and source buffer creation (largely in master-playlist-controller)
- [x] duration updates for live and VOD (largely in master-playlist-controller)
- [ ] handling partial data (largely in media-segment-request) (including captions and ID3 tags)
- [x] removed code from videojs-contrib-media-sources
- [ ] captions (largely in segment-loader)
- [ ] id3 metadata (largely in segment-loader)
- [ ] WebWorker creation, management, and cleanup (largely in segment-loader)
- [ ] discontinuities/timestamp offsets (largely in segment-loader)
- [ ] gopsToAlignWith
- [ ] segment-loader queueing
- [ ] segment start and end times (largely in segment-loader)
- [ ] source-updater queueing

## Known Issues
- Alt audio with handlePartialData set to true
- Alt audio, when offset from a 0 start time, sometimes causes video to forever buffer. This happens when the audio downloads and appends before the video has time to set the timestamp offset on the source buffer (therefore, audio will remain offset like the content itself rather than start at 0). Keeping in mind discontinuities/timelines, the audio segment loader will need to wait for video to update the source buffer's timestamp offset (if necessary) within a timeline before appending its own content.
- Muxed fmp4 content (HLS) that doesn't provide codec info in the master manifest doesn't play. This is due to us not parsing the fmp4 segments, and assuming that content is demuxed if using fmp4. We will need to parse the fmp4 segments and not assume demuxed in order to resolve this issue.
- Doubled captions when entering fullscreen. This also happens when switching renditions. It appears the caption timestamps we get back from mux.js are different for different renditions(but the same caption text). It just so happens when entering fullscreen, because we use `keepOriginalTimestamps: true` for this branch, the caption timestamps align and are shown at the same time. Both on master and when doing rendition switches, the captions are actually added twice, but with very different timestamps so they are not displayed together. It remains to be investigated why we are getting these incorrect timestamps back from mux.js in the first place.

## Explicit Areas to Test
- [ ] Not appending video init segment for every segment. From @mjneil: "I think we tried this in the past, but had to leave it as appending video init for every segment cause of a browser issue."


squarebracket:
I went over the docs and briefly over the code. Here are some notes I have regarding timing. There's a lot that I glossed over or just straight up didn't understand, so please correct me if I'm wrong about anything.

1. If you want to keep things simple for this PR, keep a single timestamp offset instead of `(audio|video)TimestampOffset`. Since you must have DTS in audio == DTS in video, you can (and should) keep use a single timestamp offset. (This is not always the case for DASH, which I'll explain below, but currently, we don't handle this DASH situation.) Whatever segment bit completes first should be used as the mapping for that timeline for all segment types. This would imply something outside the segment loaders keeping track of this information. Right now this is done in the sync controller, but it overwrites the timeline mapping for each segment type, which is why you can get desync. If you change the sync controller to set the timeline mapping for the current timeline only if it doesn't exist, then you can just set all `SourceBuffer`s' timestamp offset to be `syncController.mappingForTimeline(segmentInfo.timeline)` for everything.

2. Because of the above, the timing info is the important bit for timestamp offsets. Thus, I don't think `segmentInfo.timestampOffset` should be set until _after_ you've got the timing info, if at all -- logically speaking, the timestamp offset is a property of the media source/the current timeline. This could swap a lot of the weird checks like `segmentInfo.timestampOffset !== null` for ones that make more sense for when you'd set the timestamp offset, like `this.currentTimeline_ !== segmentInfo.timeline`.

3. As I discussed with @mjneil on slack, I don't think the `startsBeforeSourceBufferTimestampOffset` is needed. Since you can't have negative timestamps on media, and we're not rewriting timestamps, I don't see how you could ever run into a problem.

4. Since you're switching to not rewriting timestamps, you may want to ditch the "add audio silence" stuff. It's unclear to me if you're waiting for video timing info before transmuxing/appending audio. If you get audio first and start appending before you get the video timing info, you won't know how much silence to add. A more robust solution would be to simply "jump the gap" into the buffered region. This would cover all scenarios: muxed HLS, demuxed HLS, DASH, initial playback of a live feed, and seeking into a live feed (for instance, pausing for a while then playing). dash.js and shaka (at least for dash) both use this "jump the gap" method. With this TBA branch, will audio silence be added for dash, and when seeking into an unbuffered region of a live feed?


5. To support live DASH feeds properly, we need to add support for handling `presentationTimeOffset` (PTO). This complicates things a bit. Since different adaptation sets can have different PTOs, it means you'd need to have separate timestamp offsets for audio and video. The complicated bit is that a PTO says what the DTS of the first sample would be at the period start, and the current live window can be way after that. Currently, I work around this in my code by looking at the PTOs for all adaptation sets, taking the max, and then subtracting all the PTOs from the max. In this way, I get a "relative PTO" from the other adaptation sets. I can then simply do `timelineMapping + segmentInfo.pto` for the timestamp offset. However, this isn't a great solution.

6. At some point, at least after the initial merge, we _may_ want to consider not sliding the start times for live feeds to zero at stream start. It would simplify the logic for timestamp offsets quite a bit, since we could virtually always have a timestamp offset of zero. It would make handling the PTO situation quite a bit easier, too.


Let me know what you think regarding the above, and I'll put together a PR.

brandonocasey:
Closing in favor of #494 which is now up to date with master.

PR #469, opened 2019-04-12
Fix: livestream timeout issues

brandonocasey:
## Description
This pull requests fixes some HLS livestream timeouts that were happing before playback would even begin. I will go over them in code comments.

### DIFF
--- a/src/media-groups.js
+++ b/src/media-groups.js
@@ -337,10 +337,6 @@ export const setupListeners = {
   }
 };
 
-const byGroupId = (type, groupId) => (playlist) => playlist.attributes[type] === groupId;
-
-const byResolvedUri = (resolvedUri) => (playlist) => playlist.resolvedUri === resolvedUri;
-
 export const initialize = {
   /**
    * Setup PlaylistLoaders and AudioTracks for the audio groups
@@ -357,7 +353,7 @@ export const initialize = {
       sourceType,
       segmentLoaders: { [type]: segmentLoader },
       requestOptions,
-      master: { mediaGroups, playlists },
+      master: { mediaGroups },
       mediaTypes: {
         [type]: {
           groups,
@@ -380,25 +376,9 @@ export const initialize = {
 
       // List of playlists that have an AUDIO attribute value matching the current
       // group ID
-      const groupPlaylists = playlists.filter(byGroupId(type, groupId));
 
       for (let variantLabel in mediaGroups[type][groupId]) {
         let properties = mediaGroups[type][groupId][variantLabel];
-
-        // List of playlists for the current group ID that have a matching uri with
-        // this alternate audio variant
-        const matchingPlaylists =
-          groupPlaylists.filter(byResolvedUri(properties.resolvedUri));
-
-        if (matchingPlaylists.length) {
-          // If there is a playlist that has the same uri as this audio variant, assume
-          // that the playlist is audio only. We delete the resolvedUri property here
-          // to prevent a playlist loader from being created so that we don't have
-          // both the main and audio segment loaders loading the same audio segments
-          // from the same playlist.
-          delete properties.resolvedUri;
-        }
-
         let playlistLoader;
 
         if (properties.resolvedUri) {
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -37,7 +37,6 @@ export const syncPointStrategies = [
 
       let segments = playlist.segments || [];
       let syncPoint = null;
-      let lastDistance = null;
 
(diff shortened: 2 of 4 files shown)


src/media-groups.js
```diff
@@ -380,25 +376,9 @@ export const initialize = {
...
-        const matchingPlaylists =
-          groupPlaylists.filter(byResolvedUri(properties.resolvedUri));
-
-        if (matchingPlaylists.length) {
-          // If there is a playlist that has the same uri as this audio variant, assume
-          // that the playlist is audio only. We delete the resolvedUri property here
-          // to prevent a playlist loader from being created so that we don't have
-          // both the main and audio segment loaders loading the same audio segments
-          // from the same playlist.
-          delete properties.resolvedUri;
```
brandonocasey: Removing this, can cause two segments to be download for the first segment of audio only videos. I think that we should log that as an issue to fix. This makes it so that we do not timeout when a video has separate audio

src/sync-controller.js
```diff
@@ -47,15 +46,7 @@ export const syncPointStrategies = [
         if (segment.dateTimeObject) {
           let segmentTime = segment.dateTimeObject.getTime() / 1000;
           let segmentStart = segmentTime + syncController.datetimeToDisplayTime;
-          let distance = Math.abs(currentTime - segmentStart);
-
-          // Once the distance begins to increase, we have passed
-          // currentTime and can stop looking for better candidates
-          if (lastDistance !== null && lastDistance < distance) {
-            break;
```
brandonocasey: I am not sure if this is the correct fix, but I don't have a video to test with for the code that this was implemented for. Basically without this break I the syncPoint appears correct and with it the syncPoint is always `{time: 0, segmentIndex: 0}`

src/playlist-loader.js
```diff
@@ -363,8 +363,6 @@ export default class PlaylistLoader extends EventTarget {
       playlist = this.master.playlists[playlist];
     }
 
-    window.clearTimeout(this.mediaUpdateTimeout);
```
brandonocasey: This causes livestreams to timeout after the first couple of seconds and timeout. I think we should not clear this timeout here and just let `media()` run at the timeout unless it is cleared somewhere else. See #396

src/sync-controller.js
```diff
@@ -49,9 +49,9 @@ export const syncPointStrategies = [
           let segmentStart = segmentTime + syncController.datetimeToDisplayTime;
           let distance = Math.abs(currentTime - segmentStart);
 
-          // Once the distance begins to increase, we have passed
+          // Once the distance begins to increase, or if distance is 0, we have passed
           // currentTime and can stop looking for better candidates
-          if (lastDistance !== null && lastDistance < distance) {
+          if (lastDistance !== null && (distance === 0 || lastDistance < distance)) {
```
brandonocasey: This is a legitimate fix, the previous code assumed that if the `currentTime` started on the `segmentStart` that we should include that as a valid syncPoint. From my testing this is not the case. We should instead exclude syncPoints that would start exactly on `currentTime`.

EX: `segment#16` at `170s` and `currentTime` at `170s` should choose `segment#15` rather than `segment#16` as `currentTime` may need data from `segment#15` and `segment#16` to play.

test/media-groups.test.js
```diff
@@ -831,7 +831,7 @@ function(assert) {
                'no playlist loader when misconfigured');
 });
 
-QUnit.test('initialize audio does not create playlist loader for alternate tracks with' +
+QUnit.skip('initialize audio does not create playlist loader for alternate tracks with' +
```
brandonocasey: I don't see a way around this test, since it is testing for something that we do not want to happen for now.

src/media-groups.js
gkatsev: So, the issue is that for an audio-only stream, the main segment loader and the audio segment loader would both download the audio segment. Previously, this `delete` statement would prevent the audio segment loader from being created. However, this causes issues when audio and video are unmuxed?

src/sync-controller.js
gkatsev: This explanation would be good to put in the commit body when we merge.

src/media-groups.js
mjneil: Some regression testing should be done on playlists this affects as I believe there was another issue where one of the segment loaders were not calling end of stream

brandonocasey: Yes this causes issues with unmuxed audio/video. Yes we should do some regression testing, but I think that other `endofstream` fixes that have been done fixed those issues, at least for the streams that I have been testing with.

PR #472, opened 2019-04-16
fix(dash-playlist-loader): clear out timers on dispose

gkatsev:


src/dash-playlist-loader.js
```diff
@@ -332,6 +334,7 @@ export default class DashPlaylistLoader extends EventTarget {
     if (!playlist.sidx) {
       // Continue asynchronously if there is no sidx
       // wait one tick to allow haveMaster to run first on a child loader
+      window.clearTimeout(this.mediaRequest_);
```
ldayananda: why are we clearing here? seems like this should happen in `haveMaster` and `haveMetadata` where `mediaRequest` is set to `null` currently

src/dash-playlist-loader.js
```diff
@@ -626,7 +630,8 @@ export default class DashPlaylistLoader extends EventTarget {
     // would be to update the manifest at the same rate that the media playlists
     // are "refreshed", i.e. every targetDuration.
     if (this.master && this.master.minimumUpdatePeriod) {
-      window.setTimeout(() => {
+      window.clearTimeout(this.minimumUpdatePeriodTimeout_);
```
ldayananda: I think we don't intend for this to happen. If you had a timeout set to go off at 3000ms and are now(at 2500ms) trying to set one to go off at 6000ms, you should be able to. Following what `mediaUpdateTimeout` does may be what we want, where the timeout is cleared if we reload the loader or pause it

src/dash-playlist-loader.js
```diff
@@ -483,6 +487,7 @@ export default class DashPlaylistLoader extends EventTarget {
     // We don't need to request the master manifest again
     // Call this asynchronously to match the xhr request behavior below
     if (this.masterPlaylistLoader_) {
+      window.clearTimeout(this.mediaRequest_);
```
ldayananda: probably should remove this too

src/dash-playlist-loader.js
```diff
@@ -698,7 +703,8 @@ export default class DashPlaylistLoader extends EventTarget {
               // update loader's sidxMapping with parsed sidx box
               this.sidxMapping_[sidxKey].sidx = sidx;
 
-              window.setTimeout(() => {
+              window.clearTimeout(this.minimumUpdatePeriodTimeout_);
```
ldayananda: and this

src/dash-playlist-loader.js
```diff
@@ -714,7 +720,8 @@ export default class DashPlaylistLoader extends EventTarget {
         }
       }
 
-      window.setTimeout(() => {
+      window.clearTimeout(this.minimumUpdatePeriodTimeout_);
```
ldayananda: and this

PR #482, opened 2019-04-26
WIP: fix: a/v sync

squarebracket:
## Description
This PR attempts to fix A/V offset that can occur in both TS and fMP4 sources. In short, it aligns DTS by using a "global" timestamp offset per timeline.

Not pictured: something that will ensure you seek into the buffer for live feeds. We may seek into a region that has only video or only audio buffered, and there's currently nothing in the code that ensures we end up within a buffered region. 

## Specific Changes proposed
Please list the specific changes involved in this pull request.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -103,6 +103,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.seekable_ = videojs.createTimeRanges();
     this.hasPlayed_ = () => false;
+    this.bufferedEvents_ = {main: false};
 
     this.syncController_ = new SyncController(options);
     this.segmentMetadataTrack_ = tech.addRemoteTextTrack({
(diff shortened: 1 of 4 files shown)


squarebracket:
Added the gap skipping code. Note a couple things:
- This is POC-level code. It's very likely it can be cleaned up. I just finished tweaking things last night.
- I don't know why the gap-skipping code is delayed in playback watcher. It should probably be instant, or near-instant. 
- I based the code off of the shimmed buffering stuff in #370. I built it in a branch with that patch. So there may be problems with seeking that need to be tested.
- To test the gap skipping, simply take a functional demuxed hls ts or demuxed hls fmp4 playlist, and remove a segment from the beginning of the audio playlist. you'll see that it stalls when starting, but jumps the gap. Without this patch, you'll see desynced a/v.

gkatsev:
We should evaluate whether still can still be applicable to changes in master.

gesinger:
I think that a lot of things have moved around in master since this was created. The a/v sync issue will likely be addressed in master through separate means, but the improved seeking behavior should have another look once we run into the issue more (a separate issue than the original title that has changes included in this PR).

gesinger:
Closing for now.

PR #483, opened 2019-04-26
feature: expose abortEarlyBlacklistDuration setting

squarebracket:
## Description
The duration for which a playlist is blacklisted when one of its segments can't finish in time is currently hard-coded. This PR exposes that duration as a setting.

I've also re-written the description for `blacklistDuration` to describe when that setting is used, so it is clear what cases each setting affects.

## Specific Changes proposed
Expose `abortEarlyBlacklistDuration` as an initialization setting.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Video.js Compatibility: 6.0, 7.0
       - [useCueTags](#usecuetags)
       - [overrideNative](#overridenative)
       - [blacklistDuration](#blacklistduration)
+      - [abortEarlyBlacklistDuration](#abortearlyblacklistduration)
       - [bandwidth](#bandwidth)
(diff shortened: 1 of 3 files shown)


gkatsev:
So, this is a larger issue than just this but I think long term, we should rename `blacklistDuration` to `blocklistDuration` or another alternative term. Since this isn't exposed yet, I think we should start using the new term as part of this and also documentation. I'd like to also add aliases for the old exposed terms, but that should definitely be done in a separate PR. We won't be able to remove them until a major version update anyway.

gkatsev:
To elaborate, as part of some discussion in slack, I think we should move away from the term blacklist to be more inclusive and find a term that fits and doesn't have any potential negative connotations.

PR #484, opened 2019-04-27
fix: change rollup config so that npm run watch works

squarebracket:
## Description
I was annoyed with `npm run watch` not working properly, so I decided to try to fix it. I don't know if this is the right approach, but I did what video.js is doing with adding modules into the `external` config key.

With this + videojs/video.js#5966, you can change a file in VHS and it will be compiled into video.js. It takes a bit to compile, but I found that I would sometimes get some weird behaviour if I was loading both `video.js` and `videojs-http-streaming.js`.

## Specific Changes proposed
Added the modules that caused compilation failure to the `external` config key for CommonJS and ES output.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/scripts/rollup.config.js
+++ b/scripts/rollup.config.js
@@ -31,6 +31,20 @@ const umdPlugins = [
   babel(),
 ];
 
+const externals = [
+  'aes-decrypter',
+  'global/document',
+  'global/window',
+  'm3u8-parser',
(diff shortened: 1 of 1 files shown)


brandonocasey:
Going to give this a quick test, but I think this is the correct approach

PR #492, opened 2019-05-02
chore: async/await support

brandonocasey:
## Description
Add async/await support to tests, from #466


brandonocasey:
I am going to be backing this change out, It makes tests inconsistent, since it adds a lot of load to each test. I was seeing tests fail with dev tools open and pass without. It also make debugging much worse as you have to wade through 5-6 chains in the `regenerator-runtime` to get to what you actually care about.

PR #494, opened 2019-05-03
feat: transmux before append (TBA) and low latency http streaming (LHLS)

brandonocasey:
## Description
This takes most of the changes from #466, https://github.com/gesinger/http-streaming/pull/10, and changes that have occurred in master since both of those and packages them up.

I still have the following to do:
- [x] Verify that all fixes and changes from master have been brought into this branch
- [x] Make the src diff easier to read where possible, or add comments on code.
- [x] Get all the tests working and make sure that all the tests still exist
- [x] what is the difference between `mux.js` `segmentTimingInfo`/`videoSegmentTimingInfo` and `timingInfo`/`videoTimingInfo`.
- [x] Generate an encrypted ts segment with a key for unit tests

This is based on the following, which will have to be merged first:
* [x] #491 
* [x] #488 
* [x] #486 
* [x] #485


src/master-playlist-controller.js
```diff
@@ -210,15 +225,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       });
 
       this.triggerPresenceUsage_(this.master(), media);
-
-      try {
```
brandonocasey: this is handled by a `sourceopen` on the media source in `handleSourceOpen_()`

src/master-playlist-controller.js
```diff
@@ -276,30 +283,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
           this.audioSegmentLoader_.load();
         }
       }
-
-      if (!updatedPlaylist.endList) {
```
brandonocasey: Presumably we no longer have to update `seekable` or deal with `durationchange` since we are using a native media source. This code should have lived in our non-native media source I think.

src/master-playlist-controller.js
```diff
@@ -395,10 +378,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.tech_.trigger({type: 'usage', name: 'hls-aes'});
     }
 
-    if (Hls.Playlist.isFmp4(media)) {
-      this.tech_.trigger({type: 'usage', name: 'hls-fmp4'});
```
brandonocasey: This code was moved to `setupSegmentLoaderListeners_()` and we trigger the usage event when a `segmentLoader` triggers an `fmp4` event.

src/master-playlist-controller.js
```diff
@@ -472,24 +461,31 @@ export class MasterPlaylistController extends videojs.EventTarget {
       }, ABORT_EARLY_BLACKLIST_SECONDS);
     });
 
-    this.mainSegmentLoader_.on('reseteverything', () => {
-      // If playing an MTS stream, a videojs.MediaSource is listening for
```
brandonocasey: we use a native media source so this was no longer needed.

src/master-playlist-controller.js
```diff
@@ -674,6 +671,28 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.trigger('sourceopen');
   }
 
+  /**
+   * handle the sourceended event on the MediaSource
+   *
+   * @private
+   */
+  handleSourceEnded_() {
+    if (!this.inbandTextTracks_.metadataTrack_) {
```
brandonocasey: mostly taken and modified from the [`sourceended` handler in `html-media-source`](https://github.com/videojs/http-streaming/blob/master/src/mse/html-media-source.js#L220-L231)

src/master-playlist-controller.js
```diff
@@ -998,37 +1038,59 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-  updateDuration() {
-    let oldDuration = this.mediaSource.duration;
-    let newDuration = Hls.Playlist.duration(this.masterPlaylistLoader_.media());
-    let buffered = this.tech_.buffered();
-    let setDuration = () => {
-      // on firefox setting the duration may sometimes cause an exception
-      // even if the media source is open and source buffers are not
-      // updating, something about the media source being in an invalid state.
-      this.logger_(`Setting duration from ${this.mediaSource.duration} => ${newDuration}`);
-      try {
```
brandonocasey: The try catch here was moved into sourceUpdaters `actions.duration` handler. Most of the rest of this logic is reformatted but unchanged.

src/media-groups.js
```diff
@@ -138,12 +138,23 @@ export const onTrackChanged = (type, settings) => () => {
     return;
   }
 
-  if (!activeGroup.playlistLoader) {
-    // when switching from demuxed audio/video to muxed audio/video (noted by no playlist
-    // loader for the audio group), we want to do a destructive reset of the main segment
-    // loader and not restart the audio loaders
-    mainSegmentLoader.resetEverything();
-    return;
+  if (type === 'AUDIO') {
```
brandonocasey: I think this just handles moving from demuxed audio/video to muxed audio/video in a better way.

src/playlist.js
```diff
@@ -470,20 +470,6 @@ export const isAes = function(media) {
   return false;
 };
 
-/**
- * Returns whether the current playlist contains fMP4
- *
- * @return {Boolean} true if the playlist contains fMP4
- */
-export const isFmp4 = function(media) {
```
brandonocasey: We now use `util/codecs.js#isLikelyFmp4Data()` to determine if a segment is fmp4 during  media segment requests. Then we add `segment.isFmp4` to the segment object.

src/util/text-tracks.js
```diff
@@ -105,165 +103,121 @@ const deprecateOldCue = function(cue) {
...
-  // create an in-band caption track if one is present in the segment
-  if (segment.captions &&
-      segment.captions.length) {
-    if (!sourceBuffer.inbandTextTracks_) {
-      sourceBuffer.inbandTextTracks_ = {};
-    }
-
-    for (let trackId in segment.captionStreams) {
-      if (!sourceBuffer.inbandTextTracks_[trackId]) {
-        player.tech_.trigger({type: 'usage', name: 'hls-608'});
```
brandonocasey: all of this logic was the same as `createCaptionsTrackIfNotExists`.

src/util/text-tracks.js
```diff
@@ -105,165 +103,121 @@ const deprecateOldCue = function(cue) {
...
+ *   @param {Object} inbandTextTracks the inband text tracks
+ *   @param {Array} metadataArray an array of meta data
+ *   @param {Number} timestampOffset the timestamp offset of the source buffer
+ *   @param {Number} videoDuration the duration of the video
  * @private
  */
-export const addTextTrackData = function(sourceHandler, captionArray, metadataArray) {
-  let Cue = window.WebKitDataCue || window.VTTCue;
-
-  if (captionArray) {
```
brandonocasey: This is/was `addCaptionData`, not sure why we repeated the logic here, but it was removed and is now done in `segment-loader`.

src/source-updater.js
```diff
@@ -13,128 +173,172 @@ import noop from './util/noop';
...
 
-    // run completion handlers and process callbacks as updateend
-    // events fire
-    this.onUpdateendCallback_ = () => {
-      let pendingCallback = this.pendingCallback_;
-
-      this.pendingCallback_ = null;
-      this.sourceBuffer_.removing = false;
-
-      this.logger_(`buffered [${printableRange(this.buffered())}]`);
```
brandonocasey: this logic is now in the `onUpdateend` function right above the class for this file.

src/source-updater.js
```diff
@@ -162,67 +371,121 @@ export default class SourceUpdater {
...
+  videoQueueCallback(callback) {
+    if (this.videoBuffer) {
+      pushQueue({
+        type: 'video',
+        sourceUpdater: this,
+        action: actions.callback(callback),
+        name: 'callback'
+      });
     }
   }
```
brandonocasey: I think this file will eventually need some updates. Some places we have function that pass a `type` and are generic for audio and video. Then others functions are prepended with `audio` or `video` (IE `audioTimestampOffset` vs `videoTimestampOffset`) and they do nearly the same thing.

src/sync-controller.js
```diff
@@ -369,49 +361,19 @@ export default class SyncController extends videojs.EventTarget {
...
 
-  /**
-   * Probe or inspect a fmp4 or an mpeg2-ts segment to determine the start
-   * and end of the segment in it's internal "media time". Used to generate
-   * mappings from that internal "media time" to the display time that is
-   * shown on the player.
-   *
-   * @param {SegmentInfo} segmentInfo - The current active request information
-   */
-  probeSegmentInfo(segmentInfo) {
```
brandonocasey: All of the probe logic seems to have been moved into `src/util/segment`

src/transmuxer-worker.js
```diff
@@ -46,14 +59,19 @@ const wireTransmuxerEvents = function(self, transmuxer) {
...
-      });
+  transmuxer.on('id3Frame', function(id3Frame) {
+    self.postMessage({
+      action: 'id3Frame',
+      id3Frame
     });
-  }
+  });
+
+  transmuxer.on('caption', function(caption) {
```
brandonocasey: I think `caption` and `id3Frame` takes the place of `transmuxer.captionStream`?

src/transmuxer-worker.js
```diff
@@ -145,10 +316,30 @@ class MessageHandlers {
...
     this.transmuxer.flush();
+    // transmuxed done action is fired after both audio/video pipelines are flushed
+    self.postMessage({
+      action: 'done',
+      type: 'transmuxed'
+    });
   }
 
-  resetCaptions() {
-    this.transmuxer.resetCaptions();
```
brandonocasey: Seems like we only resetCaptions in `html-media-source` so this was removed.

package.json
```diff
@@ -23,7 +23,7 @@
     "docs:api": "jsdoc src -r -d docs/api",
     "docs:toc": "doctoc README.md",
     "docs:nomnoml:build": "node -e \"var b=require('./scripts/nomnoml.js'); b.build();\"",
-    "lint": "vjsstandard",
+    "lint": "vjsstandard :",
```
brandonocasey: I turned off the linter, and ran it manually. The version in VHS is very far behind and its nearly impossible to get just errors from it.

gkatsev: I've been doing: `npm run lint | grep -v "maximum\|todo\|loop"` which shows mostly real errors.

PR #496, opened 2019-05-03
fix(playlist controller): improve hasPlayed check

### LINKED ISSUE #464: Switching quality leads to infinite buffering
__Please do not delete the template, by filling out the required information we can investigate your issue more quickly.__

Before opening an issue see if your problem can be resolved with the [troubleshooting guide](https://github.com/videojs/http-streaming/blob/master/docs/troubleshooting.md). 

## Description
Briefly describe the issue.
Include a [reduced test case](https://css-tricks.com/reduced-test-cases/), we have a [starter template](https://jsbin.com/gejugat/edit?html,output) on JSBin you can use.

If you change the quality level of the video before the first TS file loads, it puts VHS into a weird state where it continually downloads the first TS file in the playlist but never appends it to the source buffer or plays. Instead it just scraps the bytes and re-download them again, wasting tons of bandwidth.

Screen Shot 2019-04-11 at 1 37 39 PM

The issue only seems to occur with certain video files (perhaps ones with very large first TS files?) -- I'm still investigating the root cause

Reduced test case (consistently reproduces the issue for me):

https://codepen.io/stevendesu/pen/gyWJep

As best as I can tell, the following seems to be what happens internally:

 1. The video begins to load, and detects all qualities. `720p` is selected as the default due to screen size.
 2. The playlists load, `hasPlayed_()` returns false, so `newPlaylist.syncInfo` is set. `syncPoint_` is set using the `Playlist` strategy (pulling the `syncInfo`)
 3. `hasPlayed_` gets replaced by a function that always returns true
 4. Within my plugin I run a for loop with `qualityLevels[i].enabled = false;` for any qualities that don't match the desired bitrate (setting `enabled` to `false` on the `720p` rendition which had been previously selected)
 5. This triggers `fastQualityChange_` which calls `resetEverything` with a callback to perform a seek
 6. Within `resetEverything` (or some method that it calls) `processingAppend_` is checked and, because it's false, the callback to perform the seek is never run - although manually performing the same seek myself does not fix infinite buffering, so I don't think this is the cause of the bug
 7. When we request the first TS file, `syncPoint_` is null so `isSyncRequest` is set to true
 8. Once the TS file loads, because `isSyncRequest` is true, we trigger `syncinfoupdate` and then scrap the TS file without appending it to the buffer
 9. Later `monitorBufferTick_` runs, notices our buffer is empty, calls `fillBuffer_`, and re-downloads the same TS file. But because `syncPoint_` is null, `isSyncRequest` is set to true
 10. Because `isSyncRequest` is true, we trigger `syncinfoupdate` then scrap the TS file without appending it to the buffer
 11. Repeat steps 9 and 10 infinitely

## Sources
Is a certain source or a certain segment affected? please provide a public (accesible over the internet) link to it below.

https://vcloud.blueframetech.com/file/hls/126136.m3u8

## Steps to reproduce

1. Include `videojs-contrib-quality-levels`
2. Create a video with `autoplay` enabled
3. Bind a listener for `addqualitylevel`
4. In this listener, disable all but one quality level. Be sure to disable whichever quality would normally be selected as default by the screen size

### Expected
The video should play at the specified resolution

### Error output
The player displays a loading spinner and the developer console shows the first TS file being downloaded every half-second infinitely

## Additional Information
Please include any additional information necessary here. Including the following:

what version of videojs-http-streaming does this occur with?
videojs-http-streaming 1.5.1

### videojs version
what version of videojs does this occur with?
video.js 7.4.1

### Browsers
what browsers are affected? please include browser and version for each
* Firefox 66.0.2
* Chrome 73.0.3683.103

### Platforms
what platforms are affected? please include operating system and version or device and version for each
* Mac OS X 10.14.4
* Windows 10

### Other Plugins
* One custom plugin, as described in the test case linked

### Other JavaScript
None

gkatsev:
This should fix #464.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -102,7 +102,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.mediaSource.addEventListener('sourceopen', this.handleSourceOpen_.bind(this));
 
     this.seekable_ = videojs.createTimeRanges();
-    this.hasPlayed_ = () => false;
+    this.hasPlayed_ = () => this.tech_.played().length > 0;
 
     this.syncController_ = new SyncController(options);
     this.segmentMetadataTrack_ = tech.addRemoteTextTrack({
@@ -624,7 +624,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
         this.tech_.one('loadedmetadata', () => {
           this.trigger('firstplay');
           this.seekTo_(seekable.end(0));
-          this.hasPlayed_ = () => true;
         });
 
         return false;
@@ -636,7 +635,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.seekTo_(seekable.end(0));
     }
 
-    this.hasPlayed_ = () => true;
     // we can begin loading now that everything is ready
     this.load();
     return true;


squarebracket:
Do you think there's any problem with potentially calling `setupFirstPlay` more than once?

gkatsev:
That is a good thing to check @squarebracket.

gkatsev:
Fixed as part of #700. Release soon.

PR #502, opened 2019-05-08
fix: only reset syncController if we passed a discontinuity

gkatsev:


gkatsev:
Unless anyone gets a good idea, I don't think there's a good test to add for this.

gkatsev:
Doesn't seem to cause issues with demuxed audio stream either.

test/segment-loader.test.js
```diff
@@ -442,6 +442,42 @@ QUnit.module('SegmentLoader: M2TS', function(hooks) {
...
+      this.requests.shift().respond(200, null, '');
+      buffered = videojs.createTimeRanges([[0, 10]]);
+      this.updateend();
+      this.clock.tick(1);
+
+      assert.equal(hlsTimestampOffsetEvents, 0,
+        'no hls-timestamp-offset event was fired');
+      this.requests[0].response = new Uint8Array(10).buffer;
+      this.requests.shift().respond(200, null, '');
+      loader.sourceUpdater_.timestampOffset_ = 11;
```
mjneil: what is this line doing?

gkatsev: terrible things but it's responsible for making the conditional that was removed pass

mjneil: I think you would want to do this before the updateend or clock tick above 465-466

src/segment-loader.js
```diff
@@ -742,9 +742,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     // - The "timestampOffset" for the start of this segment is less than
```
mjneil: Comment should be updated to reflect changes

test/segment-loader.test.js
gkatsev: apparently, nothing.

gkatsev:
Can't come up with a better test, unfortunately.

PR #506, opened 2019-05-10
fix(HLSe): slice keys properly on IE11

gkatsev:
IE11 doesn't support slice on TypedArrays, instead, do it manually if
the TypedArray doesn't have slice.


misteroneill:
Double-checked with a real IE11/Win10 - looks good.

PR #512, opened 2019-05-14
fix: do not always set ended_ to false on sourceopen

brandonocasey:
## Description
Currently the way `endOfStream` works in the tba branch:

1. Does not make much sense as most `sourceBuffer` and `mediaSource` operations go through the `sourceUpdater` now
2. `endOfStream` isn't called correctly whenever a rendition switch or audio switch happens at the end of a stream.

This full request fixes that.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -451,6 +451,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('ended', () => {
+      this.logger_('main segment loader ended');
       this.onEndOfStream();
     });
 
@@ -484,6 +485,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.audioSegmentLoader_.on('ended', () => {
+      this.logger_('audioSegmentLoader ended');
       this.onEndOfStream();
     });
   }
@@ -651,7 +653,12 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.tryToCreateSourceBuffers_();
     } catch (e) {
       videojs.log.warn('Failed to create Source Buffers', e);
-      return this.mediaSource.endOfStream('decode');
+      if (this.mediaSource.readyState !== 'open') {
+        this.trigger('error');
+      } else {
+        this.sourceUpdater_.endOfStream('decode');
+      }
+      return;
     }
 
     // if autoplay is enabled, begin playback. This is duplicative of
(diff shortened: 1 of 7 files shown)


src/master-playlist-controller.js
```diff
@@ -651,7 +653,12 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.tryToCreateSourceBuffers_();
     } catch (e) {
       videojs.log.warn('Failed to create Source Buffers', e);
-      return this.mediaSource.endOfStream('decode');
+      if (this.mediaSource.readyState !== 'open') {
+        this.trigger('error');
+      } else {
+        this.sourceUpdater_.endOfStream('decode');
```
brandonocasey: Previously an error would not be shown if endOfStream failed. `endOfStream` should no longer fail.

src/segment-loader.js
```diff
@@ -25,36 +25,6 @@ import { gopsSafeToAlignWith, removeGopBuffer, updateGopBuffer } from './util/go
...
- * @returns {Boolean} do we need to call endOfStream on the MediaSource
- */
-const detectEndOfStream = function(playlist, mediaSource, segmentIndex) {
-  if (!playlist || !mediaSource) {
-    return false;
-  }
-
-  let segments = playlist.segments;
-
-  // determine a few boolean values to help make the branch below easier
```
brandonocasey: This was moved into `this.isEndOfStream_`

src/segment-loader.js
```diff
@@ -246,7 +216,9 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.syncController_.on('syncinfoupdate', () => this.trigger('syncinfoupdate'));
 
     this.mediaSource_.addEventListener('sourceopen', () => {
-      this.ended_ = false;
+      if (!this.isEndOfStream_()) {
```
brandonocasey: This prevents `ended_` from being set to false when this segment loader is still ended, but another opens the mediaSource to append more data. IE this can happen when switching audio track after end of stream.

src/segment-loader.js
```diff
@@ -836,11 +808,6 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
-    if (this.isEndOfStream_(segmentInfo.mediaIndex)) {
```
brandonocasey: It does not make sense to end a stream, before we append anything

test/segment-loader.test.js
```diff
@@ -891,45 +891,6 @@ QUnit.module('SegmentLoader', function(hooks) {
       });
     });
 
-    QUnit.test('endOfStream does not happen while sourceUpdater is updating', function(assert) {
-      let endOfStreams = 0;
-      let bandwidthupdates = 0;
-
-      return setupMediaSource(loader.mediaSource_, loader.sourceUpdater_).then(() => {
```
brandonocasey: This test is no longer valid

test/videojs-http-streaming.test.js
```diff
@@ -630,17 +630,15 @@ function(assert) {
...
+
   this.player.tech_.hls.masterPlaylistController_.mediaSource.endOfStream = (type) => {
     endOfStreams.push(type);
-    throw new Error();
   };
 
   this.player.on('error', () => {
     const error = this.player.error();
 
-    assert.equal(endOfStreams.length, 1, 'one endOfStream called');
```
brandonocasey: we either call end of stream or we don't now.

PR #519, opened 2019-05-17
fix: fmp4 playback

brandonocasey:
Fixed fmp4 playback via mux.js mp4 codec parsing among other things.


package.json
```diff
@@ -35,6 +35,7 @@
     "build:test:segments": "node -e \"var b=require('./scripts/segments-data.js'); b.build();\"",
     "watch": "run-p watch:js watch:test:manifest watch:test:segments watch:test:js watch:js:switcher",
     "watch:js": "rollup -c scripts/rollup.config.js -w",
+    "watch:webworker": "rollup -c scripts/webworker.rollup.config.js -w",
```
brandonocasey: webworkers were not being watched for rebuild, which made mux.js dev work annoying, this makes it so that we watch and rebuild them.

ldayananda: nice!

src/master-playlist-controller.js
```diff
@@ -1161,22 +1161,20 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
 
-    const hasVideo = this.mainSegmentLoader_.startingMedia_.hasVideo;
-    const hasAudio = this.mainSegmentLoader_.startingMedia_.hasAudio ||
-      // alt audio always has audio
-      this.mediaTypes_.AUDIO.activePlaylistLoader;
+    const startingMedia = this.mainSegmentLoader_.startingMedia_;
+    const {hasVideo, videoCodec} = startingMedia;
     const media = this.masterPlaylistLoader_.media();
+    // alt audio always has audio
+    // TODO: should we wait for `trackinfo` from audioSegmentLoader?
```
ldayananda: I think just having the codec info is enough to create the source buffers

src/media-segment-request.js
```diff
@@ -170,8 +170,18 @@ const handleInitSegmentResponse =
     captionParser.init();
   }
 
-  segment.map.timescales = mp4probe.timescale(segment.map.bytes);
-  segment.map.videoTrackIds = mp4probe.videoTrackIds(segment.map.bytes);
```
ldayananda: you'll definitely want to test with inband captions

src/media-segment-request.js
```diff
@@ -223,16 +233,15 @@ const handleSegmentResponse = ({
 
   // This is likely an FMP4 and has the init segment.
   // Run through the CaptionParser in case there are captions.
-  if (captionParser && segment.map && segment.map.bytes) {
+  if (captionParser && segment.map && segment.map.track && segment.map.track.video) {
```
ldayananda: does this ensure we have init segment data already?

src/media-segment-request.js
```diff
@@ -350,39 +359,45 @@ const handleSegmentBytes = ({
...
     // segment here
-    const startTime = probeMp4StartTime(bytesAsUint8Array, segment.map.bytes);
-
-    // we don't parse fmp4 (yet), so we can't provide any track info, however, the
-    // track info callback should still be called before the dataFn is called
-    trackInfoFn(segment, null);
-    // use null for the media type since we don't technically know whether it's audio or
-    // video
+    trackInfoFn(segment, {
+      hasVideo: videoTrack ? true : false,
```
ldayananda: `typeof videoTrack !== undefined`?

src/media-segment-request.js
```diff
@@ -350,39 +359,45 @@ const handleSegmentBytes = ({
...
+      hasAudio: audioTrack ? true : false,
+      audioCodec: (audioTrack || {}).codec,
```
ldayananda: is it bad for the codec to be set to {}.codec? In other words, `undefined`? I don't think we should allow that

src/media-segment-request.js
brandonocasey: yeah

src/master-playlist-controller.js
brandonocasey: The problem is when we have a video segment loader that just has video and and audio segment loader that just has audio. We want to create two source buffers, but right now we will only create one.

ldayananda: hmm, is this different to the pre-TB4A behavior?

brandonocasey: previously we just use the codec from the playlist, now we use the codec that we parse from the mp4 init segment, if available.

src/media-segment-request.js
brandonocasey: I'm not sure there is much of a difference between not having the `audioCodec` key in the track info object and having it set to `undefined`.

src/master-playlist-controller.js
ldayananda: ah, yeah, but I mean why would we need to wait for `trackinfo` from the audioSegmentLoader?

src/media-segment-request.js
ldayananda: It may get confusing, especially if we accidentally end up trying to create a sourceBuffer with `undefined` codec

src/master-playlist-controller.js
brandonocasey: If we don't have codec info in the playlist and we have separate audio/video content The first `trackinfo` will fire for either audio or video. Which will mean that we create one sourceBuffer for audio or video. Then the other trackinfo will be ignored because sourceBuffers were already created.

ldayananda: I see. Is that a problem with the current implementation? and I assume looking at the tracks defined in the init segment will help with this?

brandonocasey: This was a problem with the current implementation yeah. Looking at the tracks codecs fixes this, but we have to wait for two `trackinfo` events. I updated the code to do that, so we should be good to go.

src/media-segment-request.js
brandonocasey: In that case though `hasVideo` will either be `false` or we will use the `playlist` or `default` codec.

ldayananda: ah, yeah I guess that would make sense for debugging too

src/media-segment-request.js
```diff
@@ -350,38 +364,58 @@ const handleSegmentBytes = ({
...
+    const timingInfo = mp4probe.startTime(segment.map.timescales, bytesAsUint8Array);
+
+    if (trackInfo.hasAudio) {
+      timingInfoFn(segment, 'audio', 'start', timingInfo);
+    }
+
+    if (trackInfo.hasVideo) {
+      timingInfoFn(segment, 'video', 'start', timingInfo);
+    }
+    dataFn(segment, {data: bytes, type: trackInfo.hasAudio ? 'audio' : 'video'});
```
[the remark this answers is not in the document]
ldayananda: I guess we assume fmp4 is always demuxed and that's why this works. A comment here would be helpful!

src/util/segment.js
```diff
@@ -35,8 +35,8 @@ export const probeTsSegment = (bytes, baseStartTime) => {
 
   const result = {
     // each type's time info comes back as an array of 2 times, start and end
-    hasVideo: timeInfo.video && timeInfo.video.length === 2,
-    hasAudio: timeInfo.audio && timeInfo.audio.length === 2
+    hasVideo: timeInfo.video && timeInfo.video.length === 2 || false,
```
ldayananda: why is this needed?

test/playback.test.js
```diff
@@ -28,6 +28,7 @@ QUnit.module('Playback', {
     let done = assert.async();
     let video = document.createElement('video-js');
 
+    // videojs.log.level('debug');
```
ldayananda: accident?

test/playback.test.js
```diff
@@ -90,7 +91,7 @@ QUnit.test('replay', function(assert) {
   });
 });
 
-QUnit.skip('playlist with fmp4 and ts segments', function(assert) {
+QUnit.test('playlist with fmp4 segments', function(assert) {
```
ldayananda: this test was intended to be for a playlist with mixed ts and fmp4 segments, this change is **not** equivalent

src/media-segment-request.js
```diff
@@ -223,16 +238,15 @@ const handleSegmentResponse = ({
 
   // This is likely an FMP4 and has the init segment.
   // Run through the CaptionParser in case there are captions.
-  if (captionParser && segment.map && segment.map.bytes) {
+  if (captionParser && segment.map && segment.map.track && segment.map.track.video) {
```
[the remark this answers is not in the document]
brandonocasey: Yeah I took all this code out in the doubled captions pull request, but I can remove it here too.

src/media-segment-request.js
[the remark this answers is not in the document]
brandonocasey: If `track.hasAudio` is `true` at this point it is audio only, since if we have both audio and video we set `hasAudio` to `false` and merge the `audioCodec` into the `videoCodec`. I will add a comment though.

src/util/segment.js
brandonocasey: previously this will result in `hasVideo` and `hasAudio` being set to undefined, rather than a `boolean` value of `false`.

test/playback.test.js
brandonocasey: I think that this code is helpful for debugging, and we should keep the comment there.

test/playback.test.js
brandonocasey: agreed, I have an issue open for fixing mixed ts and fmp4 segments. For now it seemed like a test for only fmp4 would be OK.

src/segment-loader.js
```diff
@@ -1232,6 +1226,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       // this case, just use the loader type, since fmp4 should always be demuxed.
       mediaType =
         this.loaderType_ === 'main' && this.startingMedia_.hasVideo ? 'video' : 'audio';
+      timeType = mediaType;
```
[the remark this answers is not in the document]
brandonocasey: Removed this if statement and added a `logger_`

src/media-segment-request.js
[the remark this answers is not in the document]
ldayananda: Yeah, I meant that if you have a ts segment, you could have a muxed stream and alt audio in the one segment. But, since we're dealing with fmp4 here we assume it's demuxed so if `hasAudio` is true, it must be alt audio

src/util/segment.js
ldayananda: Oh, if `timeInfo.video` was undefined, `hasVideo` would be undefined? Gotcha, makes more sense.

test/playback.test.js
ldayananda: seems a bit strange to keep a comment in tests though 🤔 It might make sense to break it out as an improvement and have a path to running the tests with debug logging turned on

ldayananda: but fine to keep it for now

test/playback.test.js
ldayananda: it's theoretically fine but you will need to rewrite this test later when mixed fmp4 and ts is fixed, either works.

gkatsev:
I've been testing this and it's working great. We should generate or modify a playlist that specifically exercises the changes in here.

PR #530, opened 2019-05-30
fix: Prevent double 608 captions on rendition change

brandonocasey:
608 captions can currently "multiply", ie The same cue and text for a different rendition is added and both the caption from the old rendition and the new rendition are shown at the same time. This happens because:

1. We switch renditions segments are downloaded and parsed for captions.
2. We add the cues for those captions to our tracks.
3. If we already had similar cues from the previous rendition, we don't do anything about it. This causes duplicated captions!

The Fix:
Before adding caption cues to a track, we remove any existing caption cues at those times.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -690,6 +690,10 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.metadataQueue_.id3 = [];
     this.metadataQueue_.caption = [];
     this.abort();
+
+    if (this.captionParser_) {
+      this.captionParser_.clearParsedCaptions();
+    }
   }
 
   /**
@@ -1212,7 +1216,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     }
   }
 
-  handleCaptions_(simpleSegment, captions) {
+  handleCaptions_(simpleSegment, captionData) {
     if (this.checkForAbort_(simpleSegment.requestId) ||
       this.abortRequestEarly_(simpleSegment.stats)) {
       return;
@@ -1220,7 +1224,7 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     // This could only happen with fmp4 segments, but
     // should still not happen in general
-    if (captions.length === 0) {
+    if (captionData.length === 0) {
       this.logger_('SegmentLoader received no captions from a caption event');
       return;
     }
@@ -1231,7 +1235,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     // can be adjusted by the timestamp offset
     if (!segmentInfo.hasAppendedData_) {
       this.metadataQueue_.caption.push(
-        this.handleCaptions_.bind(this, simpleSegment, captions));
+        this.handleCaptions_.bind(this, simpleSegment, captionData));
       return;
     }
 
@@ -1239,14 +1243,39 @@ export default class SegmentLoader extends videojs.EventTarget {
       this.sourceUpdater_.audioTimestampOffset() :
       this.sourceUpdater_.videoTimestampOffset();
 
-    captions.forEach((caption) => {
-      createCaptionsTrackIfNotExists(
-        this.inbandTextTracks_, this.hls_.tech_, caption.stream);
-      addCaptionData({
-        captionArray: [caption],
-        inbandTextTracks: this.inbandTextTracks_,
-        timestampOffset
-      });
+    const tracks = {};
+
+    // get total start/end and captions for each track/stream
+    captionData.forEach((caption) => {
+      // caption.stream is actually a track name...
+      // set to the existing values in tracks or default values
+      tracks[caption.stream] = tracks[caption.stream] || {
+        startTime: Infinity,
+        captions: [],
+        endTime: 0
+      };
+
+      const track = tracks[caption.stream];
+
+      track.startTime = Math.min(track.startTime, (caption.startTime + timestampOffset));
+      track.endTime = Math.max(track.endTime, (caption.endTime + timestampOffset));
+      track.captions.push(caption);
+    });
+
+    Object.keys(tracks).forEach((trackName) => {
+      const {startTime, endTime, captions} = tracks[trackName];
+      const inbandTextTracks = this.inbandTextTracks_;
+
+      this.logger_(`adding cues from ${startTime} -> ${endTime} for ${trackName}`);
+
+      createCaptionsTrackIfNotExists(inbandTextTracks, this.hls_.tech_, trackName);
+      // clear out any cues that start and end at the same time period for the same track.
+      // We do this because a rendition change that also changes the timescale for captions
+      // will result in captions being re-parsed for certain segments. If we add them again
+      // without clearing we will have two of the same captions visible.
+      removeCuesFromTrack(startTime, endTime, inbandTextTracks[trackName]);
+
+      addCaptionData({captionArray: captions, inbandTextTracks, timestampOffset});
     });
 
     // Reset stored captions since we added parsed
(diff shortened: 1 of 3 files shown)


src/media-segment-request.js
```diff
@@ -236,23 +229,6 @@ const handleSegmentResponse = ({
     segment.bytes = new Uint8Array(newBytes);
   }
 
-  // This is likely an FMP4 and has the init segment.
-  // Run through the CaptionParser in case there are captions.
-  if (captionParser && segment.map && segment.map.track && segment.map.track.video) {
-    // Initialize CaptionParser if it hasn't been yet
-    if (!captionParser.isInitialized()) {
```
brandonocasey: Everything I removed in this file was redundant and we never hit it. I think I messed up merging from master during the intial tba/lhls pr

src/segment-loader.js
```diff
@@ -715,6 +715,10 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.metadataQueue_.id3 = [];
     this.metadataQueue_.caption = [];
     this.abort();
+
+    if (this.captionParser_) {
```
brandonocasey: Clear parsed captions on a rendition switch.

src/segment-loader.js
```diff
@@ -1260,20 +1264,44 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      const {startTime, endTime, captions} = tracks[trackName];
+      const {inbandTextTracks_} = this;
+
+      this.logger_(`adding cues from ${startTime} -> ${endTime} for ${trackName}`);
+
+      createCaptionsTrackIfNotExists(inbandTextTracks_, this.hls_.tech_, trackName);
+      // clear out any old cues. We do this because a rendition change
+      // will cause captions to be re-parsed for certain segments, as they are
+      // downloaded, but we do not want to have two of the same captions visible.
+      removeCuesFromTrack(startTime, endTime, inbandTextTracks_[trackName]);
```
brandonocasey: This is the actual fix, but there was a lot of needed refactor to get here.

src/vtt-segment-loader.js
```diff
@@ -32,6 +32,8 @@ export default class VTTSegmentLoader extends SegmentLoader {
     this.mediaSource_ = null;
 
     this.subtitlesTrack_ = null;
+
+    this.loaderType_ = 'subtitle';
```
brandonocasey: Not entirely sure if we want this in this pull request, but it's a bit weird for the `subtitleSegmentLoader_` to list itself as `main`. I had a hell of a time before I realized that `vtt-segment-loader` has nothing to do with 608 captions and all of that is done in the `mainSegmentLoader`

src/media-segment-request.js
```diff
@@ -165,11 +165,6 @@ const handleInitSegmentResponse =
 
   segment.map.bytes = new Uint8Array(request.response);
 
-  // Initialize CaptionParser if it hasn't been yet
```
ldayananda: I think we were supposed to keep this one. Theoretically we can receive the init segment and first segment in any order, whichever is received first should init the caption parser.

ldayananda: Although maybe this was in SL originally...

src/segment-loader.js
```diff
@@ -1260,20 +1264,44 @@ export default class SegmentLoader extends videojs.EventTarget {
...
 
-    captions.forEach((caption) => {
-      createCaptionsTrackIfNotExists(
-        this.inbandTextTracks_, this.hls_.tech_, caption.stream);
+    const tracks = {};
+
+    // get total start/end and captions for each track/stream
+    captionData.forEach((caption) => {
+      // caption.stream is actually a track name...
+      const track = tracks[caption.stream] = tracks[caption.stream] || {
```
ldayananda: `tracks[caption.stream] = tracks[caption.stream]` is this part correct?

src/segment-loader.js
ldayananda: won't this end up clearing out all the cues except the most recently parsed caption cue if you don't rendition switch?

src/segment-loader.js
```diff
@@ -1260,20 +1264,44 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      createCaptionsTrackIfNotExists(inbandTextTracks_, this.hls_.tech_, trackName);
+      // clear out any old cues. We do this because a rendition change
+      // will cause captions to be re-parsed for certain segments, as they are
+      // downloaded, but we do not want to have two of the same captions visible.
+      removeCuesFromTrack(startTime, endTime, inbandTextTracks_[trackName]);
+
       addCaptionData({
-        captionArray: [caption],
-        inbandTextTracks: this.inbandTextTracks_,
+        captionArray: captions,
```
ldayananda: I originally had this forced to be an array to have consistency between using the `PartialTransmuxer` (which outputs each caption as it is seen, an object) and `FullTransmuxer` (which outputs all the captions for a segment)

src/media-segment-request.js
ldayananda: Nevermind, I see this is now done in `handleSegmentBytes` which looks like the right way to do this

src/segment-loader.js
brandonocasey: This clears out all cues in the range of time that the current cues are going to be inserted. IE: If we have cues from 10-20s and we get more cues from 10-20s we clear the previous cues from 10-20s before inserting the new ones. From what I have seen this only happens during a rendition switch with a timescale change.

src/segment-loader.js
brandonocasey: It will always be an array now due to `track[trackName].captions` being an array now.

src/segment-loader.js
brandonocasey: I cleaned up this portion, and added a comment

src/segment-loader.js
```diff
@@ -1231,22 +1235,47 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    const tracks = {};
+
+    // get total start/end and captions for each track/stream
+    captionData.forEach((caption) => {
+      // caption.stream is actually a track name...
+      // set to the existing values in tracks or default values
+      tracks[caption.stream] = tracks[caption.stream] || {
+        startTime: Infinity,
+        captions: [],
+        endTime: 0
```
misteroneill: If these are defaults, should the values for `startTime` and `endTime` here be swapped?

src/segment-loader.js
```diff
@@ -1231,22 +1235,47 @@ export default class SegmentLoader extends videojs.EventTarget {
...
 
-    captions.forEach((caption) => {
-      createCaptionsTrackIfNotExists(
-        this.inbandTextTracks_, this.hls_.tech_, caption.stream);
-      addCaptionData({
-        captionArray: [caption],
-        inbandTextTracks: this.inbandTextTracks_,
-        timestampOffset
-      });
+    const tracks = {};
```
ldayananda: maybe it would be less confusing to call this `captionTracks` so it's less likely to be confused with video/audio tracks

src/segment-loader.js
ldayananda: these defaults are a clever way of getting lines 1260-1261 to be accurate

ldayananda:
Actually, looks like one of the test failures isn't a timeout, and looks related

src/segment-loader.js
brandonocasey: done

src/segment-loader.js
brandonocasey: added a comment to clarify this

misteroneill: Gotcha, gotcha. It just seemed backwards. I get it now.

PR #542, opened 2019-06-11
fix: Return the correct buffer for each segmentloader

brandonocasey:
With the current implementation a separate video and audio segment loader will try to use the combined buffered value, this causes all kinds of issues. For high bit rate sources we see a lot of append over max on source buffer errors.

### DIFF
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -150,6 +150,7 @@
       var urls = [
         'node_modules/video.js/dist/alt/video.core',
         'node_modules/videojs-contrib-eme/dist/videojs-contrib-eme',
+        'node_modules/videojs-contrib-quality-levels/dist/videojs-contrib-quality-levels',
         'dist/videojs-http-streaming'
       ].map(function(url) {
         return url + (event.target.checked ? '.min' : '') + '.js';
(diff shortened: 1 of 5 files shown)


brandonocasey:
Added the source I used for testing, although you will have to force the 4k rendition to get it to happen.

gkatsev:
Would be nice to add the quality-levels plugin on this page as well. I added `'node_modules/videojs-contrib-quality-levels/dist/videojs-contrib-quality-levels'` to the list of scripts in the change handler thingy. I then used `Array.from(player.qualityLevels()).forEach(l=>l.enabled = l.height > 1000)` to choose the 4k rendition

scripts/index.js
```diff
@@ -150,6 +150,7 @@
       var urls = [
         'node_modules/video.js/dist/alt/video.core',
         'node_modules/videojs-contrib-eme/dist/videojs-contrib-eme',
+        'node_modules/videojs-contrib-quality-levels/dist/videojs-contrib-quality-levels',
```
ldayananda: this breaks the Netlify build, probably because `videojs-contrib-quality-levels` would have to be added to the devDependencies

gkatsev: Ah, whoops. Needs to be copied over in the netlify script

gkatsev: just needs to be added here https://github.com/videojs/http-streaming/blob/0d8eb578d166dbc943794bf4ecf1c3a538730e48/scripts/netlify.js#L5-L10

brandonocasey: interesting will do

brandonocasey: is there any reason that we don't just deploy out of the root project directory?

gkatsev:
The only failing test is the live dash playback one. It is also failing unreliably, even if most of the time. When applying #545 it can reliably pass.
Going to go ahead a merge this particularly since master is frozen until all the TBA/LHLS work is in.

PR #544, opened 2019-06-12
fix: issues with the demo in ie11 and firefox

brandonocasey:
## Description
* Adds back in url type specifier
* Fixes issues and styling in firefox
* fixes dispatch event in ie 11

### DIFF
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -1,22 +1,35 @@
 /* eslint-disable no-var, object-shorthand, no-console */
 (function(window) {
   // all relevant elements
-  var urlForm = document.getElementById('load-url');
+  var urlButton = document.getElementById('load-url');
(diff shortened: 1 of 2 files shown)


gkatsev:
This isn't working for me locally or on netlify

scripts/index.js
```diff
@@ -19,6 +19,19 @@
     console.warn('unhandled input type ' + el.type);
   };
 
+  var newEvent = function(name) {
+    var event
+
+    if (typeof window.Event === 'function') {
+      event = new window.Event('submit');
```
gkatsev: I think this should be `name` not `'submit'`

brandonocasey: yep

gkatsev:
Tests didn't pass but no actual code that is tested was changed.

PR #545, opened 2019-06-12
fix: delay first audio appends when we have not done a video append

brandonocasey:
## Description
When audio is appended before video the media source timestamp may be off which causes playback to stall after it starts. It can also make it seem like the video is frozen for the first few seconds of playback. This is the fix.

### DIFF
--- a/src/source-updater.js
+++ b/src/source-updater.js
@@ -204,6 +204,7 @@ export default class SourceUpdater extends videojs.EventTarget {
       audio: null,
       video: null
     };
+    this.delayedAudioAppendQueue_ = [];
   }
 
   ready() {
@@ -294,6 +295,12 @@ export default class SourceUpdater extends videojs.EventTarget {
       };
     }
 
+    if (type === 'audio' && this.videoBuffer && !this.videoBuffer.appendQueued) {
+      this.delayedAudioAppendQueue_.push([{type, bytes, videoSegmentTimingInfoCallback}, doneFn]);
+      this.logger_(`delayed audio append of ${bytes.length} until video append`);
+      return;
+    }
+
     pushQueue({
       type,
       sourceUpdater: this,
@@ -301,6 +308,21 @@ export default class SourceUpdater extends videojs.EventTarget {
       doneFn,
       name: 'appendBuffer'
     });
+
+    if (type === 'video') {
+      this.videoBuffer.appendQueued = true;
+      if (!this.delayedAudioAppendQueue_.length) {
+        return;
+      }
+      const queue = this.delayedAudioAppendQueue_.slice();
+
+      this.logger_(`queuing delayed audio ${queue.length} appendBuffers`);
+
+      this.delayedAudioAppendQueue_.length = 0;
+      queue.forEach((que) => {
+        this.appendBuffer.apply(this, que);
+      });
+    }
   }
 
   audioBuffered() {
--- a/src/source-updater.js
+++ b/src/source-updater.js
@@ -295,7 +295,7 @@ export default class SourceUpdater extends videojs.EventTarget {
       };
     }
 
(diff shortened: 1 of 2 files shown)


src/source-updater.js
```diff
@@ -294,13 +295,34 @@ export default class SourceUpdater extends videojs.EventTarget {
...
     pushQueue({
       type,
       sourceUpdater: this,
       action,
       doneFn,
       name: 'appendBuffer'
     });
+
+    if (type === 'video') {
+      this.videoAppendQueued = true;
```
gkatsev: we should default this to `false` in the constructor as well.

Does it make sense to reset this value anywhere beyond dispose?

test/source-updater.test.js
```diff
@@ -356,10 +356,10 @@ QUnit.test('removeAudio does not remove video buffer', function(assert) {
     video: 'avc1.4D001E'
   });
 
-  this.sourceUpdater.appendBuffer({type: 'audio', bytes: mp4Audio()}, () => {
```
gkatsev: should this test still not append audio first?

src/source-updater.js
brandonocasey: I will add it to the constructor. I don't think we need to set it anywhere else because we should always be disposing the source updater and creating another when we need to work with different/new source buffers.

test/source-updater.test.js
brandonocasey: 🤔

brandonocasey: Since audio won't be appended until video is, the callback won't be run, I think the test is still testing the same thing though since it always waited for both callbacks anyway.

gkatsev: Ah, I see. Is it possible to make a test that exercises delaying appending the audio buffer?

brandonocasey: Yeah let me right a test for the new behavior

brandonocasey: Done

test/source-updater.test.js
```diff
@@ -1055,3 +1056,28 @@ QUnit.test('dispose removes sourceopen listener', function(assert) {
   });
 });
 
+QUnit.test('audio appends are delayed until video append for the first append', function(assert) {
```
brandonocasey: new test for this

test/karma.conf.js
```diff
@@ -103,7 +103,7 @@ module.exports = function(config) {
     autoWatch: false,
     singleRun: true,
     concurrency: 1,
-    captureTimeout: 300000,
+    captureTimeout: 30000,
```
brandonocasey: 300s vs 30s

test/karma.conf.js
```diff
@@ -16,7 +16,7 @@ module.exports = function(config) {
       clearContext: false,
       qunit: {
         showUI: true,
-        testTimeout: 30000
+        testTimeout: 5000
```
brandonocasey: playback tests are coded to 50s in beforeEach, so this is just a default for other tests

.travis.yml
```diff
@@ -1,22 +1,20 @@
...
+# node version is specified using the .nvmrc file
+cache: npm
 before_install:
   - npm install -g greenkeeper-lockfile@1
 before_script:
   # pulseaudio is needed for dockerized chrome since it has a dummy sink and
   # the dummy audio ALSA kernel module isn't loaded by Travis
   # libavcodec54 is needed by firefox for h.264 support
-  - sudo apt-get install --no-install-recommends pulseaudio libavcodec54
+  - sudo apt-get install pulseaudio
```
gkatsev: don't we need libavcodec54 for firefox?

brandonocasey: not in xenial it seems, but regardless I am backing out all the test changes. I just can't get things to work and they don't really belong in this branch.

gkatsev:
I can get the tests passing locally.

PR #555, opened 2019-06-17
fix: faster dev build

brandonocasey:
## Description
Noticed that dev builds for vhs are very slow, Seems like we run all of build before `npm run start` which means we build everything, then we start the server, then we watch which builds everything again. I think we should just start the server and build at the same time.


scripts/switcher.rollup.config.js
```diff
@@ -17,6 +20,7 @@ export default {
     format: 'umd'
   },
   plugins: [
+    json(),
```
brandonocasey: I think this was previously hidden, since rollup was going into a screen, `package.json` from `videojs-http-streaming` was not being imported correctly.

utils/switcher/run-simulation.js
```diff
@@ -5,7 +5,7 @@ import {
   openMediaSource,
   standardXHRResponse,
 } from '../../test/test-helpers';
-import {Hls} from '../../';
+import {Hls} from '../../src/videojs-http-streaming.js';
```
brandonocasey: I think this was previously hidden, since rollup was going into a screen, seems like we need to be more explicit about our import.

scripts/test.rollup.config.js
```diff
@@ -5,6 +5,9 @@ import resolve from 'rollup-plugin-node-resolve';
 import worker from '@gkatsev/rollup-plugin-bundle-worker';
 import multiEntry from "rollup-plugin-multi-entry";
 
+/* to prevent going into a screen during rollup */
+process.stderr.isTTY = false;
```
gkatsev: Ideally, we'll update rollup so we can set the clear screen option to false, but shouldn't hold up this PR.

package.json
```diff
@@ -33,9 +32,11 @@
     "build:test:js": "rollup -c scripts/test.rollup.config.js",
     "build:test:manifest": "node -e \"var b=require('./scripts/manifest-data.js'); b.build();\"",
     "build:test:segments": "node -e \"var b=require('./scripts/segments-data.js'); b.build();\"",
-    "watch": "run-p watch:js watch:test:manifest watch:test:segments watch:test:js watch:js:switcher",
+    "prewatch": "npm run clean",
+    "watch": "run-p watch:webworker watch:js watch:test:manifest watch:test:segments watch:test:js watch:js:switcher",
```
gkatsev: I don't see a watch:webworker script

scripts/test.rollup.config.js
brandonocasey: right after tba/lhls I planned on updating this project to generator conventions, where possible.

package.json
```diff
@@ -33,9 +32,11 @@
     "build:test:js": "rollup -c scripts/test.rollup.config.js",
     "build:test:manifest": "node -e \"var b=require('./scripts/manifest-data.js'); b.build();\"",
     "build:test:segments": "node -e \"var b=require('./scripts/segments-data.js'); b.build();\"",
-    "watch": "run-p watch:js watch:test:manifest watch:test:segments watch:test:js watch:js:switcher",
+    "prewatch": "npm run clean",
+    "watch": "run-p watch:webworker watch:js watch:test:manifest watch:test:segments watch:test:js",
```
brandonocasey: switcher is currently broken, with tba/lhls, going to add an issue to get it working again, but we don't need to build/rebuild something (takes ~5s) that is broken.

brandonocasey:
Most rollup changes were taken from `videojs-generate-rollup-config`

gkatsev:
Playback tests are reliably failing. This is because some other PRs are missing from here?

brandonocasey:
I would think that is the case, but let me check what happens locally when I bring in the fmp4 playback fixes, since those are the tests that seem to fail.

PR #556, opened 2019-06-19
Feat/fmp4 ts support

brandonocasey:
For fmp4 playlists with ts segment discontinuities the track ids of the fmp4 files and the ts segments that we transmux to fmp4 must match up.

Depends on: 
https://github.com/videojs/mux.js/pull/276


src/media-segment-request.js
```diff
@@ -279,7 +282,9 @@ const transmuxAndNotify = ({
     audioAppendStart: segment.audioAppendStart,
     gopsToAlignWith: segment.gopsToAlignWith,
     isPartial,
+    remux: isMuxed,
     onData: (result) => {
+      result.type = result.type === 'combined' ? 'video' : result.type;
```
brandonocasey: The muxer returns `combined` when `remux` is set to true. We treat `combined` or `muxed` as a `video` buffer.

src/segment-loader.js
```diff
@@ -1515,6 +1515,10 @@ export default class SegmentLoader extends videojs.EventTarget {
       // the segment has a map
       this.appendInitSegment_[type] = map ? true : false;
 
+      // we need to clear out the fmp4 active init segment id, since
+      // we are appending the muxer init segment
+      this.activeInitSegmentId_ = null;
```
brandonocasey: This section of this function only really runs if we transmuxed from a ts to an fmp4. The problem: Example `fmp4-> fmp4 -> ts -> fmp4 -> fmp4`. We need to append an init segment on the first `fmp4`, the `ts`, and the `fmp4` right after the `ts`. If we don't set the `activeInitSegmentId_` to null here we will not append an init segment for the `fmp4` right after the `ts`

src/segment-loader.js
```diff
@@ -1594,7 +1598,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       type,
       initSegment: result.initSegment,
       playlist: segmentInfo.playlist,
-      map: segmentInfo.segment.map
+      map: segmentInfo.isFmp4 ? segmentInfo.segment.map : null
```
brandonocasey: ts segments in an fmp4 playlist will have a map, but they are not fmp4

src/transmuxer-worker.js
```diff
@@ -307,6 +307,10 @@ class MessageHandlers {
     this.transmuxer.setAudioAppendStart(Math.ceil(secondsToVideoTs(data.appendStart)));
   }
 
+  setRemux(data) {
```
brandonocasey: Remux basically means that `mux.js` should output one transmuxed segment for `audio` and `video` data combined.

Issue #557, opened 2019-06-21
DASH doesn't follow <Location>


TomHetmer:
## Description
DASH playlist are loaded from the same (invalid) location instead of following <Location> redirects in the manifest.
## Sources
http://1425889403.rsc.cdn77.org/ls-45420-1/rewind-60.mpd

## Steps to reproduce
Have a DASH manifest that always contains ie.
`<Location>https://xxx/rewind-60.mpd?resume=1514764800:1561118402.431117</Location>`
The location changes for each request.
Video.js keeps using the main URL and fails to load any new segments.

### Expected
Video.js will load the mpd from the updated URL every time.

### videojs-http-streaming version
videojs-http-streaming 1.9.3

### videojs version
video.js 7.5.5

nzt4567:
Hi, this happens to us as well, any chance on taking a look at this please? We wanted to switch to video.js from shaka in our new projects but this currently seems to be a dealbreaker 😢

Issue #457, opened 2019-04-08
Sidx is activated for playlists which have no sidx


couchcrew-thomas:
## Description
The DASH LIVE stream stalls after the first segment was played. The player continues loading the manifest in an endless loop. It works fine on Shaka and DASH.js players.
Test case: https://jsbin.com/yaribiqipu/edit?html,output

## Sources
DASH LIVE: https://demo-dash-live.zahs.tv/sd/manifest.mpd

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.

1. Load the stream
2. Start playback
3. Wait for the first segment to finish

### Expected
The live stream continues to play after the first segment.

### Error output
There are no console errors.

## Additional Information
Shaka player and DASH.js player can play the stream just fine.

### videojs-http-streaming version
what version of videojs-http-streaming does this occur with?
videojs-http-streaming 

### videojs version
what version of videojs does this occur with?
video.js 7.4.1

### Browsers
all latest browsers

### Other Plugins
none, just video.js with built in VHS.

### Other JavaScript
none

vksbansal:
I am to reproduce the issue 100 present

steps:

1. pause live video **player.pause()**
2. seek to some point **player.currentTime(seekPoint);**
3. play the video **player.play()**
4. video will start to play from the live location, not from the seek point.

Please help, please.... its a major bug.

Hope you will pay attention

Many thanks,
Vikas Bansal

squarebracket:
Something seems to be going wrong here when parsing the playlist, as it thinks it should be using sidx but then gets tripped up by the lack of sidxInfo.

PR #561, opened 2019-06-24
fix: Sidx is activated for playlists which have no sidx #457

marceljager:
## Description
Fix bug when sidx was activated even for playlists without sidx
https://github.com/videojs/http-streaming/issues/457 + https://github.com/videojs/http-streaming/issues/557

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -681,9 +681,10 @@ export default class DashPlaylistLoader extends EventTarget {
 
       const master = this.parseMasterXml();
       const updatedMaster = updateMaster(this.master, master);
+      const currentSidxInfo = this.media().sidx;
 
-      if (updatedMaster) {
-        const sidxKey = generateSidxKey(this.media().sidx);
+      if (updatedMaster && currentSidxInfo) {
+        const sidxKey = generateSidxKey(currentSidxInfo);
 
         // the sidx was updated, so the previous mapping was removed
         if (!this.sidxMapping_[sidxKey]) {
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -308,8 +308,8 @@ export default class DashPlaylistLoader extends EventTarget {
 
     // switch to previously loaded playlists immediately
     if (mediaChange &&
-        this.loadedPlaylists_[playlist.uri] &&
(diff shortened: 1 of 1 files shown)


gkatsev:
I'm not sure I understand how this helps with #557. Can you elaborate?

src/dash-playlist-loader.js
```diff
@@ -681,9 +681,10 @@ export default class DashPlaylistLoader extends EventTarget {
 
       const master = this.parseMasterXml();
       const updatedMaster = updateMaster(this.master, master);
+      const currentSidxInfo = this.media().sidx;
 
-      if (updatedMaster) {
-        const sidxKey = generateSidxKey(this.media().sidx);
+      if (updatedMaster && currentSidxInfo) {
```
gkatsev: at first glance just this change makes sense but then you realize that line 717/718 should be run when we have an updatedMaster but no sidx.
Maybe we need to move the generateSidxKey inside a a currentSidxInfo check inside the updatedMaster check? If that makes sense.

ldayananda: I think this change makes sense. If there's no sidx, the current master should be updated to be the `updatedMaster`

ldayananda: Ah, sorry I think it would make sense if the change is to line 689/690:

``` if (currentSidxInfo && !this.sidxMapping_[sidxKey]```

we might then need to have `sidxKey` only run when there's a sidx as @gkatsev suggested.

```js
if (updatedMaster) {
  if (currentSidxInfo) {
     const sidxKey = generateSidxKey(this.media().sidx); 

     if (!this.sidxMapping_[sidxKey]) {
```

marceljager: Thanks, that make sense. I fixed it

marceljager:
> I'm not sure I understand how this helps with #557. Can you elaborate?

The first segment for the file was loaded but second segment was not loading because of sidx error. After this fix everything works just fine. So there is no problem with Location

gkatsev:
I can see the stream from #557 working with this change, however, I'm still confused regarding `<Location>` because I don't think we do anything with it and the network tab doesn't seem to be using that domain for subsequent requests.

gkatsev:
Published this as 1.10.4. Thanks!

PR #564, opened 2019-06-25
fix(subtitles): Consider `hidden` tracks as active

mister-ben:
## Description
`cuechange` events should be triggered for a track with mode `hidden`. Currently a track is only considered active if its mode is `showing`, cues will not be parsed and no cuechange occurs.

## Specific Changes proposed
Check for `hidden` as well as `showing`.

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev:
Hm... I seem to be able to listen to the cuechange even of the 608 track in bipbop with and without this change...

gkatsev:
At least in master, currently.

mister-ben:
That has a default track which starts as `showing`. When you change its mode to `hidden` the cues from the vtt segments which were already parsed still exist on the track and will trigger cuechange. Once you get beyond the end of that range no new cues will be added.

gkatsev:
The 608 track isn't on by default and that's the one I listened to cuechange on.

mister-ben:
Ah, 608. This fixes the vtt segments. With this, they'll behave like the 608 or a remote text track would.

mister-ben:
Shouldn't go stale

PR #565, opened 2019-06-25
test: faster playback tests

brandonocasey:
## Description
Make the playback tests faster using `playbackRate`

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1396,11 +1396,8 @@ export default class SegmentLoader extends videojs.EventTarget {
     // to the initSegment cache
     if (simpleSegment.map) {
       simpleSegment.map = this.initSegmentForMap(simpleSegment.map, true);
-      // move over media segment request init segment additions
+      // move the map bytes onto the segment loader's segment state object
       segmentInfo.segment.map.bytes = simpleSegment.map.bytes;
-      segmentInfo.segment.map.tracks = Object.assign({}, simpleSegment.map.tracks);
-      segmentInfo.segment.map.timescales = Object.assign({}, simpleSegment.map.timescales);
-
     }
 
     // if this request included a segment key, save that data in the cache
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1396,8 +1396,11 @@ export default class SegmentLoader extends videojs.EventTarget {
     // to the initSegment cache
     if (simpleSegment.map) {
       simpleSegment.map = this.initSegmentForMap(simpleSegment.map, true);
-      // move the map bytes onto the segment loader's segment state object
+      // move over media segment request init segment additions
       segmentInfo.segment.map.bytes = simpleSegment.map.bytes;
+      segmentInfo.segment.map.tracks = Object.assign({}, simpleSegment.map.tracks);
+      segmentInfo.segment.map.timescales = Object.assign({}, simpleSegment.map.timescales);
+
     }
 
     // if this request included a segment key, save that data in the cache
--- a/test/playback.test.js
+++ b/test/playback.test.js
@@ -12,14 +12,20 @@ let when = function(element, type, cb, condition) {
   });
 };
 
-let playFor = function(player, time, cb) {
-  let targetTime = player.currentTime() + time;
-
-  when(player, 'timeupdate', cb, () => player.currentTime() >= targetTime);
+const playFor = function(player, time, cb) {
+  if (player.paused()) {
+    player.play();
+  }
+  window.setTimeout(() => {
+    if (player.currentTime() <= time) {
+      return playFor(player, time, cb);
+    }
+    cb();
+  }, 10);
 };
 
 QUnit.module('Playback', {
-  before(assert) {
+  before() {
     this.fixture = document.createElement('div');
     document.body.appendChild(this.fixture);
   },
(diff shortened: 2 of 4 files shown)


src/segment-loader.js
```diff
@@ -1396,8 +1396,11 @@ export default class SegmentLoader extends videojs.EventTarget {
     // to the initSegment cache
     if (simpleSegment.map) {
       simpleSegment.map = this.initSegmentForMap(simpleSegment.map, true);
-      // move the map bytes onto the segment loader's segment state object
+      // move over media segment request init segment additions
       segmentInfo.segment.map.bytes = simpleSegment.map.bytes;
+      segmentInfo.segment.map.tracks = Object.assign({}, simpleSegment.map.tracks);
```
brandonocasey: seems like increasing playbackRate showed a race condition where an alt audio track wouldn't have, tracks or timescales set correctly on a subsequent request.

test/playback.test.js
```diff
@@ -243,6 +249,7 @@ QUnit.test('loops', function(assert) {
       player.vhs.mediaSource.addEventListener('sourceopen', () => {
         assert.ok(true, 'sourceopen triggered after ending stream');
         done();
+        player.loop(false);
```
gkatsev: is this necessary?

test/playback.test.js
```diff
@@ -31,16 +37,23 @@ QUnit.module('Playback', {
     // videojs.log.level('debug');
     video.style = 'display: none;';
 
+    video.setAttribute('controls', '')
+    video.setAttribute('muted', '')
     video.width = 600;
     video.height = 300;
+    video.defaultPlaybackRate = 16;
```
gkatsev: I think it may be useful to still have one simple playback test that runs at a playback rate of 1.

test/playback.test.js
```diff
@@ -105,7 +119,7 @@ QUnit.test('playlist with fmp4 segments', function(assert) {
   });
 
   player.src({
-    src: 'https://storage.googleapis.com/shaka-demo-assets/angel-one-hls/hls.m3u8',
+    src: 'https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_adv_example_hevc/master.m3u8',
```
gkatsev: why the source change?

test/playback.test.js
```diff
@@ -12,14 +12,20 @@ let when = function(element, type, cb, condition) {
   });
 };
 
-let playFor = function(player, time, cb) {
-  let targetTime = player.currentTime() + time;
-
-  when(player, 'timeupdate', cb, () => player.currentTime() >= targetTime);
+const playFor = function(player, time, cb) {
```
gkatsev: the behavior of the `playFor` method changed when removing `targetTime`, you can no longer call `playFor` several times with different values and expect it to play through those seconds.

test/playback.test.js
```diff
@@ -12,14 +12,20 @@ let when = function(element, type, cb, condition) {
   });
 };
 
-let playFor = function(player, time, cb) {
-  let targetTime = player.currentTime() + time;
-
-  when(player, 'timeupdate', cb, () => player.currentTime() >= targetTime);
```
gkatsev: if `when` is no longer used, we can remove it.

test/playback.test.js
```diff
@@ -31,16 +37,23 @@ QUnit.module('Playback', {
...
+    this.player = videojs(video);
+
+    this.player.ready(done, true);
   },
   afterEach() {
     this.player.dispose();
+
+  },
+  after() {
+    document.body.removeChild(this.fixture);
```
gkatsev: huh, apparently we don't have a global-shims file like we do in Video.js that sets up the qunit fixture https://github.com/videojs/video.js/blob/master/test/globals-shim.js

test/playback.test.js
brandonocasey: no, its a holdover from something I was testing.

test/playback.test.js
brandonocasey: The apple source seems to load a lot quicker, but we can go back if we want. It was only added for TBA

test/playback.test.js
brandonocasey: There is a global `qunit-fixture` but its hidden by default. So if you had to go debug a specific unit test player and wanted to view it, it would be rather hard.

gkatsev: Is that available in our version of qunit? It should be simple to show with css, if we wanted. However, I don't have a huge preference. Since this is how it's been going so far, probably fine to just leave it.

test/playback.test.js
brandonocasey: woops fixed

test/playback.test.js
brandonocasey: I will see how easy it is to show, it is also in our version of qunit

gkatsev: Ah, I guess that's a new thing? Guess we can remove it from global-shims in Video.js then.

test/playback.test.js
```diff
@@ -57,9 +54,11 @@ QUnit.module('Playback', {
   }
 });
 
-QUnit.test('Advanced Bip Bop', function(assert) {
+QUnit.test('Advanced Bip Bop default speed', function(assert) {
   let done = assert.async();
 
+  this.player.defaultPlaybackRate(1);
```
gkatsev: fyi, this can also be just `playbackRate`. I don't think it makes a huge difference.

brandonocasey: I tried that, but it seems like if we try to set a playbackRate before we have a source that is ready, we reset to the `defaultPlaybackRate` upon playback.

gkatsev: Ah, makes sense. Carry on then :)

src/segment-loader.js
gkatsev: Maybe this should be in the commit message as well? Or even split the test changes from this?

gkatsev: I don't think we can use `Object.assign` as IE11 doesn't have it.

brandonocasey: I will split it out

PR #568, opened 2019-06-25
fix: prioritize playlist codecs over mux.js codecs, so encrypted content doesn't fail

brandonocasey:
fix: prioritize playlist codecs over mux.js codecs, so encrypted content doesn't fail

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -17,8 +17,7 @@ import Config from './config';
 import {
   parseCodecs,
   mapLegacyAvcCodecs,
-  codecsForPlaylist,
-  translateLegacyCodec
+  codecsForPlaylist
 } from './util/codecs.js';
 import { createMediaTypes, setupMediaGroups } from './media-groups';
 import logger from './util/logger';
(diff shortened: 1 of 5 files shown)


src/media-segment-request.js
```diff
@@ -355,22 +355,21 @@ const handleSegmentBytes = ({
       hasAudio: !!tracks.audio
     };
 
-    if (tracks.audio && tracks.audio.codec) {
+    // if we have a audio track, with a codec that is not set to
+    // encrypted audio
+    if (tracks.audio && tracks.audio.codec && tracks.audio.codec !== 'enca') {
```
brandonocasey: if the codec is enca, there isn't really a codec

src/master-playlist-controller.js
```diff
@@ -1168,19 +1167,34 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
     const media = this.masterPlaylistLoader_.media();
     const playlistCodecs = codecsForPlaylist(this.masterPlaylistLoader_.master, media);
     const codecs = {};
 
-    if (mainStartingMedia.hasAudio || hasAltAudio) {
-      codecs.audio = mainStartingMedia.audioCodec || (audioStartingMedia || {}).audioCodec ||
-        translateLegacyCodec(playlistCodecs.audio) || DEFAULT_AUDIO_CODEC;
-    }
+    // priority of codecs: playlist -> mux.js parsed codecs -> default
+    if (mainStartingMedia.isMuxed) {
```
brandonocasey: For muxed we mix the video and audio codec into one source buffer as a "video" codec

src/util/codecs.js
```diff
@@ -177,23 +167,16 @@ export const codecsForPlaylist = function(master, media) {
...
-        'Attempting to use the default audio codec (mp4a.40.2)');
-      codecInfo.audioProfile = defaultCodecs.audioProfile;
-    }
   }
 
   const codecs = {};
 
   if (codecInfo.videoCodec) {
-    codecs.video = `${codecInfo.videoCodec}${codecInfo.videoObjectTypeIndicator}`;
+    codecs.video = translateLegacyCodec(`${codecInfo.videoCodec}${codecInfo.videoObjectTypeIndicator}`);
```
brandonocasey: We only used `translateLegacyCodec` to translate codecs from this function, seems like we should just use in in here.

src/util/codecs.js
```diff
@@ -3,17 +3,8 @@
  * codec strings, or translating codec strings into objects that can be examined.
  */
 
-import videojs from 'video.js';
 import {findBox} from 'mux.js/lib/mp4/probe';
 
-// Default codec parameters if none were provided for video and/or audio
-const defaultCodecs = {
```
brandonocasey: We no longer return defaults from here, that way we can use mux.js codecs if they exist.

PR #573, opened 2019-06-27
test: Fix a tests that could fail via a race condition

brandonocasey:
## Description
Sometimes this test fails on Firefox, because audio is appended to the audio buffer, before video, even though we call appendBuffer on video first. This causes the test to end without running the video appendBuffer callback and we get one less assertion than we expect.

### DIFF
--- a/test/source-updater.test.js
+++ b/test/source-updater.test.js
@@ -810,7 +810,7 @@ function(assert) {
 
 QUnit.test('setDuration blocks audio and video queue entries until it finishes',
 function(assert) {
-  const done = assert.async();
+  const done = assert.async(2);
 
   assert.expect(6);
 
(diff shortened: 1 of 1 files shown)


test/source-updater.test.js
```diff
@@ -810,7 +810,7 @@ function(assert) {
 
 QUnit.test('setDuration blocks audio and video queue entries until it finishes',
 function(assert) {
-  const done = assert.async();
+  const done = assert.async(2);
```
brandonocasey: wait for two `done` calls, one from the video appendBuffer and one from the audio appendBuffer.

gkatsev: That's a thing? Neat.

PR #574, opened 2019-06-27
fix: remove captions within a range

brandonocasey:
## Description
Only remove cues in a specific time range.

### DIFF
--- a/src/util/text-tracks.js
+++ b/src/util/text-tracks.js
@@ -246,7 +246,7 @@ export const removeCuesFromTrack = function(start, end, track) {
     cue = track.cues[i];
 
     // Remove any overlapping cue
-    if (cue.startTime <= end && cue.endTime >= start) {
+    if (cue.startTime >= start && cue.endTime <= end) {
       track.removeCue(cue);
     }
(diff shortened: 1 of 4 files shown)


src/util/text-tracks.js
```diff
@@ -246,7 +246,7 @@ export const removeCuesFromTrack = function(start, end, track) {
     cue = track.cues[i];
 
     // Remove any overlapping cue
-    if (cue.startTime <= end && cue.endTime >= start) {
+    if (cue.startTime >= start && cue.endTime <= end) {
```
brandonocasey: So here is an example to explain this some more:
You have a 5 cues: 
1. `0 -> 2`
1. `2 -> 4`

We come into this function and we want to remove cues from `0 -> 3` Currently cues 1 and 2 will be removed because: 
1. `0 <= 3 && 2 >= 0` is true
2. `2 <= 3 && 3 >= 0` is true

Now only 1 will be removed:
1. `0 >= 0 && 2 <= 3` is true
 2. ` 2 >= 0 && 4 <= 3` is false

PR #575, opened 2019-06-27
experiment: a nasty ts/fmp4 playback hack for 1.x

brandonocasey:
## Description
A nasty hack to get ts/fmp4 playback working with 1.x before tba/lhls

### DIFF
--- a/src/mse/html-media-source.js
+++ b/src/mse/html-media-source.js
@@ -302,39 +302,34 @@ export default class HtmlMediaSource extends videojs.EventTarget {
 
     // Create a VirtualSourceBuffer to transmux MPEG-2 transport
     // stream segments into fragmented MP4s
-    if ((/^(video|audio)\/mp2t$/i).test(parsedType.type)) {
-      let codecs = [];
+    let codecs = [];
 
-      if (parsedType.parameters && parsedType.parameters.codecs) {
-        codecs = parsedType.parameters.codecs.split(',');
-        codecs = translateLegacyCodecs(codecs);
-        codecs = codecs.filter((codec) => {
-          return (isAudioCodec(codec) || isVideoCodec(codec));
-        });
-      }
+    if (parsedType.parameters && parsedType.parameters.codecs) {
+      codecs = parsedType.parameters.codecs.split(',');
+      codecs = translateLegacyCodecs(codecs);
+      codecs = codecs.filter((codec) => {
+        return (isAudioCodec(codec) || isVideoCodec(codec));
+      });
+    }
 
-      if (codecs.length === 0) {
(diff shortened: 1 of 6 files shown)


src/mse/html-media-source.js
```diff
@@ -302,39 +302,34 @@ export default class HtmlMediaSource extends videojs.EventTarget {
...
-      let codecs = [];
-
-      if (parsedType.parameters && parsedType.parameters.codecs) {
-        codecs = parsedType.parameters.codecs.split(',');
-        codecs = translateLegacyCodecs(codecs);
-        codecs = codecs.filter((codec) => {
-          return (isAudioCodec(codec) || isVideoCodec(codec));
-        });
-      }
+    let codecs = [];
```
brandonocasey: Always create virtual source buffers, as we may need to append at any time

src/mse/virtual-source-buffer.js
```diff
@@ -72,7 +72,7 @@ export default class VirtualSourceBuffer extends videojs.EventTarget {
     this.safeAppend_ = videojs.browser.IE_VERSION >= 11;
 
     let options = {
-      remux: false,
+      remux: true,
```
brandonocasey: Hard coded to true for the example.

src/segment-loader.js
```diff
@@ -1325,6 +1326,11 @@ export default class SegmentLoader extends videojs.EventTarget {
     }
 
     this.logger_(segmentInfoString(segmentInfo));
+    segmentInfo.bytes.transmux = segment.uri === 'media_00001.ts';
```
brandonocasey: This would just use isFmp4 detection from the tba branch, but hard coding is easier to show an example.

src/segment-loader.js
```diff
@@ -1304,17 +1304,18 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-          this.activeInitSegmentId_ = initId;
-        });
+        initSegment = this.initSegment(segment.map).bytes;
+        this.activeInitSegmentId_ = initId;
+      } else {
+        initSegment = new Uint8Array();
       }
+
+      segmentInfo.bytes.data = segmentInfo.bytes;
+      segmentInfo.bytes.initSegment = {data: initSegment, byteOffset: initSegment.byteOffset, byteLength: initSegment.byteLength};
```
brandonocasey: mapping things over so that the VirtualSourceBuffer can use it.

src/segment-loader.js
brandonocasey: This is basically the hack here, as adding ` transmux` property to the bytes passed to appendBuffer is really strange.

gkatsev:
I think we've all agreed we shouldn't do this. Thanks for putting this experiment together.

PR #577, opened 2019-06-27
Chore: vjsstandard 8

brandonocasey:
Update the code to videojs-standard 8 

**EDIT: only the  "manual fixes" commit is manual fixes, everything else was automatic.**


scripts/index.js
```diff
@@ -111,6 +109,11 @@
     onload();
   };
 
+
+  ['debug', 'autoplay', 'muted', 'minified', 'partial', 'url', 'type'].forEach(function(name) {
```
brandonocasey: vars need to be declared at the top

scripts/nomnoml.js
```diff
@@ -12,7 +13,6 @@ module.exports = {
 
     while (files.length > 0) {
       const file = path.resolve(nomnomlSourceDir, files.shift());
-      const extname = path.extname(file);
```
brandonocasey: unused

PR #580, opened 2019-06-28
chore: migrate to generator-videojs-plugin 7.6.3

brandonocasey:
## Description
Migrate to the newest version of `generator-videojs-plugin` so that this project is up to date with current conventions.

## Things of note
* Builds scripts were changed to be actual scripts, rather than needing to be required
* Moved test data creation to `test/dist` rather than just `test`

## TODO (in another pull request)
* Move test data creation into a rollup plugin
* Move `utils/manifest` to `test/manifest` as its weird to have test assets in `utils`
* Remove custom-assertions, as they modify qunit assert, which is really strange. Just make them helpers.
* Get tests running on ie11 and edge
* Update sinon


test/playback.test.js
```diff
@@ -237,7 +237,7 @@ QUnit.test('DASH sidx', function(assert) {
 
     // Firefox sometimes won't loop if seeking directly to the duration, or to too close
     // to the duration (e.g., 10ms from duration). 100ms seems to work.
-    player.currentTime(player.duration() - 0.1);
+    player.currentTime(player.duration() - 0.5);
```
gkatsev: are we having problems with 0.1?

scripts/create-test-segments.js
```diff
@@ -72,15 +83,7 @@ module.exports = {
       console.log('files in segments dir were changed rebuilding segments data');
       this.build();
     });
-  },
-
-  clean() {
```
gkatsev: why was clean removed? Should --clean still be an option above?

package.json
```diff
@@ -53,71 +47,73 @@
...
   },
   "files": [
-    "dist/"
+    "CONTRIBUTING.md",
+    "dist/",
+    "docs/",
+    "index.html",
+    "scripts/",
+    "src/",
+    "test/"
```
gkatsev: We have a lot of fixtures in test/, probably not needed for published release

package.json
```diff
@@ -12,37 +12,31 @@
...
-    "clean:dist": "rimraf dist dist-test",
-    "clean:test": "node scripts/clean-tests.js",
-    "postclean": "mkdirp dist",
-    "docs": "run-p docs:*",
-    "docs:api": "jsdoc src -r -d docs/api",
-    "docs:toc": "doctoc README.md",
-    "docs:nomnoml:build": "node -e \"var b=require('./scripts/nomnoml.js'); b.build();\"",
+    "clean": "shx rm -rf ./dist ./test/dist",
+    "postclean": "shx mkdir -p ./dist ./test/dist",
+    "build:test-manifests": "node ./scripts/create-test-manifests.js",
```
gkatsev: I don't think these two need to be run on `npm run build`, just on test should be enough

gkatsev: I would also prefer not to build the test dist file on just regular build but that sounds like it should happen in a separate PR

test/playback.test.js
brandonocasey: I was seing failures in firefox with this number every few runs, due to the increase in playbackRate.

scripts/create-test-segments.js
brandonocasey: These are now created in test/dist, which we cleanup

package.json
brandonocasey: Well since we build test assets on build we have to do them on build for now.

scripts/create-test-segments.js
gkatsev: So, we don't need lines 37-40?

PR #590, opened 2019-07-03
Added attribute NAME from playlist to Representation object

yurguis:
## Description
Working with a playlist like this one
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=6758400,NAME="Source"
hls_000s_test_source/index.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=3379200,NAME="High"
hls_000s_test_high/index.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=732160,NAME="Medium"
hls_000s_test_mid/index.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=337920,NAME="Low"
hls_000s_test_low/index.m3u8
```
I couldn't create a list for quality selection based on NAME attribute because the Representation object does not use this attribute.

## Specific Changes proposed
Add this attribute to Representation object.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "ignore": [
       "dist",
       "docs",
+      "deploy",
       "test/dist",
       "utils",
       "src/*.worker.js"
--- a/src/rendition-mixin.js
+++ b/src/rendition-mixin.js
@@ -64,6 +64,10 @@ class Representation {
       this.height = resolution.height;
     }
 
+    if (playlist.attributes.NAME) {
+      this.name = playlist.attributes.NAME;
+    }
+
     this.bandwidth = playlist.attributes.BANDWIDTH;
 
     // The id is simply the ordinality of the media playlist
--- a/src/rendition-mixin.js
+++ b/src/rendition-mixin.js
@@ -1,5 +1,18 @@
 import { isIncompatible, isEnabled } from './playlist.js';
 
+// See https://tools.ietf.org/html/rfc8216#section-4.3.4.2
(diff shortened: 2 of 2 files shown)


mister-ben:
No `NAME` attribute is defined in https://tools.ietf.org/html/rfc8216#section-4.3.4.2. Adding the attributes object to the representation might be a better option than adding special treatment of a non-standard attribute.

yurguis:
Yes, you are right, there is no `NAME` attribute defined in RFC8216, however, lately I have seen a lot of playlists using it and this was a quick fix for my project's need but in the long run it would be better to add an attributes object just like hls.js does in its m3u8 parser.

yurguis:
@mister-ben I have updated the PR adding all optional attributes to Rendition object, before we were just adding RESOLUTION. Also every not defined attribute in RFC8216 in the playlist will be added as a property to an `extras` property from the Rendition object. I don't know if this is exactly what you meant in your comment but this is cleaner than the first change I did.

yurguis:
I could achieve what I wanted by using `player.tech.hls.playlists.master.playlist` and then using `disabled = true` to force an specific quality level. This works pretty much like the Representation object, which makes me wonder why the need of something extra if we already have what we need?
There is no need for this PR anymore.

PR #593, opened 2019-07-08
chore: Create all test data using rollup

brandonocasey:
Create test data via a rollup plugin, rather than a separate script. This means:

* less npm scripts
* The data is tied to rollup directly, rather than forcing scripts to run before or with rollup.
* We can use rollup's watch!


test/dash-playlist-loader.test.js
```diff
@@ -16,8 +16,8 @@ import {
 } from './test-helpers';
 // needed for plugin registration
 import '../src/videojs-http-streaming';
-import testDataManifests from './dist/test-manifests.js';
-import { sidx as sidxResponse } from './dist/test-segments';
+import testDataManifests from 'create-test-data!manifests';
```
brandonocasey: anything that uses manifests or segments is now `create-test-data!segments/manifests` so that it can interface with the plugin.

scripts/create-test-data.js
```diff
@@ -0,0 +1,135 @@
...
+    "import window from 'global/window';\n" +
+    'const cache = {};\n' +
+    `const base64ToUint8Array = ${base64ToUint8Array.toString()};\n` +
+    `const utf16CharCodesToString = ${utf16CharCodesToString.toString()};\n` +
+    segmentDataExportStrings.join('\n');
+
+  return segmentsFile;
+};
+
+/* we refer to them as .js, so that babel and other plugins can work on them */
```
brandonocasey: Everything above this is copy pasted code.

scripts/create-test-data.js
```diff
@@ -0,0 +1,136 @@
...
+    const file = path.resolve(segmentsDir, files.shift());
+    const extname = path.extname(file);
+
+    if (extname === '.ts' || extname === '.mp4' || extname === '.key') {
+      // read the file directly as a buffer before converting to base64
+      const base64Segment = fs.readFileSync(file).toString('base64');
+
+      segmentData[path.basename(file, extname)] = base64Segment;
+    } else {
+      this.warn(`ignoring non-segment file ${path.relative(baseDir, file)} for segment test data`);
```
brandonocasey: This was changed to use rollups warning log.

scripts/create-test-data.js
```diff
@@ -0,0 +1,136 @@
...
+      manifests += fs.readFileSync(file, 'utf8')
+        .split(/\r\n|\n/)
+      // quote and concatenate
+        .map((line) => '    \'' + line + '\\n\' +\n')
+        .join('')
+      // strip leading spaces and the trailing '+'
+        .slice(4, -3);
+      manifests += ',\n';
+    } else {
+      this.warn(`ignoring non-manifest file ${path.relative(baseDir, file)} for manifest test data`);
```
brandonocasey: This was changed to use rollups warning log.

scripts/create-test-data.js
```diff
@@ -0,0 +1,136 @@
...
+    return acc;
+  }, []);
+
+  const segmentsFile =
+    'const cache = {};\n' +
+    `const base64ToUint8Array = ${base64ToUint8Array.toString()};\n` +
+    `const utf16CharCodesToString = ${utf16CharCodesToString.toString()};\n` +
+    segmentDataExportStrings.join('\n');
+
+  return segmentsFile;
```
brandonocasey: previously we wrote the file, now we just return.

scripts/create-test-data.js
```diff
@@ -0,0 +1,136 @@
...
+    } else {
+      this.warn(`ignoring non-manifest file ${path.relative(baseDir, file)} for manifest test data`);
+    }
+  }
+
+  // clean up and close the objects
+  manifests = manifests.slice(0, -2);
+  manifests += '\n};\n';
+
+  return manifests;
```
brandonocasey: previously we wrote the file, now we just return.

brandonocasey:
Most recent commit can be moved to it's own pull request if we want. Basically it seems like ie 11 was having issues with `localStorage` causing the tests to fail. I just wrapped a few more things in `try catch` and made the tests conditional to localStorage being available.

scripts/create-test-data.js
```diff
@@ -0,0 +1,136 @@
...
+    segmentDataExportStrings.join('\n');
+
+  return segmentsFile;
+};
+
+/* we refer to them as .js, so that babel and other plugins can work on them */
+const segmentsKey = 'create-test-data!segments.js';
+const manifestsKey = 'create-test-data!manifests.js';
+
+module.exports = function() {
```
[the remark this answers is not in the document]
gkatsev: but probably not worth doing as part of this PR.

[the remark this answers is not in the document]
brandonocasey: We can but then it makes it harder to use outside of rollup when/if needed. AKA we would have to run with babel-node or only use rollup to get things working.

PR #596, opened 2019-07-10
chore: update stats util

gkatsev:
- Make quality levels look a bit nicer
- Add separate audio and video buffer output

### DIFF
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -269,13 +269,9 @@ const transmuxAndNotify = ({
       });
       trackInfoFn = null;
 
-      if (probeResult.hasAudio) {
-        audioStartFn(probeResult.audioStart);
-      }
-      if (probeResult.hasVideo) {
-        videoStartFn(probeResult.videoStart);
-      }
+      audioStartFn(probeResult.audioStart);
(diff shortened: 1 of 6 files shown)


src/media-segment-request.js
```diff
@@ -269,9 +269,13 @@ const transmuxAndNotify = ({
       });
       trackInfoFn = null;
 
-      audioStartFn(probeResult.audioStart);
+      if (probeResult.hasAudio) {
```
brandonocasey: This seems more like a fix/different pull request

gkatsev:
@brandonocasey yeah, eventually, I'd like to combine the stats page and index into one thing but probably should be done separately.

src/media-segment-request.js
gkatsev: yeah, this and another change I have can/should go in another PR, I'll make sure to remove it when I'm fully done with the changes I want to make for this.

brandonocasey:
I just meant link to it in the `index.html` at the root of this project. I removed it as part of another pull request.

gkatsev:
Ah, I see. I think it's fine to leave it out for now since I'd like to combine them.

PR #597, opened 2019-07-11
fix: check loader type properly

gkatsev:
In the segment loader, we checked for `this.loaderType` but should be checking against `this.loaderType_`.
In addition, this passes the segmentInfo to the appendBuffer action, so we can have a more meaningful debug log.
Also, don't log audio buffer or video buffer data if they don't exist.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -659,6 +659,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     };
     this.resetLoader();
     this.remove(0, this.duration_(), done);
+
     // clears fmp4 captions
     if (this.captionParser_) {
       this.captionParser_.clearAllCaptions();
@@ -848,7 +849,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       return false;
     }
 
-    if (this.loaderType === 'main' &&
+    if (this.loaderType_ === 'main' &&
         segmentInfo.startOfSegment < this.sourceUpdater_.videoTimestampOffset()) {
       return true;
     }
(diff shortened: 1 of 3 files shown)


src/segment-loader.js
```diff
@@ -1557,7 +1558,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     const videoSegmentTimingInfoCallback =
       this.handleVideoSegmentTimingInfo_.bind(this, segmentInfo.requestId);
 
-    this.sourceUpdater_.appendBuffer({type, bytes, videoSegmentTimingInfoCallback}, (error) => {
+    this.sourceUpdater_.appendBuffer({segmentInfo, type, bytes, videoSegmentTimingInfoCallback}, (error) => {
```
brandonocasey: I was actually thinking that we may want to remove logging from `appendBuffer` and just log here instead, I guess this works too.

gkatsev: I think it's useful logging directly in the action because then you'll know exactly when it happens instead of just that it was queued.

ldayananda: you might want to pass just the important bits of segmentInfo here as it can be quite large. Otherwise LGTM

gkatsev: Theoretically, it shouldn't be an issue and could be useful when debugging. I think for this PR, I'd prefer to leave as-is, but we do have memory/performance related work that we're going to do.

src/segment-loader.js
```diff
@@ -848,7 +849,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       return false;
     }
 
-    if (this.loaderType === 'main' &&
+    if (this.loaderType_ === 'main' &&
```
gkatsev: btw, this is the main fix for this PR. Everything else just helps with debugging.

PR #598, opened 2019-07-11
fix: discos can happen on other loaders too

gkatsev:


gkatsev:
The proper fix is in #601.

PR #601, opened 2019-07-16
fix: Correctly use videoSegmentTimingInfo from the transmuxer

brandonocasey:
## Description
This must have happened sometime during the merging of tba/lhls and https://github.com/videojs/http-streaming/pull/371

Seems like we were expecting the native `SourceBuffer` to fire `videoSegmentTimingInfo`  which won't happen because we were firing it in `VirtualSourceBuffer` when the transmuxer fired it. Now we simply pass callbacks to `mediaSegmentRequest` for functions that we want to handle from the transmuxer.


src/source-updater.js
```diff
@@ -272,38 +272,18 @@ export default class SourceUpdater extends videojs.EventTarget {
    * @param {Function} done the function to call when done
    * @see http://www.w3.org/TR/media-source/#widl-SourceBuffer-appendBuffer-void-ArrayBuffer-data
    */
-  appendBuffer({segmentInfo, type, bytes, videoSegmentTimingInfoCallback}, doneFn) {
+  appendBuffer(type, bytes, doneFn) {
```
brandonocasey: I changed this from `(type, bytes, doneFn)` -> `({type, bytes, videoSegmentTimingInfoCallback}, doneFn)` when we needed to pass `videoSegmentTimingInfoCallback`. Now that we no longer need to, I put it back to this.

gkatsev: Any reason why not to keep the options object method signature? I would like to see `segmentInfo` continue to be pass through here to the appendBuffer action.

test/source-updater.test.js
```diff
@@ -6,7 +6,7 @@ import SourceUpdater from '../src/source-updater';
 import {mp4VideoInit, mp4AudioInit, mp4Video, mp4Audio} from './dist/test-segments';
 import { timeRangesEqual } from './custom-assertions.js';
 
-const checkIntialDuration = function({duration}) {
+const checkInitialDuration = function({duration}) {
```
brandonocasey: spelling

src/source-updater.js
brandonocasey: Yeah I will put it back to how it was

src/source-updater.js
```diff
@@ -272,38 +272,18 @@ export default class SourceUpdater extends videojs.EventTarget {
...
-        if (this.videoBuffer) {
-          this.videoBuffer.removeEventListener('videoSegmentTimingInfo', videoSegmentTimingInfoCallback);
-        }
-        originalDoneFn(err);
-      };
-    }
-
     if (type === 'audio' && this.videoBuffer && !this.videoAppendQueued_) {
-      this.delayedAudioAppendQueue_.push([{type, bytes, videoSegmentTimingInfoCallback}, doneFn]);
+      this.delayedAudioAppendQueue_.push([{type, bytes}, doneFn]);
```
gkatsev: I guess we should pass `segmentInfo` along here too?

PR #603, opened 2019-07-17
fix: Load type from url correctly

brandonocasey:
## Description
Since `type` is text and not url it was being treated like a text box and not loaded correctly from the url. This fixes that.

### DIFF
--- a/scripts/index-demo-page.js
+++ b/scripts/index-demo-page.js
@@ -118,7 +118,7 @@
     Object.keys(state).forEach(function(elName) {
       var el = stateEls[elName];
 
-      if (el.type === 'url') {
+      if (el.type === 'url' || el.type === 'text') {
         el.value = state[elName];
       } else {
         el.checked = state[elName] === 'true' ? true : false;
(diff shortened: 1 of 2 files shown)


gkatsev:
Should be chore, since not directly in VHS. Also, while you're here, can you fix the script/index.js to script/index-demo-page in the netlify build script?

scripts/index-demo-page.js
```diff
@@ -8,14 +8,23 @@
 
   var getInputValue = function(el) {
     if (el.type === 'url' || el.type === 'text') {
-      return el.value;
+      return encodeURIComponent(el.value);
```
brandonocasey: types with a `+` such as dash would save correctly but on page reload they wouldn't work. we need to encode and decode uri components.

scripts/index-demo-page.js
```diff
@@ -8,14 +8,23 @@
...
     } else if (el.type === 'checkbox') {
       return el.checked;
     }
 
     console.warn('unhandled input type ' + el.type);
   };
 
+  var setInputValue = function(el, value) {
```
brandonocasey: Moved this to a function so that set and get are right next to one another, should make it harder to forget.

PR #607, opened 2019-07-18
fix: Issues with video stutter at the start of playback.

brandonocasey:
## Description
Using the streams from: https://github.com/videojs/video.js/issues/5987

If audio is appended any time before video then the stutter in video occurs. So we prevent the audioSegmentLoader from loading until __after__ the main segment loader has appended.

> Note: We are leaking a ton of listeners in here, but I wanted to fix the bug by itself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "docs:api": "jsdoc src -r -d docs/api",
     "docs:toc": "doctoc README.md",
     "docs:nomnoml:build": "node -e \"var b=require('./scripts/nomnoml.js'); b.build();\"",
-    "lint": "vjsstandard",
+    "lint": "vjsstandard | grep -v \"maximum\\|todo\\|loop\"",
     "prestart": "npm run build",
     "start": "run-p start:server watch",
     "start:server": "node scripts/server.js",
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -39,6 +39,7 @@ const sumLoaderStat = function(stat) {
   return this.audioSegmentLoader_[stat] +
          this.mainSegmentLoader_[stat];
 };
+const isBuffered = (buffers, time) => buffers.some((buffer) => buffer && Ranges.findRange(buffer, time).length !== 0);
 
 /**
  * the master playlist controller controller all interactons
@@ -852,42 +853,36 @@ export class MasterPlaylistController extends videojs.EventTarget {
    * @return {TimeRange} the current time
    */
   setCurrentTime(currentTime) {
-    let buffered = Ranges.findRange(this.tech_.buffered(), currentTime);
+    const media = this.masterPlaylistLoader_ && this.masterPlaylistLoader_.media();
+    const segments = media && media.segments || [];
 
-    if (!(this.masterPlaylistLoader_ && this.masterPlaylistLoader_.media())) {
-      // return immediately if the metadata is not ready yet
-      return 0;
-    }
-
-    // it's clearly an edge-case but don't thrown an error if asked to
-    // seek within an empty playlist
-    if (!this.masterPlaylistLoader_.media().segments) {
-      return 0;
-    }
-
-    // In flash playback, the segment loaders should be reset on every seek, even
-    // in buffer seeks. If the seek location is already buffered, continue buffering as
-    // usual
-    // TODO: redo this comment
-    if (buffered && buffered.length) {
+    // Don't start segment loaders at the new location if:
+    // * master media is not ready
+    // * master media has no segments
+    if (!media || !segments.length) {
       return currentTime;
     }
 
-    // cancel outstanding requests so we begin buffering at the new
-    // location
-    this.mainSegmentLoader_.resetEverything();
-    this.mainSegmentLoader_.abort();
-    if (this.mediaTypes_.AUDIO.activePlaylistLoader) {
-      this.audioSegmentLoader_.resetEverything();
-      this.audioSegmentLoader_.abort();
-    }
-    if (this.mediaTypes_.SUBTITLES.activePlaylistLoader) {
-      this.subtitleSegmentLoader_.resetEverything();
-      this.subtitleSegmentLoader_.abort();
-    }
+    const {AUDIO, SUBTITLES} = this.mediaTypes_;
+    const ctSourceBuffered = isBuffered(this.mediaSource.sourceBuffers.map((s) => s.buffered), currentTime);
+    const ctSubtitleBuffered = isBuffered([this.subtitleSegmentLoader_.buffered_()], currentTime);
 
-    // start segment loader loading in case they are paused
-    this.load();
+
+    const loadersToReset = {
+      main: ctSourceBuffered,
+      audio: Boolean(AUDIO.activePlaylistLoader) && ctSourceBuffered,
+      subtitle: Boolean(SUBTITLES.activePlaylistLoader) && ctSubtitleBuffered,
+    };
+
+    Object.keys(loadersToReset).forEach((k) => {
+      const loader = this[`${k}SegmentLoader_`];
+
+      loader.resetEverything();
+      loader.abort();
+      loader.load();
+    });
+
+    return currentTime;
   }
 
   /**
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -39,7 +39,6 @@ const sumLoaderStat = function(stat) {
   return this.audioSegmentLoader_[stat] +
          this.mainSegmentLoader_[stat];
 };
-const isBuffered = (buffers, time) => buffers.some((buffer) => buffer && Ranges.findRange(buffer, time).length !== 0);
 
 /**
  * the master playlist controller controller all interactons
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "docs:api": "jsdoc src -r -d docs/api",
     "docs:toc": "doctoc README.md",
     "docs:nomnoml:build": "node -e \"var b=require('./scripts/nomnoml.js'); b.build();\"",
-    "lint": "vjsstandard",
+    "lint": "vjsstandard | grep -v \"maximum\\|todo\\|loop\"",
```
brandonocasey: because this linter is out of date and spits out a lot of warnings that are indistinguishable from errors.

src/media-groups.js
```diff
@@ -269,27 +269,24 @@ export const setupListeners = {
...
     });
 
-    playlistLoader.on('loadedplaylist', () => {
+    const loadPlaylist = () => {
+      playlistLoader.off(['loadedmetadata', 'loadedplaylist'], loadPlaylist);
       segmentLoader.playlist(playlistLoader.media(), requestOptions);
+    };
 
-      // If the player isn't paused, ensure that the segment loader is running
-      if (!tech.paused()) {
```
ldayananda: this code needs to stick around. I think there's a DASH flow that needs it

src/media-groups.js
```diff
@@ -269,27 +269,24 @@ export const setupListeners = {
       segmentLoaders: { [type]: segmentLoader }
     } = settings;
 
-    playlistLoader.on('loadedmetadata', () => {
+    // only start to append segments to source buffers after
+    // the main segment loader has already appended
+    settings.segmentLoaders.main.one('appended', () => {
       const media = playlistLoader.media();
 
-      segmentLoader.playlist(media, requestOptions);
```
ldayananda: this needs to happen before calling load on the segmentLoader... for the alt audio loader too. Particularly with DASH

brandonocasey: I would think `loadedplaylist` and `loadedmetadata` would have to happen before `appended` on the main segment loader? which means playlist will be set beforehand.

src/media-groups.js
brandonocasey: This code still exists in the `appended` listener

src/media-groups.js
```diff
@@ -269,27 +269,24 @@ export const setupListeners = {
...
       segmentLoader.playlist(playlistLoader.media(), requestOptions);
+    };
 
-      // If the player isn't paused, ensure that the segment loader is running
-      if (!tech.paused()) {
-        segmentLoader.load();
-      }
-    });
-
+    playlistLoader.one(['loadedmetadata', 'loadedplaylist'], loadPlaylist);
```
mjneil: this will break live

mjneil: and probably MAAT

src/media-groups.js
ldayananda: hopefully 😄  you might want to double check DASH sidx to see if this is always the case

src/media-groups.js
brandonocasey: as in it should be `on` instead of `one`?

mjneil: yes, the `playlistloader` will trigger `loadedplaylist` every time it refreshes the live manifest. If you stop listening to it after one event, the segment loader will never get new segments.

src/master-playlist-controller.js
```diff
@@ -852,42 +852,37 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    const segments = media && media.segments || [];
+    const buffered = this.tech_.buffered();
+    const ctBuffered = Ranges.findRange(this.tech_.buffered(), currentTime);
+
+    // Start segment loaders at the new location if:
+    // * master media is ready
+    // * master media has segments
+    // * It is not too early for tech to have a buffer
+    // * The seek requested is not already buffered
+    if (media && segments.length  && buffered.length && !ctBuffered.length) {
```
brandonocasey: The bug seemed to be that we `resetEverything` and `abort` right after before playback starts, but after things have started loading. To work around this we check that we check wether we have anything buffered at all, and if we don't then we do not reset or abort.

src/master-playlist-controller.js
```diff
@@ -852,42 +852,37 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-    if (this.mediaTypes_.SUBTITLES.activePlaylistLoader) {
-      this.subtitleSegmentLoader_.resetEverything();
-      this.subtitleSegmentLoader_.abort();
+      // start segment loader loading in case they are paused
+      this.load();
     }
 
-    // start segment loader loading in case they are paused
-    this.load();
+    return currentTime;
```
brandonocasey: When we reset previously we would not return a `currentTime` this is problematic because this function is used in middleware.

brandonocasey: Furthermore we always return the requested `currentTime` here instead of `0` sometimes because we don't actually want to change `currentTime` in our middleware we just want to reset/abort our loaders if current time is outside of a buffer.

src/master-playlist-controller.js
```diff
@@ -852,42 +852,37 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    const buffered = this.tech_.buffered();
+    const ctBuffered = Ranges.findRange(this.tech_.buffered(), currentTime);
+
+    // Start segment loaders at the new location if:
+    // * master media is ready
+    // * master media has segments
+    // * It is not too early for tech to have a buffer
+    // * The seek requested is not already buffered
+    if (media && segments.length  && buffered.length && !ctBuffered.length) {
+      console.log('reset');
```
ldayananda: mistake?

src/master-playlist-controller.js
ldayananda: don't we return time here? https://github.com/videojs/http-streaming/blob/1.x/src/middleware-set-current-time.js#L21

src/master-playlist-controller.js
ldayananda: should this be `media && segments.length  && !buffered.length && !ctBuffered.length)`?

brandonocasey: no because if we don't have any `tech` buffer yet then its likely that we haven't even started playback yet.

src/master-playlist-controller.js
brandonocasey: We do your right, I thought that middleware used the return value from this function, which it does not.

src/master-playlist-controller.js
brandonocasey: yeah

src/media-groups.js
brandonocasey: Refactored so that `appended`, `loadedplaylist` and `loadedmetadata` all listen with `loadPlaylist`

src/master-playlist-controller.js
ldayananda: it's possible, given the comment on previous line 868, that this change will break Flash playback

src/media-groups.js
```diff
@@ -266,30 +266,30 @@ export const setupListeners = {
...
+      segmentLoaders: { [type]: segmentLoader, main: mainSegmentLoader }
     } = settings;
 
-    playlistLoader.on('loadedmetadata', () => {
+
+    const loadPlaylist = () => {
       const media = playlistLoader.media();
 
-      segmentLoader.playlist(media, requestOptions);
+      segmentLoader.playlist(playlistLoader.media(), requestOptions);
```
mjneil: you can just use the `media` variable declared in the line above

src/master-playlist-controller.js
```diff
@@ -852,42 +853,36 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-      this.audioSegmentLoader_.resetEverything();
-      this.audioSegmentLoader_.abort();
-    }
-    if (this.mediaTypes_.SUBTITLES.activePlaylistLoader) {
-      this.subtitleSegmentLoader_.resetEverything();
-      this.subtitleSegmentLoader_.abort();
-    }
+      // if this loader does not have the requested time buffered
+      // reset it, abort current requests, and load at the new location.
+      if (!isBuffered(loader.buffered_(), currentTime)) {
```
brandonocasey: OK I think I figured out the only change we need now. Seems like the stutter always happens due to a `resetEverything` right after playback starts. This happens because the `buffered` value from tech doesn't tell you what the loader itself has buffered. It is a combination of all of the `buffered` values from our `SourceBuffers`. So when we try to seek to a `currentTime` of `0` if the `tech` buffered did not match all buffers were reset. Now we only reset buffers if the loaders `buffered` value does not contain currentTime.

gkatsev: So, for each loader type, figure out if we have the time we're seeking to buffered, otherwise, reset that loader?

brandonocasey: yeah

src/master-playlist-controller.js
```diff
@@ -852,42 +853,36 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    const media = this.masterPlaylistLoader_ && this.masterPlaylistLoader_.media();
+    const segments = media && media.segments || [];
 
-    if (!(this.masterPlaylistLoader_ && this.masterPlaylistLoader_.media())) {
-      // return immediately if the metadata is not ready yet
-      return 0;
+    // Don't start segment loaders at the new location if:
+    // * master media is not ready
+    // * master media has no segments
+    if (!media || !segments.length) {
```
gkatsev: This likely doesn't exit early if we have a masterPlaylistLoader but no media loaded yet. Potentially why the tests are failing.

src/middleware-set-current-time.js
```diff
@@ -19,17 +19,6 @@ videojs.use('*', (player) => {
       }
 
       return time;
-    },
-
-    // Sync VHS after play requests.
-    // This specifically handles replay where the order of actions is
-    // play, video element will seek to 0 (skipping the setCurrentTime middleware)
```
brandonocasey: We handle looping here: https://github.com/videojs/http-streaming/blob/master/src/videojs-http-streaming.js#L379

This play part of middleware is just causing issues.

gkatsev: This was for properly seeking back to 0 during a replay. We'd want to make sure that that works properly.

gkatsev: Looks like the relevant test is failing.

src/ranges.js
```diff
@@ -51,8 +51,8 @@ const filterRanges = function(timeRanges, predicate) {
  */
 export const findRange = function(buffered, time) {
   return filterRanges(buffered, function(start, end) {
-    return start - TIME_FUDGE_FACTOR <= time &&
-      end + TIME_FUDGE_FACTOR >= time;
+    return start - SAFE_TIME_DELTA <= time &&
```
brandonocasey: Since we were using `TIME_FUDGE_FACTOR` instead of `SAFE_TIME_DELTA` when comparing against current time badly aligned audio and video would cause us to `resetEverything`. Example:


currentTime = `0` 
video buffered = `[{start: 0.08, end: 6.17}]` 
audio buffered = `[{start: 0, end: 6.1}]` 


Due to how the code works the tech and loader buffered values will adopt the `video` buffered value. This means that with `TIME_FUDGE_FACTOR` 
`0.08 - (0.03333) <= 0` is false but with `SAFE_TIME_DELTA` this will be true: `0.08 - (0.1) <= 0`. 


I think that `SAFE_TIME_DELTA` is more appropriate here because of the comment for it, and seeing that it fixes this behavior. 

```
// Comparisons between time values such as current time and the end of the buffered range
// can be misleading because of precision differences or when the current media has poorly
// aligned audio and video, which can cause values to be slightly off from what you would
// expect. This value is what we consider to be safe to use in such comparisons to account
// for these scenarios.
```

brandonocasey: I also think it's possible that we should be using `SAFE_DELTA_TIME` in three other places, where we are comparing against `currentTime`. It might be worth looking into these in another pull request.

https://github.com/videojs/http-streaming/blob/master/src/segment-loader.js#L1144
https://github.com/videojs/http-streaming/blob/master/src/ranges.js#L70
https://github.com/videojs/http-streaming/blob/master/src/playback-watcher.js#L385

mjneil: Those other locations would require more investigation, but this specific change feels pretty safe to me

PR #610, opened 2019-07-22
fix: end the updating state on receiving nothing back from the transmuxer

misteroneill:
From @imbcmdth:

If the transmuxer never emits data but does signal that it is done with the current segment, we can finish handling this segment.

An example situation where this happens is when there is no audio or video content in a segment but there is timed metadata content. The transmuxer needs to be able to hold onto the metadata and wait for the next contentful segment. This change allows for that scenario.

## Requirements Checklist


gkatsev:
Easiest is probably a playback test.

gkatsev:
Yeah, I think the easiest way to get tests for us for this PR and https://github.com/videojs/mux.js/pull/279, is to just make a stream available with a cuepoint that we can test in the playback tests.

gkatsev:
I can confirm that with this change and videojs/mux.js#279 a stream with zero-length segments works.

PR #611, opened 2019-07-22
chore: update deps and package-lock

gkatsev:
This is particularly important to pick up a new version of fsevents, so we can use a newer version of node.


gkatsev:
I've tested the build output and it works in Chrome, Firefox, Safari, and IE11.

PR #615, opened 2019-07-26
feat: Add support zero length segments

brandonocasey:
## Description
#610 but for master


src/segment-loader.js
```diff
@@ -2040,6 +2040,9 @@ export default class SegmentLoader extends videojs.EventTarget {
     const prioritizedTimingInfo = useVideoTimingInfo && segmentInfo.videoTimingInfo ?
       segmentInfo.videoTimingInfo : segmentInfo.audioTimingInfo;
 
+    if (!prioritizedTimingInfo) {
```
brandonocasey: The only substantial change from the other pull request. We still get `timinginfo` even from segments without `timinginfo`. We just have to ignore it here.

gkatsev: I can confirm that the test fails without this and succeeds with this.

PR #616, opened 2019-07-26
fix: improve timestampOffset calculation to avoid gaps

alex-barstow:
## Description
Currently, the value for `timestampOffset` comes from `segment.end`, which is set from the DTS values returned by the tsprobe. This is not the correct value to use for the `timestampOffset`, and can result in missing frames or gaps between segments.

## Specific Changes proposed
The TBA/LHLS branch (now #master) improves upon the current calculation of the `timestampOffset` by using the buffered end value instead of `segment.end`. However, this can still result in gaps in some cases if there is a difference between the pts time and dts time of the first frame that exceeds an internal browser threshold (in Chrome this threshold seems to be right around .08). The way to address that is to factor in the difference between the pts start time and dts start time of the first frame: `timestampOffset = buffered.end - (ptsTime - dtsTime)`.

**Note:** This change will also need to be re-implemented in `master` on top of the TBA/LHLS changes at some point.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1284,6 +1284,17 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     if (segmentInfo.timestampOffset !== null &&
         segmentInfo.timestampOffset !== this.sourceUpdater_.timestampOffset()) {
+      const buffered = this.buffered_();
+
+      // Update the timestampOffset with more accurate timing info from the
+      // current buffered range and results of the probe, if it is available
+      if (buffered.length && timingInfo.segmentTimestampInfo) {
+        const ptsStartTime = timingInfo.segmentTimestampInfo[0].ptsTime;
+        const dtsStartTime = timingInfo.segmentTimestampInfo[0].dtsTime;
+
+        segmentInfo.timestampOffset = buffered.end(buffered.length - 1) - (ptsStartTime - dtsStartTime);
+      }
+
       this.sourceUpdater_.timestampOffset(segmentInfo.timestampOffset);
       // fired when a timestamp offset is set in HLS (can also identify discontinuities)
       this.trigger('timestampoffset');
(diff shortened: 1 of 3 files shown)


src/segment-loader.js
```diff
@@ -1284,6 +1284,17 @@ export default class SegmentLoader extends videojs.EventTarget {
...
         segmentInfo.timestampOffset !== this.sourceUpdater_.timestampOffset()) {
+      const buffered = this.buffered_();
+
+      // Update the timestampOffset with more accurate timing info from the
+      // current buffered range and results of the probe, if it is available
+      if (buffered.length && timingInfo && timingInfo.segmentTimestampInfo) {
+        const ptsStartTime = timingInfo.segmentTimestampInfo[0].ptsTime;
+        const dtsStartTime = timingInfo.segmentTimestampInfo[0].dtsTime;
+
+        segmentInfo.timestampOffset = buffered.end(buffered.length - 1) - (ptsStartTime - dtsStartTime);
```
mjneil: Instead of updating `timestampOffset` here and using `buffered.end` I think we should fix the source of the issue, i.e. `segment.end` and then just shift the `timestampOffset` by pts/dts difference here.

>Currently, the value for timestampOffset comes from segment.end, which is set from the DTS values returned by the tsprobe. This is not the correct value to use for the timestampOffset

I wouldn't say its not the correct value, we want to set the `timestampOffset` to the end of the previous segment so the next segment gets placed immediately after it, its just "incorrect" because `segment.end` is given the wrong value.

gkatsev: is it possible that we're not appending at the end of the timeline but somewhere in the middle?

alex-barstow:
Following offline discussion, we decided that fixing `segment.end` via the ts probe is not ideal since the probe is not used in #master. Instead, we opted to always set `startOfSegment` to the `buffered.end` rather than using `segment.end`.

gkatsev:
Tested with our various sources in index.html in master and seems to work fine.

PR #624, opened 2019-08-05
fix: Use pts value for timingInfo probes (#616/#660 from 1.x)

brandonocasey:
## Description
~~**I am not sure if this is correct or not, and will need feedback. This does fix the issue from #616 though**~~

#616 and #660 but for master. Since we already set `timestampOffset` to `bufferedEnd` in master all that was needed here was to change the `start` times that we probe for so that they use `ptsTime`. When trying to minus dtsTime and ptsTime as seen in #616 playback never works. After setting it to just use the `ptsTime` value everything works.  From my testing `ptsTime` works **a lot more consistently** then `dtsTime` 

Depends on https://github.com/videojs/mux.js/pull/300

### DIFF
--- a/test/segment-loader.test.js
+++ b/test/segment-loader.test.js
@@ -20,6 +20,7 @@ import {
 } from './loader-common.js';
 import {
   muxed as muxedSegment,
+  oneSecond as oneSecondSegment,
   audio as audioSegment,
   video as videoSegment,
   mp4Video as mp4VideoSegment,
@@ -423,7 +424,7 @@ QUnit.module('SegmentLoader', function(hooks) {
           this.currentTime = 210;
         });
 
-        standardXHRResponse(this.requests.shift(), muxedSegment());
+        standardXHRResponse(this.requests.shift(), oneSecondSegment());
         // the appended event will not fire, as segment-loader will realize that its guess
         // was off and will reset everything to load at the new point, therefore, wait for
         // the syncinfoupdate event rather than the appended event
@@ -630,7 +631,7 @@ QUnit.module('SegmentLoader', function(hooks) {
 
         assert.notOk(playlist.segments[0].end, 'does not start with duration');
 
-        standardXHRResponse(this.requests.shift(), muxedSegment());
+        standardXHRResponse(this.requests.shift(), oneSecondSegment());
 
         return new Promise((resolve, reject) => {
           loader.one('appended', resolve);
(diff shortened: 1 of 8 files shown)


test/segment-loader.test.js
```diff
@@ -423,7 +424,7 @@ QUnit.module('SegmentLoader', function(hooks) {
           this.currentTime = 210;
         });
 
-        standardXHRResponse(this.requests.shift(), muxedSegment());
+        standardXHRResponse(this.requests.shift(), oneSecondSegment());
```
brandonocasey: The probed values for `muxedSegment` are not legitimate it seems. When using  `oneSecond.ts`, a one second ts file that I generated everything probes as expected in the test.

test/segment-loader.test.js
```diff
@@ -1020,7 +1021,7 @@ QUnit.module('SegmentLoader', function(hooks) {
           loader.one('error', reject);
         });
       }).then(() => {
-        assert.equal(playlistUpdated.segments[0].start, 0, 'set start on segment of new playlist');
+        assert.equal(playlistUpdated.segments[0].start, 0.11072222222222239, 'set start on segment of new playlist');
```
brandonocasey: This is the probed pts for one second segment.

gkatsev:
I think in discussion, this was brought up and the concern was whether we'll create a regression in some version of Firefox. We'd probably want to test against Firefox ESR in addition to latest Firefox for this. However, I think this was suggested as a correct change.

brandonocasey:
good to know I'm not nuts, I will test it out in both Firefox's

brandonocasey:
Appears to work on Firefox and Firefox ESR.  Would be good to have someone else test when they get a chance.

gjanblaszczyk:
I have just tested this PR and I found one problem with seeking on Live HLS DVR. The player is randomly hanging after seek... You can check it on this HLS playlist: http://live-api.miraclechannel.com/hls/miracle1.m3u8?DVR

gkatsev:
Thanks, we'll definitely look at that. I think there's still a bit more work to do before we can merge this PR.

gkatsev:
Seeing extra buffering with the content from https://github.com/videojs/video.js/issues/6282 in this PR compared to VHS 1.11.2.

brandonocasey:
After a rebase this content appears to be working fine. Must have been solved in master.

PR #625, opened 2019-08-06
Add option to pass an object as the source

gesinger:
## Description

This feature adds support for passing a JSON object as the source via a data URI, which VHS looks for alongside a new vendor (VHS) specific media type, 'application/vnd.vhs+json'. If present, VHS will use the provided manifest object as the first requested manifest. See the README for more details.

## Specific Changes Proposed

Besides the addition of the manifest object source option and the media type application/vnd.vhs+json, modifications had to be made through the pipeline, from the top-level of the plugin at videojs-http-streaming.js through the *playlist-loader.js and media-groups.js modules. By the time the segment loaders are reached, the handling should be the same as if the playlist was loaded through the normal mechanism of a request to the src URL.

## Examples

```

player.src({
  src: `data:application/vnd.vhs+json,${JSON.stringify(master)}`,
  type: 'application/vnd.vhs+json'
});
```

```

});

player.src({
  src: `data:application/vnd.vhs+json,${JSON.stringify(master)}`,
  type: 'application/vnd.vhs+json'
});
```

## Extra Notes

If we want to break this PR up, I may be able to do some of reused function extractions as separate PRs.

## Requirements Checklist
  - [X] Docs/guides updated


gesinger:
Updated to use a data URI for the source instead of the extra `manifestObject` option. Thanks @gkatsev and @mjneil !

PR #629, opened 2019-08-08
feat: setup EME key systems for HLS as well as DASH 

alex-barstow:
## Description
Currently, VHS augments the EME key systems of DASH sources with DRM-related information from the manifest. In order to support Widevine HLS content, we will need to expand this key system handling to work for HLS content as well. Since the existing logic is based on the assumption that we are dealing with DASH, and thus unmuxed video and audio, some tweaks are needed in order to account for the fact that HLS can be either muxed or unmuxed.

## Specific Changes proposed
- get video and audio mimeTypes and codecs from `mainSegmentLoader.mimeType_` and `audioSegmentLoader.mimeType_` instead of playlist attributes, since we already do some of the [work needed](https://github.com/videojs/http-streaming/blob/1.x/src/master-playlist-controller.js#L1094) to format those strings for EME when we initially setup the source buffers. The one exception to that is muxed HLS, where some string manipulation is still required to parse out the individual codecs from a comma-delimited list.
- test changes
  - segment loader mocking needed to be added to account for changes to the `emeKeySystems()` function
  - added `openMediaSource()` call to one eme integration test so segment loader mimeTypes will be set
- created player for HLS Integration tests since this new code path for HLS requires a player to exist

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -141,43 +141,62 @@ Hls.canPlaySource = function() {
     'your player\'s techOrder.');
 };
 
-const emeKeySystems = (keySystemOptions, videoPlaylist, audioPlaylist) => {
+const emeKeySystems = (keySystemOptions, mainSegmentLoader, audioSegmentLoader) => {
   if (!keySystemOptions) {
     return keySystemOptions;
   }
 
-  const splitCodecs = videoPlaylist.attributes.CODECS.split(',');
-  let videoCodec;
-  let audioCodec;
+  let videoMimeAndCodec;
+  let audioMimeAndCodec;
 
-  // If there are comma delimited codecs, we are dealing with HLS
-  if (splitCodecs.length > 1) {
-    splitCodecs.forEach(codec => {
+  // if there is a mimeType associated with the audioSegmentLoader, then the audio
+  // and video mimeType and codec strings are already in the format we need to
+  // pass with the other key systems
+  if (audioSegmentLoader.mimeType_) {
+    videoMimeAndCodec = mainSegmentLoader.mimeType_;
+    audioMimeAndCodec = audioSegmentLoader.mimeType_;
+
+  // if there is no audioSegmentLoader mimeType, then we have to create the
+  // the audio and video mimeType/codec strings from information extrapolated
+  // from the mainSegmentLoader mimeType (ex. 'video/mp4; codecs="mp4, avc1"' -->
+  // 'video/mp4; codecs="avc1"' and 'audio/mp4; codecs="mp4"')
+  } else {
+    const splitMime = mainSegmentLoader.mimeType_.split(';');
+    const subtype = splitMime[0].split('/')[1];
+    const parameter = splitMime[1].trim();
+    const splitParameter = parameter.split('"')[1];
+    const codecs = splitParameter.split(',');
+    codecs[1] = codecs[1].trim();
+
+    let audioCodec;
+    let videoCodec;
+
+    codecs.forEach(codec => {
       if (isAudioCodec(codec)) {
         audioCodec = codec;
       } else if (isVideoCodec(codec)) {
         videoCodec = codec;
       }
     });
-  // If there aren't, we are dealing with DASH which specifies an individual
-  // codec per playlist
-  } else {
-      videoCodec = videoPlaylist.attributes.CODECS;
-      audioCodec = audioPlaylist && audioPlaylist.attributes.CODECS;
+
+    if (!videoCodec || !audioCodec) {
+      videojs.log.warn('Insufficient codec information in the manifest.' +
+        ' There may be problems playing encrypted content');
+    }
+
+    videoMimeAndCodec = `video/${subtype}; codecs="${videoCodec}"`;
+    audioMimeAndCodec = `audio/${subtype}; codecs="${audioCodec}"`;
   }
 
-  if (!videoCodec || !audioCodec) {
-    videojs.log.warn('Insufficient codec information in the manifest.' +
-      ' There may be problems playing encrypted content');
-  }
 
   // upsert the content types based on the selected playlist
   const keySystemContentTypes = {};
+  const videoPlaylist = mainSegmentLoader.playlist_;
 
   for (let keySystem in keySystemOptions) {
     keySystemContentTypes[keySystem] = {
-      audioContentType: `audio/mp4; codecs="${audioCodec}"`,
-      videoContentType: `video/mp4; codecs="${videoCodec}"`
+      audioContentType: audioMimeAndCodec,
+      videoContentType: videoMimeAndCodec
     };
 
     if (videoPlaylist.contentProtection &&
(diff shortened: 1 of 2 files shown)


src/videojs-http-streaming.js
```diff
@@ -140,18 +141,56 @@ Hls.canPlaySource = function() {
...
+    videoMimeAndCodec = mainSegmentLoader.mimeType_;
+    audioMimeAndCodec = audioSegmentLoader.mimeType_;
+
+  // if there is no audioSegmentLoader mimeType, then we have to create the
+  // the audio and video mimeType/codec strings from information extrapolated
+  // from the mainSegmentLoader mimeType (ex. 'video/mp4; codecs="mp4, avc1"' -->
+  // 'video/mp4; codecs="avc1"' and 'audio/mp4; codecs="mp4"')
+  } else {
+    const splitMime = mainSegmentLoader.mimeType_.split(';');
+    const subtype = splitMime[0].split('/')[1];
```
brandonocasey: Should trim this just in case

src/videojs-http-streaming.js
```diff
@@ -140,18 +141,56 @@ Hls.canPlaySource = function() {
...
+  if (audioSegmentLoader.mimeType_) {
+    videoMimeAndCodec = mainSegmentLoader.mimeType_;
+    audioMimeAndCodec = audioSegmentLoader.mimeType_;
+
+  // if there is no audioSegmentLoader mimeType, then we have to create the
+  // the audio and video mimeType/codec strings from information extrapolated
+  // from the mainSegmentLoader mimeType (ex. 'video/mp4; codecs="mp4, avc1"' -->
+  // 'video/mp4; codecs="avc1"' and 'audio/mp4; codecs="mp4"')
+  } else {
+    const splitMime = mainSegmentLoader.mimeType_.split(';');
```
brandonocasey: Would it make sense to put this whole block of code in `src/utils/codecs`? something like `parseMimeType` that returns something like: `{video: {type: 'mp4', codecs: ['mp4, 'avc1']`?

src/videojs-http-streaming.js
```diff
@@ -140,18 +141,56 @@ Hls.canPlaySource = function() {
     'your player\'s techOrder.');
 };
 
-const emeKeySystems = (keySystemOptions, videoPlaylist, audioPlaylist) => {
+const emeKeySystems = (keySystemOptions, mainSegmentLoader, audioSegmentLoader) => {
   if (!keySystemOptions) {
     return keySystemOptions;
   }
 
+  let videoMimeAndCodec;
```
brandonocasey: should these just be `videoMimeType` and `audioMimeType`? seems like that is they way that it is referred to in most of the code

src/videojs-http-streaming.js
alex-barstow: How about using `parseContentType()` from the codec utils which already does almost the same thing?

src/videojs-http-streaming.js
```diff
@@ -140,18 +141,52 @@ Hls.canPlaySource = function() {
...
+    videoMimeType = mainSegmentLoader.mimeType_;
+    audioMimeType = audioSegmentLoader.mimeType_;
+
+  // if there is no audioSegmentLoader mimeType, then we have to create the
+  // the audio and video mimeType/codec strings from information extrapolated
+  // from the mainSegmentLoader mimeType (ex. 'video/mp4; codecs="mp4, avc1"' -->
+  // 'video/mp4; codecs="avc1"' and 'audio/mp4; codecs="mp4"')
+  } else {
+    const parsedMimeType = parseContentType(mainSegmentLoader.mimeType_);
+    const codecs = parsedMimeType.parameters.codecs.split(',').map(codec => codec.trim());
```
brandonocasey: Not a big deal since we won't run this code that much, but it would probably be better to do the trim at the top of the `forEach` to prevent looping through the same list twice.

alex-barstow: Yup, good point

src/videojs-http-streaming.js
```diff
@@ -140,18 +141,52 @@ Hls.canPlaySource = function() {
...
   }
 
+  let videoMimeType;
+  let audioMimeType;
+
+  // if there is a mimeType associated with the audioSegmentLoader, then the audio
+  // and video mimeType and codec strings are already in the format we need to
+  // pass with the other key systems
+  if (audioSegmentLoader.mimeType_) {
+    videoMimeType = mainSegmentLoader.mimeType_;
```
gkatsev: are these equivalent to `videoPlaylist.attributes.CODECS` and same for audio?

src/videojs-http-streaming.js
```diff
@@ -172,16 +207,16 @@ const emeKeySystems = (keySystemOptions, videoPlaylist, audioPlaylist) => {
 };
 
 const setupEmeOptions = (hlsHandler) => {
-  if (hlsHandler.options_.sourceType !== 'dash') {
-    return;
-  }
+  const mainSegmentLoader = hlsHandler.masterPlaylistController_.mainSegmentLoader_;
+  const audioSegmentLoader = hlsHandler.masterPlaylistController_.audioSegmentLoader_;
```
gkatsev: Does this work for both muxed and umuxed content?

test/videojs-http-streaming.test.js
```diff
@@ -3109,7 +3109,7 @@ QUnit.skip('populates quality levels list when available', function(assert) {
   assert.ok(this.player.tech_.hls.qualityLevels_, 'added quality levels from video with source');
 });
 
-QUnit.test('configures eme if present on selectedinitialmedia', function(assert) {
+QUnit.test('configures eme for DASH if present on selectedinitialmedia', function(assert) {
```
gkatsev: the changes in this test is due to how we're setting up EME now? Is it possible to make it so that the test doesn't change?

test/videojs-http-streaming.test.js
```diff
@@ -3720,7 +3761,7 @@ QUnit.module('HLS Integration', {
     this.env = useFakeEnvironment(assert);
     this.requests = this.env.requests;
     this.mse = useFakeMediaSource();
-    this.tech = new (videojs.getTech('Html5'))({});
+    this.player = createPlayer();
```
gkatsev: if you create the player and then set `this.tech` afterwards, you won't need to change all the tests below.

src/videojs-http-streaming.js
alex-barstow: The first bullet under "Specific Changes" addresses this a bit. The values of the codecs themselves are the same, but `mimeType_` is already formatted as `'type/subtype; codecs="codec1, codec2"'` while `attributes.CODECS` is only `'codec1, codec2'`

gkatsev: ah, cool, forgot to refer back to that when re-reviewing.

src/videojs-http-streaming.js
alex-barstow: What do you mean by "this"?

test/videojs-http-streaming.test.js
alex-barstow: That is true, but I decided to follow the convention elsewhere in the tests of referencing the tech as a property of the player when a player exists. Happy to change it if you feel strongly about it.

src/videojs-http-streaming.js
gkatsev: The current changes really. Specifically, with muxed audio, we don't have a separate audio segment loader. We should make sure that this works in either case, or at least, doesn't break existing streams.

gkatsev: Oh, I guess it's unlikely to break existing streams because dash is always unmuxed.

test/videojs-http-streaming.test.js
alex-barstow: I think the mocking in this test does need to be updated since the old one mocked objects that are no longer used to setup eme

src/videojs-http-streaming.js
alex-barstow: If the content is muxed, then `audioSegmentLoader` is `null` and `emeKeySystems()` will pull the relevant audio/video codec info from the `mainSegmentLoader.mimeType_` and format it into separate audio/video mime types accordingly. If it is unmuxed, then the audio/video codec info has already separated out and formatted by [`configureLoaderMimeTypes_()`](https://github.com/videojs/http-streaming/blob/1.x/src/master-playlist-controller.js#L1127-L1130) into the exact format we need to setup eme.

test/videojs-http-streaming.test.js
gkatsev: this is fine, carry on 🙂

src/videojs-http-streaming.js
```diff
@@ -140,18 +141,54 @@ Hls.canPlaySource = function() {
...
+
+      if (isAudioCodec(codec)) {
+        audioCodec = codec;
+      } else if (isVideoCodec(codec)) {
+        videoCodec = codec;
+      }
+    });
+
+    videoMimeType = `${parsedMimeType.type}; codecs="${videoCodec}"`;
+    audioMimeType = `${parsedMimeType.type.replace('video', 'audio')}; codecs="${audioCodec}"`;
```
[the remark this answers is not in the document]
gkatsev: Yeah, we did just recently get a question about that. I think, for now, we can continue not supporting it, but definitely something we should fix eventually.

PR #632, opened 2019-08-09
perf: save 1157b gzip by using module mux.js

brandonocasey:
## Description
`74033` -> `72876` this includes the optimizations from #626 , so it is `1157b` more than that

Depends on https://github.com/videojs/mux.js/pull/290


brandonocasey:
Going to close this as we won't have time to get to it any time soon, and it can happen easily after we transition mux.js to es6 code style.

PR #635, opened 2019-08-16
test(zero-length): update segment and unskip test

gkatsev:
One thing I noticed is that this test fails in Safari, currently.


gkatsev:
Safari is failing because the test assumes that the metadata track will be enabled automatically and that VHS is adding its segment-metadata track first. For Safari, we'll either need to overrideNative or enable the metadata track manually. However, we don't officially support Safari nor do we test on it. So, I don't think it's a problem.

PR #637, opened 2019-08-19
Perf: Save 597 gz bytes through code sharing

brandonocasey:
## Description
`74033` -> `72945`

We inline 4-5 `@babel/runtime` helpers, 4 different `Stream` classes, `3`  base 64 to uin8array functions, and two resolve urls (which each includes their own version of url-toolkit.). This full request removes all the duplicate code and moves it to `@videojs/vhs-utils`. It also updates `videojs-generate-rollup-config` to deal with sharing `@babel/runtime` helpers.

Depends on the following generator and small code updates:
- [x] https://github.com/brightcove/pkcs7/pull/13
- [x] https://github.com/videojs/aes-decrypter/pull/68
- [x] https://github.com/videojs/mpd-parser/pull/76
- [x] https://github.com/videojs/m3u8-parser/pull/95
- [x] ~~https://github.com/videojs/mux.js/pull/295~~
- [x] https://github.com/videojs/vhs-utils released

Closes #640
Closes #641


src/resolve-url.js
```diff
@@ -2,22 +2,9 @@
  * @file resolve-url.js - Handling how URLs are resolved and manipulated
  */
 
-import URLToolkit from 'url-toolkit';
```
gkatsev: does URLToolkit need to be a direct dep of VHS anymore?

gkatsev:
I wonder if there could be any issues with the stream.js compat across the libs. They mostly look the same but there's some minor changes.

src/resolve-url.js
brandonocasey: it does not

brandonocasey:
stream.js only had one difference between aes-decrypter, m3u8-parser, and mux.js and it was actually a bug that was fixed in Mux.js. https://github.com/videojs/vhs-utils/blob/master/src/stream.js#L51

src/resolve-url.js
brandonocasey: It actually isn't anymore, just a dev dependency, as it's used in tests.

gkatsev: Ah, cool, missed that it was moved over.

gkatsev:
FYI, tests are failing. Also, we'd want to make sure that the deps get non-prerelease releases before we merge this.

test/playback.test.js
```diff
@@ -287,29 +287,33 @@ QUnit[testFn]('DASH sidx with alt audio should end', function(assert) {
   });
 });
 
-QUnit[testFn]('loops', function(assert) {
-  const done = assert.async();
-  const player = this.player;
+// TODO: why does this make the next test
+// throw an "The operation was aborted." on firefox
+if (!videojs.browser.IS_FIREFOX) {
```
brandonocasey: Firefox 69 started throwing an error "The Operation was aborted", and no amount of promise silencing seems to help. The test itself actually passes, but it throws an error that causes a global failure.

test/source-updater.test.js
```diff
@@ -810,48 +810,50 @@ QUnit.test('setDuration waits for video buffer to finish updating', function(ass
   assert.ok(this.sourceUpdater.updating(), 'updating during appends');
 });
 
-QUnit.test(
-  'setDuration waits for both audio and video buffers to finish updating',
-  function(assert) {
-    const done = assert.async();
-    let appendsFinished = 0;
+if (!videojs.browser.IS_EDGE) {
```
brandonocasey: This test is failing intermittently on edge, even in master, since edge throws on setDuration sometimes.

test/playback.test.js
brandonocasey: This happens in master too.

misteroneill: I wonder if this should instead use `QUnit.skip`? Maybe `QUnit[videojs.browser.IS_FIREFOX ? 'skip' : testFn]`?

test/source-updater.test.js
misteroneill: Same comment about skipping here.

PR #638, opened 2019-08-19
fix: id3 firing at an incorrect time due to mux.js issue

brandonocasey:
#635 and #634 but for master


gkatsev:
So, once we update mux it's all good in master? That's nice.

PR #649, opened 2019-08-30
feat: add option to pass an object as the source

gesinger:
Updated 1.x PR (https://github.com/videojs/http-streaming/pull/625) to master.

## Description

This feature adds support for passing a JSON object as the source via a data URI, which VHS looks for alongside a new vendor (VHS) specific media type, 'application/vnd.vhs+json'. If present, VHS will use the provided manifest object as the first requested manifest. See the README for more details.

## Specific Changes Proposed

Besides the addition of the manifest object source option and the media type application/vnd.vhs+json, modifications had to be made through the pipeline, from the top-level of the plugin at videojs-http-streaming.js through the *playlist-loader.js and media-groups.js modules. By the time the segment loaders are reached, the handling should be the same as if the playlist was loaded through the normal mechanism of a request to the src URL.

## Examples

```

player.src({
  src: `data:application/vnd.vhs+json,${JSON.stringify(master)}`,
  type: 'application/vnd.vhs+json'
});
```

```

});

player.src({
  src: `data:application/vnd.vhs+json,${JSON.stringify(master)}`,
  type: 'application/vnd.vhs+json'
});
```

## Extra Notes

If we want to break this PR up, I may be able to do some of reused function extractions as separate PRs.

## Requirements Checklist
  - [X] Docs/guides updated


src/videojs-http-streaming.js
```diff
@@ -95,6 +95,17 @@ const simpleTypeFromSourceType = (type) => {
...
   }
 
+  // Denotes the special case of a pre-parsed manifest object passed in instead of the
+  // traditional source URL.
+  //
+  // See https://en.wikipedia.org/wiki/Media_type for details on specifying media types.
+  //
+  // In this case, vnd is for vendor, VHS is for this project, and the +json suffix
+  // identifies the structure of the media type.
+  if (type === 'application/vnd.vhs+json') {
```
gkatsev: should this be `vnd.videojs.vhs`?

src/videojs-http-streaming.js
```diff
@@ -95,6 +95,17 @@ const simpleTypeFromSourceType = (type) => {
...
+    return 'vhs-json';
```
gkatsev: is it possible to know whether it is HLS or DASH so that later if you do `player.currentType()` you'll get back an HLS or DASH type?

gkatsev: I guess it doesn't have to be here specifically.

src/videojs-http-streaming.js
```diff
@@ -471,7 +502,8 @@ class HlsHandler extends Component {
     }
     this.setOptions_();
     // add master playlist controller options
-    this.options_.url = this.source_.src;
+    this.options_.src = (this.source_.src && (/^data:/i).test(this.source_.src)) ?
```
gkatsev: seems like this can be simplified to just `expandDataUri(this.source_.src)` since it returns the string argument if it isn't a data uri.

src/videojs-http-streaming.js
```diff
@@ -883,5 +915,6 @@ export {
   HlsHandler,
   HlsSourceHandler,
   emeKeySystems,
-  simpleTypeFromSourceType
+  simpleTypeFromSourceType,
+  expandDataUri
```
gkatsev: is this something that's needed to be exposed?

src/videojs-http-streaming.js
gesinger: Good call, yeah, will update.

src/videojs-http-streaming.js
gesinger: Technically the VHS JSON object is streaming format independent. For instance, we can concatenate DASH and HLS together into one VHS JSON object.

src/videojs-http-streaming.js
gesinger: Just for tests

gesinger:
Closed in favor of a newer branch (PR not yet opened). Comments will be addressed on that branch.

PR #656, opened 2019-09-11
fix: update timestamp and dont wait for append, for zero length segments

brandonocasey:
## Description
Currently master waits for zero length segments to append, and they never will. So playback stalls and the video is assumed to have ended once we reach a zero length segment. Once I fixed that I realized that we have to set timingInfo start to zero for a zero length segment as we won't get a `timinginfo` event from mux.js.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1948,6 +1948,20 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     segmentInfo.waitingOnAppends = 0;
 
+    // segments with no data
+    if (!segmentInfo.hasAppendedData_) {
+      // override settings for metadata only segments
+      segmentInfo.timestampOffset = segmentInfo.startOfSegment;
+      segmentInfo.timingInfo = {start: 0};
+      segmentInfo.waitingOnAppends++;
+
+      // update the timestampoffset
+      this.updateSourceBufferTimestampOffset_(segmentInfo);
+      // append is "done" instantly with no data.
+      this.checkAppendsDone_(segmentInfo);
+      return;
+    }
+
     // Since source updater could call back synchronously, do the increments first.
     if (waitForVideo) {
       segmentInfo.waitingOnAppends++;
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1957,6 +1957,11 @@ export default class SegmentLoader extends videojs.EventTarget {
 
       // update the timestampoffset
       this.updateSourceBufferTimestampOffset_(segmentInfo);
+
+      // make sure the metadata queue is processed even though we have
+      // no video/audio data.
+      this.processMetadataQueue_();
+
       // append is "done" instantly with no data.
       this.checkAppendsDone_(segmentInfo);
       return;


src/segment-loader.js
```diff
@@ -1948,6 +1948,20 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     segmentInfo.waitingOnAppends = 0;
 
+    // segments with no data
+    if (!segmentInfo.hasAppendedData_) {
+      // override settings for metadata only segments
+      segmentInfo.timestampOffset = segmentInfo.startOfSegment;
+      segmentInfo.timingInfo = {start: 0};
+      segmentInfo.waitingOnAppends++;
```
brandonocasey: We have to mock that we are waiting on an append, because in `checkAppendsDone` we do `segmentInfo. waitingOnAppends--` and then check against `0`.

src/segment-loader.js
```diff
@@ -1948,6 +1948,20 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     segmentInfo.waitingOnAppends = 0;
 
+    // segments with no data
+    if (!segmentInfo.hasAppendedData_) {
+      // override settings for metadata only segments
+      segmentInfo.timestampOffset = segmentInfo.startOfSegment;
+      segmentInfo.timingInfo = {start: 0};
```
brandonocasey: These segments start at 0

src/segment-loader.js
```diff
@@ -1948,6 +1948,20 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     segmentInfo.waitingOnAppends = 0;
 
+    // segments with no data
+    if (!segmentInfo.hasAppendedData_) {
+      // override settings for metadata only segments
+      segmentInfo.timestampOffset = segmentInfo.startOfSegment;
```
brandonocasey: For zero length segments we have to use the start of segment and not buffered end.

DispatchCommit:
Is this intended to resolve issues where a live HLS stream may drop connection for a brief period before regaining connection, but the player will stop playback until enough data has been loaded to jump past the time where the stream dropped?
I've been struggling to find the root cause of that bug for a long time now.

gkatsev:
@DispatchCommit no, this unrelated to what you describe.

gkatsev:
We should review this and also make sure that the test from 1.x for this is available here and is failing before this change and passing with this change.

gkatsev:
Looks like the manifest and segment for the zero-length segments are in master. Wonder why it's passing even without these changes.

src/segment-loader.js
gesinger: Instead of that, can this function just call `handleAppendsDone_` directly?

src/segment-loader.js
```diff
@@ -1948,6 +1948,20 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      segmentInfo.timingInfo = {start: 0};
+      segmentInfo.waitingOnAppends++;
+
+      // update the timestampoffset
+      this.updateSourceBufferTimestampOffset_(segmentInfo);
```
gesinger: Since there's no `handleData_` call (there being no data), should we call to `processMetadataQueue_ ` here for segments that have no video/audio data but have id3 tags/captions?

src/segment-loader.js
brandonocasey: We probably could but I was worried that we wouldn't get the abort handling that exists in `checkAppendsDone_`

src/segment-loader.js
brandonocasey: I think that we should yeah.

PR #657, opened 2019-09-11
feat: setup EME key systems for HLS as well as DASH (#629)

brandonocasey:
## Description
#629 but for master!


src/source-updater.js
```diff
@@ -199,6 +199,7 @@ export default class SourceUpdater extends videojs.EventTarget {
     // initial timestamp offset is 0
     this.audioTimestampOffset_ = 0;
     this.videoTimestampOffset_ = 0;
+    this.codecs = {};
```
brandonocasey: Store the codecs used to setup the source buffers

src/util/codecs.js
```diff
@@ -186,3 +186,25 @@ export const codecsForPlaylist = function(master, media) {
...
 };
+
+/*
+ * Check if a codec string refers to an audio codec.
+ *
+ * @param {String} codec codec string to check
+ * @return {Boolean} if this is an audio codec
+ * @private
+ */
+export const isAudioCodec = function(codec) {
```
brandonocasey: brought these back from 1.x

brandonocasey:
You're right, `selectedintialmedia` won't really work. I just wonder if `sourceUpdater` `ready` will be too late.

src/util/codecs.js
gkatsev: not as part of this PR, but may make sense to move this into vhs-utils.

test/manifests/demuxed-two.m3u8
```diff
@@ -0,0 +1,6 @@
+#EXTM3U
+#EXT-X-VERSION:4
+#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio",NAME="en",DEFAULT=YES,AUTOSELECT=YES,LANGUAGE="en",URI="media.m3u8"
+#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=564300,CODECS="mp4a.40.2,avc1.420015",AUDIO="audio"
+media.m3u8
```
gesinger: The URI is the same for both alt audio and muxed.

brandonocasey: yeah for my use case I only needed the m3u8 to be an existing manifest, in `demuxed` it isn't. That is why I added `demuxed-two`, as I don't want to break or change anything else.

brandonocasey: If you want I can make it so that they use different manifests, but they won't actually be used for this test and generating valid manifests might take a bit more time.

gesinger: I'm not sure I understand. I was just pointing out that they're the same because I'm not sure this would be a valid manifest. I don't see something specific in the HLS spec that goes against it, but it would be strange to see in the real world.

PR #660, opened 2019-09-17
fix: improve timestampOffset calculation for fmp4s

alex-barstow:
## Description
Starting in v69, Chrome started rolling out [changes](https://developers.google.com/web/updates/2018/08/chrome-69-media-updates#pts) to report media buffer ranges and durations based on `pts` (presentation timestamp) values instead of `dts` (decode timestamp). Until recently, this project was still expecting the former behavior, setting `timestampOffset` values using dts values. This occasionally conflicted with the new Chrome behavior, resulting in gaps in the buffer that negatively impacted playback. We addressed this issue in #616 for .ts segments, but apparently not fmp4s. This PR achieves the same thing as #616 for fmp4 segments by modifying the `probeMp4Segment_()` function to use "composition time" values instead of decode time. (**Note:** fmp4 timing does not rely on the same concepts of "pts" and "dts" as m2ts timing, instead using "decode time" and "composition time", the latter being a value calculated from the  decode time and a "composition time offset".)

The specific values we need to calculate the composition start time (in seconds) for a given fmp4 segment are: 
- the base media decode time of the fragment
- the composition time offset of the fragment
- the timescale of the media
  - From the ISO BMFF spec:
    > timescale is an integer that specifies the time-scale for this media; this is the number of time units that pass in one second. For example, a time coordinate system that measures time in sixtieths of a second has a time scale of 60.
  - timescale information is contained within the initialization segment, so we also need to get the specific `trackId` of a given media segment in order to associate it with a timescale.

The final calculation is:

`compositionStartTimeInSeconds = (baseMediaDecodeTime + compositionTimeOffset) / timescale`

## Specific Changes proposed
- The mux.js version bump exposes [parseTraf()](https://github.com/videojs/mux.js/commit/7ac044db5aea7bfd976ad357b5f5bc81bd9dbad7) to simplify retrieval of the base media decode time, composition time offset, and track id, since they are all contained within the `traf` box.
- Inside `probeMp4Segment_()`, parse the first `traf` box to get the information we need, then calculate and return the composition start and end times in seconds
  - Unlike `probeTsSegment_()`, the mp4 probe function also has the side effect of setting the `timestampOffset`. We may want to change this in the future so, like the ts probe, it simply returns the timing information and the `timestampOffset` can be set in `handleSegment_()`.

### DIFF
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -456,6 +456,7 @@ export default class SyncController extends videojs.EventTarget {
     // so we use it here for consistency
     timescale = timescales[trackId] || 90e3;
 
+    // calculate the composition start time, in seconds
     compositionStartTime = (baseMediaDecodeTime + compositionTimeOffset) / timescale;
 
     if (segmentInfo.timestampOffset !== null) {
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -4,7 +4,6 @@
 
 import mp4probe from 'mux.js/lib/mp4/probe';
 import tsInspector from 'mux.js/lib/tools/ts-inspector.js';
-import mp4Inspector from 'mux.js/lib/tools/mp4-inspector.js';
 import { sumDurations } from './playlist';
 import videojs from 'video.js';
 import logger from './util/logger';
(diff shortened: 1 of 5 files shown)


src/sync-controller.js
```diff
@@ -415,25 +416,57 @@ export default class SyncController extends videojs.EventTarget {
...
    */
   probeMp4Segment_(segmentInfo) {
-    let segment = segmentInfo.segment;
-    let timescales = mp4probe.timescale(segment.map.bytes);
-    let startTime = mp4probe.startTime(timescales, segmentInfo.bytes);
+    const segment = segmentInfo.segment;
+    // get timescales from init segment
+    const timescales = mp4probe.timescale(segment.map.bytes);
+    // get `traf` boxes from the media segment
+    const trafBoxes = mp4probe.findBox(segmentInfo.bytes, ['moof', 'traf']);
```
mjneil: I feel like this logic would be better suited staying within mux.js and creating a new function like `mp4probe.compositionStartTime`

test/segment-loader.test.js
```diff
@@ -233,7 +229,35 @@ QUnit.module('SegmentLoader: M2TS', function(hooks) {
       loader.mimeType(this.mimeType);
       loader.load();
 
-      this.startTime.returns(11);
+      mp4probe.timescale = function() {
```
mjneil: If you overwrite functions on an import like this, you should save the original functions to a variable and restore them at the end the test, or they could impact over tests in the module.

src/sync-controller.js
alex-barstow: Agreed, done: https://github.com/videojs/mux.js/pull/304

test/segment-loader.test.js
alex-barstow: No longer relevant since the tests have been moved to mux: https://github.com/videojs/mux.js/pull/304

alex-barstow:
Made the recommended changes and moved the box parsing logic (and tests) into a new function in mux.js called `compositionStartTime()`: https://github.com/videojs/mux.js/pull/304

Once that's approved I'll bump the mux version in this PR.

brandonocasey:
We also need to make sure that we don't forget to add a similar fix to master.

Issue #664, opened 2019-10-01
HLS: wrong duration for live DVR with discontinuity and program date time.


gjanblaszczyk:
__Please do not delete the template, by filling out the required information we can investigate your issue more quickly.__

Before opening an issue see if your problem can be resolved with the [troubleshooting guide](https://github.com/videojs/http-streaming/blob/master/docs/troubleshooting.md). 

## Description
The live DVR stream has the wrong duration.

## Sources
The problematic HLS playlist looks like this:

#EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:5
#EXTINF:5.000,
#EXT-X-PROGRAM-DATE-TIME:2019-09-25T08:06:55+00:00
1569398812.ts
#EXTINF:5.000,
1569398817.ts
#EXT-X-DISCONTINUITY
#EXTINF:5.000,
#EXT-X-PROGRAM-DATE-TIME:2019-09-25T08:10:05+00:00
1569399005.ts

You can find it here: https://demo.screen9.com:8081/demos/grbla/hls/video1/video2/playlist-short.m3u8

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.

1. Navigate to https://codepen.io/grbla/pen/LYPwGea and play the video.
2. Checks the video duration. It should be around 2 min but It is around 3 min.

### Expected
The duration of the video should be around 2 min but It is 3 min.

### Error output
If there are any errors in the console, from the player, or anywhere else please include them here:

## Additional Information
Please include any additional information necessary here. Including the following:

### videojs-http-streaming version
what version of videojs-http-streaming does this occur with?
videojs-http-streaming 1.11.0-alpha.1

### videojs version
what version of videojs does this occur with?
video.js 7.6.5

### Browsers
what browsers are affected? please include browser and version for each
* Chrome and Firefox

### Platforms
what platforms are affected? please include operating system and version or device and version for each
* Windows 10

### Other Plugins
are any other videojs plugins being used on the page? If so, please list them with version below.
*

### Other JavaScript
are you using any other javascript libraries or frameworks on the page? if so please list them below.
*

PR #665, opened 2019-10-01
Feat: improve program date sync strategy

gjanblaszczyk:
## Description
This PR fixes problem described in #664 .

## Specific Changes proposed
Improves programDateTime strategy to handle segment discontinuity correctly.

## Requirements Checklist
- [X] If necessary, more likely in a feature request than a bug fix
  - [ ] Docs/guides updated

### DIFF
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -34,6 +34,7 @@ export const syncPointStrategies = [
       const segments = playlist.segments || [];
       let syncPoint = null;
       let lastDistance = null;
+      let totalSegmentTime = 0;
 
       currentTime = currentTime || 0;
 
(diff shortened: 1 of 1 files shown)


src/sync-controller.js
```diff
@@ -42,7 +43,12 @@ export const syncPointStrategies = [
 
         if (segment.dateTimeObject) {
           const segmentTime = segment.dateTimeObject.getTime() / 1000;
-          const segmentStart = segmentTime + syncController.datetimeToDisplayTime;
+          let segmentStart = segmentTime + syncController.datetimeToDisplayTime;
+
+          // make sure segment start time is not higher than total segment time
+          if (segment.discontinuity && segmentStart > totalSegmentTime) {
+            segmentStart = totalSegmentTime;
```
mjneil: This will not work for a real live playlist. The point of `segmentStart` here is to know the start time of the segment relative to when the stream was first joined. In a live playlist, segments fall off the window, so starting `totalSegmentTime = 0` is inaccurate as you lose the total segment time that has fallen off the live window. 

What I think should be done is `syncController.datetimeToDisplayTime` should be tracked per discontinuity timeline, instead of once at the start for the whole stream, similar to how the time mapping objects are stored per timeline in `syncController.timelines`

gjanblaszczyk: Thanks, this is a really good idea!

PR #672, opened 2019-10-10
1.x: support multiple stream-inf with same URI

### LINKED ISSUE #663: Multiple STREAM-INF with same URI not supported
Since the playlist loader uses the playlist `URI` as a key to index into the playlists array, if there are multiple `EXT-X-STREAM-INF` tags using the same `URI` (e.g. to support different audio codecs for different devices), the playlist object references get overwritten at this key and can prevent playback on some devices.

Example source: https://hls-vod-cld.cdn1.smithsonianchannel.com/8405be-thisfemaleleopard_88867508/mobile_master.m3u8

mjneil:
## Description
Fixes #663 

backport of #670 

## Specific Changes proposed
Instead of using playlist `URI` as a unique identifier for streams, create an `id` property that is the combination of `uri` and index within the master manifest.

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -471,7 +471,7 @@ export default class DashPlaylistLoader extends EventTarget {
 
         properties.playlists[0].uri = phonyUri;
         properties.playlists[0].id = id;
-        // setup ID and URI references (URI for backwards compatability)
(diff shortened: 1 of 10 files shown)


gkatsev:
Marked as blocked because @mjneil realized that this could potentially be a breaking change.

PR #683, opened 2019-11-12
Extract manifest modification functions from playlist loaders to manifest module

gesinger:
## Description

After parsing an HLS or DASH manifest string via m3u8-parser or mpd-parser, the respective playlist loader (playlist-loader or dash-playlist-loader) would modify the manifest objects to add properties or change their structure so that VHS could use them as if they were standardized.

The functions for doing this lived in playlist-loader and dash-playlist-loader, however, with the upcoming possibility of passing in a vhs-json object as the source (representing a manifest string already parsed by m3u8-parser/mpd-parser), the functions needed to be extracted to be reused appropriately, in a way that could act on either HLS or DASH playlists in the same fashion.

This extracts that functionality to the manifests.js module.

## Requirements Checklist
  - [ ] Docs/guides updated


src/manifest.js
```diff
@@ -0,0 +1,187 @@
...
+
+      properties.playlists[0].uri = phonyUri;
+      properties.playlists[0].id = id;
+      // set up ID references
+      master.playlists[id] = properties.playlists[0];
+    }
+  });
+
+  setupMediaPlaylists(master);
+  resolveMediaGroupUris(master);
```
brandonocasey: Should we return master here? Seems like the function this was based on used to do that.

src/manifest.js
```diff
@@ -0,0 +1,187 @@
...
+      'VIDEO': {},
+      'CLOSED-CAPTIONS': {},
+      'SUBTITLES': {}
+    },
+    uri: window.location.href,
+    playlists: [{
+      uri,
+      id,
+      resolvedUri: uri,
+      attributes: {}
```
brandonocasey: Should we add back some of the old inline comments from the code this function is based on?

src/manifest.js
```diff
@@ -0,0 +1,187 @@
...
+      // throughout VHS, but some formats (e.g., DASH) don't have external URIs
+      // TODO: consider adding dummy URIs in mpd-parser
+      const phonyUri = `placeholder-uri-${i}`;
+
+      master.playlists[i].uri = phonyUri;
+    }
+  }
+
+  forEachMediaGroup(master, (properties, mediaType, groupKey, labelKey) => {
+    if (properties.playlists &&
```
brandonocasey: Should we return here rather than doing it in the `if` scope?

src/manifest.js
gesinger: Returning master would make sense, and I can definitely add it, but where did you see it used to be returned?

brandonocasey: Looking back parseXml returned it and since it still does I guess that we don't need to do it here.

PR #698, opened 2019-12-11
feat: suggestedPresentationDelay

gkatsev:
Use the suggestedPresentation delay when available instead of calculating it like it is done in HLS.

Depends on https://github.com/videojs/mpd-parser/pull/82


src/playlist.js
```diff
@@ -223,14 +223,14 @@ export const sumDurations = function(playlist, startIndex, endIndex) {
...
-export const safeLiveIndex = function(playlist) {
+export const safeLiveIndex = function(playlist, suggestedPresentationDelay) {
   if (!playlist.segments.length) {
     return 0;
   }
 
   let i = playlist.segments.length - 1;
   let distanceFromEnd = playlist.segments[i].duration || playlist.targetDuration;
-  const safeDistance = distanceFromEnd + playlist.targetDuration * 2;
+  const safeDistance = suggestedPresentationDelay || distanceFromEnd + playlist.targetDuration * 2;
```
misteroneill: Super minor, but would it be clearer to have parentheses around the second part of the `||`?

src/playlist.js
```diff
@@ -292,10 +292,10 @@ export const playlistEnd = function(playlist, expired, useSafeLiveEnd) {
   * @return {TimeRanges} the periods of time that are valid targets
   * for seeking
   */
-export const seekable = function(playlist, expired) {
+export const seekable = function(playlist, expired, suggestedPresentationDelay) {
```
gesinger: I think from here and through the functions that get called, we should rename `suggestedPresentationDelay`. It does obfuscate the source a little, but in theory this logic should be generic for any streaming format, and since it's used if provided, it goes a bit beyond suggested. Maybe something as simple as `presentationDelay`.

gesinger: We may also need a jsdoc for the param (and for functions that get called with it).

src/playlist.js
```diff
@@ -256,7 +256,7 @@ export const safeLiveIndex = function(playlist) {
  * @returns {Number} the end time of playlist
  * @function playlistEnd
  */
-export const playlistEnd = function(playlist, expired, useSafeLiveEnd) {
+export const playlistEnd = function(playlist, expired, useSafeLiveEnd, suggestedPresentationDelay) {
```
gesinger: The jsdoc for `useSafeLiveEnd` may need to be changed, since it assumes a value for presentation delay.

src/master-playlist-controller.js
```diff
@@ -925,6 +925,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
     }
 
     let media = this.masterPlaylistLoader_.media();
+    const suggestedPresentationDelay = this.masterPlaylistLoader_.master.suggestedPresentationDelay
```
gesinger: Minor, but might be better to move this closer to where it's used.

src/playlist.js
gkatsev: Added docs everywhere and renamed it liveEdgePadding

src/playlist.js
```diff
@@ -216,21 +216,28 @@ export const sumDurations = function(playlist, startIndex, endIndex) {
...
-export const safeLiveIndex = function(playlist) {
+export const safeLiveIndex = function(playlist, liveEdgePadding) {
   if (!playlist.segments.length) {
     return 0;
   }
 
   let i = playlist.segments.length - 1;
   let distanceFromEnd = playlist.segments[i].duration || playlist.targetDuration;
-  const safeDistance = distanceFromEnd + playlist.targetDuration * 2;
+  const safeDistance = liveEdgePadding || (distanceFromEnd + playlist.targetDuration * 2);
```
mjneil: What if someone wants to explicitly put `0` in the MPD for suggestedPresentationDelay? (IOP recommends it be at least `minBufferTime`, but nothing in spec prevents it from being set to `0`)

gkatsev: Yup, that should be valid. I guess that's why having some tests added would be good.

src/playlist.js
```diff
@@ -216,21 +216,34 @@ export const sumDurations = function(playlist, startIndex, endIndex) {
  * window which is the duration of the last segment plus 2 target durations from the end
  * of the playlist.
  *
+ * A liveEdgePadding can be provided which will be used instead of calculating the safe live edge.
+ * This corresponds to suggestedPresentationDelay in DASH manifests.
+ *
  * @param {Object} playlist
  *        a media playlist object
+ * @param {Number} liveEdgePadding
```
gesinger: May want to just denote that it's optional

test/master-playlist-controller.test.js
```diff
@@ -1973,6 +1974,7 @@ function(assert) {
   Playlist.seekable = () => {
     return videojs.createTimeRanges(mainTimeRanges);
   };
+  this.masterPlaylistController.masterPlaylistLoader_.master = media;
```
gesinger: Not a big deal, but may be worth adding a comment saying it's just a placeholder, or putting in a different value, since the master should always reference either a master playlist object or a skeleton of one created to wrap around a media playlist.

test/playlist.test.js
```diff
@@ -492,6 +492,113 @@ QUnit.test('safeLiveIndex is 0 when no safe live point', function(assert) {
     'returns media index 0 when playlist has no safe live point');
 });
 
+QUnit.test('safeLiveIndex is accounts for liveEdgePadding in simple case', function(assert) {
```
gesinger: s/is//

test/playlist.test.js
```diff
@@ -492,6 +492,113 @@ QUnit.test('safeLiveIndex is 0 when no safe live point', function(assert) {
...
+    'returns 4 when liveEdgePadding is 12 and duration is 6');
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 6), 5,
+    'returns 5 when liveEdgePadding is 6 and duration is 6');
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 0), 6,
+    'returns 6 when liveEdgePadding is 0 and duration is 6');
+});
+
+QUnit.test('safeLiveIndex is accounts for liveEdgePadding in non-simple case', function(assert) {
```
gesinger: s/is//

test/playlist.test.js
```diff
@@ -492,6 +492,113 @@ QUnit.test('safeLiveIndex is 0 when no safe live point', function(assert) {
...
+    'returns 2 when liveEdgePadding is 12');
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 6), 3,
+    'returns 3 when liveEdgePadding is 6');
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 4), 3,
+    'returns 4 when liveEdgePadding is 5');
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 1), 4,
+    'returns 5 when liveEdgePadding is 0.5');
```
gesinger: I think the message should be different

test/playlist.test.js
```diff
@@ -492,6 +492,113 @@ QUnit.test('safeLiveIndex is 0 when no safe live point', function(assert) {
...
+    'returns 1 when liveEdgePadding is 18');
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 12), 2,
+    'returns 2 when liveEdgePadding is 12');
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 6), 3,
+    'returns 3 when liveEdgePadding is 6');
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 4), 3,
+    'returns 4 when liveEdgePadding is 5');
```
gesinger: I think the message should be different

test/playlist.test.js
```diff
@@ -492,6 +492,113 @@ QUnit.test('safeLiveIndex is 0 when no safe live point', function(assert) {
...
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 1), 4,
+    'returns 5 when liveEdgePadding is 0.5');
+
+  assert.equal(Playlist.safeLiveIndex(playlist, 0.5), 5,
+    'returns 5 when liveEdgePadding is 0.5');
+
```
gesinger: Would be worth adding a case around 0.25

src/playlist.js
```diff
@@ -216,21 +216,34 @@ export const sumDurations = function(playlist, startIndex, endIndex) {
...
   }
 
-  let i = playlist.segments.length - 1;
-  let distanceFromEnd = playlist.segments[i].duration || playlist.targetDuration;
-  const safeDistance = distanceFromEnd + playlist.targetDuration * 2;
+  let i = playlist.segments.length;
+  let lastSegmentDuration = playlist.segments[i - 1].duration || playlist.targetDuration;
+  const safeDistance = typeof liveEdgePadding === 'number' ? liveEdgePadding : lastSegmentDuration + playlist.targetDuration * 2;
+
+  if (safeDistance === 0) {
```
gesinger: Do we want this to account for any distance less than last segment duration?

test/playlist.test.js
gkatsev: Suggestions for another message?

test/playlist.test.js
gkatsev: We can, though, should be equivalent 0 and 0.5.

test/playlist.test.js
gesinger: returns 3 when liveEdgePadding is 4

test/playlist.test.js
gesinger: It should be, but I think it's a distinct enough case in that 0 is a special case, and the last segment's duration is a separate case.

test/playlist.test.js
gesinger: returns 4 when liveEdgePadding is 1

test/playlist.test.js
gkatsev: oh, haha, completely missed that this doesn't actually match the code

test/playlist.test.js
gkatsev: yeah, I guess doesn't hurt to have it there.

src/playlist.js
gkatsev: no, I think we should only special-case 0 exactly. Any other value should be ran through the loop below to find the appropriate return value.

test/playlist.test.js
gkatsev: also, yeah, misspoke about it matching 0. It should match 0.5.

gkatsev: added

PR #707, opened 2020-01-07
fix: race condition preventing qualityLevels from being populating

### LINKED ISSUE #677: Qualities not loading and addqualitylevel event not triggered
## Description
Hello,
If i open the player, in a new tab WITHOUT FOCUSING IT (middle mouse button open for example), and let some time pass (so all javascript may run), the quality levels are not loaded and addqualitylevel event is not even triggered... if you do the same but focusing the window, all runs good.

## Steps to reproduce

1. Open a video page with m3u8 source on a new tab
2. Wait a couple of seconds
3. Go see it (no qualities loaded)

### Expected
Qualities Loaded

videojs-http-streaming 1.9.3

### videojs version
video.js 7.6.5 (im sure that this happens since 7.1.0 or older)

### Browsers
Only tested with Chrome

### Platforms
Only tested with windows 10

### Other Plugins
*

### Other JavaScript
*

brandonocasey:
## Description
Quality levels is not being populated due to a race condition where the tech is not ready until after we have fired `selectedintialmedia`. You can get this to happen by in chrome by loading the tab in the background for 10s or by using a slow browser like ie 11, where it will happen in the main tab.

## Steps to reproduce
1. Open `index.html`
2. Open the link to `Stats` in a new page, and wait 10s before switching to that tab.
3. Note that the `qualityLevels` do not show up on the page, and `player.qualityLevels().levels_` is a 0 length array
4. Note that `player.representations()` has the correct representations in the returned array. 
5. Note that the `qualityLevels` buttons do not show up on the page.

## Specific Changes proposed
1. The first commit deploys the stats page, and fixes a bug with the stats page. That allows this bug to be tested before the fix.  
2. The second commit remove the tech ready wrapper around `this.setupQualityLevels_()`

## Testing
1. Open the [netlify deploy from #708](https://deploy-preview-708--videojs-http-streaming.netlify.com/) and follow the steps to reproduce.
2. Open [netlify deploy for this pull request](https://deploy-preview-707--videojs-http-streaming.netlify.com/) and follow the steps to reproduce, noting that steps 3, 4, and 5 are no longer an issue.

## Requirements Checklist

Closes #708
Fixes #677

### DIFF
--- a/scripts/netlify.js
+++ b/scripts/netlify.js
@@ -1,12 +1,14 @@
 const path = require('path');
 const sh = require('shelljs');
-const fs = require('fs');
-
-const vjs = 'node_modules/video.js/dist/alt/video.core.js';
-const vjsCss = 'node_modules/video.js/dist/video-js.css';
-const eme = 'node_modules/videojs-contrib-eme/dist/videojs-contrib-eme.js';
 const deployDir = 'deploy';
-const files = [vjs, vjsCss, eme];
+
+const files = [
+  'node_modules/video.js/dist/video-js.css',
+  'node_modules/video.js/dist/alt/video.core.js',
+  'node_modules/videojs-contrib-eme/dist/videojs-contrib-eme.js',
+  'node_modules/videojs-contrib-quality-levels/dist/videojs-contrib-quality-levels.js',
+  'node_modules/d3/d3.min.js'
+];
 
 // cleanup previous deploy
 sh.rm('-rf', deployDir);
@@ -20,5 +22,5 @@ files
 
 // copy over files, dist, and html files
 files
-.concat('dist', 'index.html', 'index.min.html')
+.concat('dist', 'index.html', 'index.min.html', 'utils')
 .forEach((file) => sh.cp('-r', file, path.join(deployDir, file)));
--- a/scripts/netlify.js
+++ b/scripts/netlify.js
@@ -20,7 +20,7 @@ files
 .map((file) => path.dirname(file))
 .forEach((dir) => sh.mkdir('-p', path.join(deployDir, dir)));
 
-// copy over files, dist, and html files
+// copy files/folders to deploy dir
 files
 .concat('dist', 'index.html', 'index.min.html', 'utils')
 .forEach((file) => sh.cp('-r', file, path.join(deployDir, file)));
--- a/src/rendition-mixin.js
+++ b/src/rendition-mixin.js
@@ -92,6 +92,9 @@ let renditionSelectionMixin = function(hlsHandler) {
 
   // Add a single API-specific function to the HlsHandler instance
   hlsHandler.representations = () => {
+    if (!playlists || !playlists.master || !playlists.master.playlists) {
+      return [];
+    }
     return playlists
       .master
       .playlists
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -737,17 +737,21 @@ class HlsHandler extends Component {
   setupQualityLevels_() {
     let player = videojs.players[this.tech_.options_.playerId];
 
-    if (player && player.qualityLevels) {
-      this.qualityLevels_ = player.qualityLevels();
-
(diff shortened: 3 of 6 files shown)


scripts/netlify.js
```diff
@@ -1,12 +1,14 @@
...
-const fs = require('fs');
-
-const vjs = 'node_modules/video.js/dist/alt/video.core.js';
-const vjsCss = 'node_modules/video.js/dist/video-js.css';
-const eme = 'node_modules/videojs-contrib-eme/dist/videojs-contrib-eme.js';
 const deployDir = 'deploy';
-const files = [vjs, vjsCss, eme];
+
+const files = [
+  'node_modules/video.js/dist/video-js.css',
```
brandonocasey: naming files what getting a bit absurd in here

src/rendition-mixin.js
```diff
@@ -92,6 +92,9 @@ let renditionSelectionMixin = function(hlsHandler) {
 
   // Add a single API-specific function to the HlsHandler instance
   hlsHandler.representations = () => {
+    if (!playlists || !playlists.master || !playlists.master.playlists) {
```
brandonocasey: We should probably return an empty array if we don't have playlists yet, Right now we will just throw an error and some of the tests where failing due to this.

src/videojs-http-streaming.js
```diff
@@ -737,17 +737,21 @@ class HlsHandler extends Component {
   setupQualityLevels_() {
     let player = videojs.players[this.tech_.options_.playerId];
 
-    if (player && player.qualityLevels) {
-      this.qualityLevels_ = player.qualityLevels();
+    // if there isn't a player or there isn't a qualityLevels plugin
+    // or qualityLevels_ listeners have already been setup, do nothing.
+    if (!player || !player.qualityLevels || this.qualityLevels_) {
```
brandonocasey: Only setup quality levels listeners once, the rest of the code is the same except we return early instead of adding the logic in an `if` scope.

utils/stats/index.html
```diff
@@ -20,7 +20,7 @@
 
   <!-- player stats visualization -->
   <link href="stats.css" rel="stylesheet">
-  <script src="/node_modules/d3/d3.min.js"></script>
+  <script src="../../node_modules/d3/d3.min.js"></script>
```
brandonocasey: have to use local paths to get it to work on netlify.

utils/stats/index.html
```diff
@@ -345,25 +345,23 @@ <h3>Timed Metadata</h3>
         videojs.Hls.displayStats(document.querySelector('.switching-stats'), player);
         videojs.Hls.displayCues(document.querySelector('.segment-timeline'), player);
 
-        player.ready(function() {
-          var qualityLevels = player.qualityLevels();
+        var qualityLevels = player.qualityLevels();
```
brandonocasey: quality levels can exist before tech/player ready

test/videojs-http-streaming.test.js
```diff
@@ -3094,8 +3094,7 @@ QUnit.test('passes useCueTags hls option to master playlist controller', functio
   videojs.options.hls = origHlsOptions;
 });
 
-// TODO: This test fails intermittently. Turn on when fixed to always pass.
-QUnit.skip('populates quality levels list when available', function(assert) {
+QUnit.test('populates quality levels list when available', function(assert) {
```
brandonocasey: This test was previously skipped because of intermittent failures. I think fixing this bug fixes the test.

scripts/netlify.js
```diff
@@ -20,5 +22,5 @@ files
 
 // copy over files, dist, and html files
```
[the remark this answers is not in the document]
brandonocasey: updated the comment to make it more relevant

gkatsev:
This fixes https://github.com/videojs/http-streaming/issues/677 as well.

gkatsev:
@brandonocasey can you make this change against master as well? Thanks!

PR #708, opened 2020-01-07
ignore: update netlify to deploy stats page. Fix stats

brandonocasey:
his branch contains the the stats page deploy updates and fixes from #707 but not the fix itself so that it can be used to test #707


gkatsev:
Should just be `chore:`, `ignore` is only for internal projects.

gkatsev:
Ah, I guess we're not merging this anyway as it'll be part of #707.

brandonocasey:
I think the deploy will stay around, so I will just close this one to avoid further confusion.

PR #716, opened 2020-01-15
feat: add option to pass an object as the source

gesinger:
Updated version of https://github.com/videojs/http-streaming/pull/649

## Description

This feature adds support for passing a JSON object as the source via a data URI, which VHS looks for alongside a new vendor specific media type, 'application/vnd.videojs.vhs+json'. If present, VHS will use the provided manifest object as the first requested manifest. See the README for more details.

## Specific Changes Proposed

Besides the addition of the manifest object source option and the media type application/vnd.videojs.vhs+json, modifications had to be made through the pipeline, from the top-level of the plugin at videojs-http-streaming.js through the *playlist-loader.js and media-groups.js modules. By the time the segment loaders are reached, the handling should be the same as if the playlist was loaded through the normal mechanism of a request to the src URL.

## Examples

```

player.src({
  src: `data:application/vnd.videojs.vhs+json,${JSON.stringify(master)}`,
  type: 'application/vnd.videojs.vhs+json'
});
```

```

});

player.src({
  src: `data:application/vnd.videojs.vhs+json,${JSON.stringify(master)}`,
  type: 'application/vnd.videojs.vhs+json'
});
```

## Requirements Checklist
  - [X] Docs/guides updated


src/videojs-http-streaming.js
```diff
@@ -226,6 +226,26 @@ const updateVhsLocalStorage = (options) => {
...
+ *
+ * @param {string} dataUri
+ *        The data URI
+ *
+ * @return {string|Object}
+ *         The parsed object/string, or the original string if no supported media type
+ *         was found
+ */
+const expandDataUri = (dataUri) => {
+  if ((/^data:application\/vnd\.videojs\.vhs\+json/i).test(dataUri)) {
```
brandonocasey: Might be better to use `toLower` on `dataUri` here instead of using regex. We also might want to require the comma (`,`) at the end. as currently this would match `data:application/vnd.videojs.vhs+json` and then die since there isn't any json data or comma.

brandonocasey: Might be better to use `toLower` on `dataUri` here instead of using regex as that will make it much easier to see what the uri looks like and it is also much more performant. We also might want to require the comma (`,`) at the end. as currently this would match `data:application/vnd.videojs.vhs+json` and then die since there isn't any json data or comma.

src/master-playlist-controller.js
```diff
@@ -73,8 +73,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
       handlePartialData
     } = options;
 
-    if (!url) {
-      throw new Error('A non-empty playlist URL is required');
+    if (!src) {
+      throw new Error('A non-empty playlist URL or playlist is required');
```
brandonocasey: perhaps this should read json manifest string, I think playlist is too generic.

gkatsev:
Do we actually want to expose the m3u8 parser here? I realize it's useful but it means that we won't be able to fiddle with it as much (not that we have done recently) since it'll now be the public API of VHS. I'm not opposed to it, just want us to think about it explicitly and make an informed decision.

PR #734, opened 2020-02-11
fix: add dispose functions and fix memory leaks #643 for 1.x

brandonocasey:
#643 for 1.x


src/mse/html-media-source.js
```diff
@@ -340,4 +340,17 @@ export default class HtmlMediaSource extends videojs.EventTarget {
     this.sourceBuffers.push(buffer);
     return buffer;
   }
+
+  dispose() {
```
brandonocasey: This and the `virtual-source-buffer` dispose are the only changes from `master` as they did not exist.

PR #736, opened 2020-02-13
fix: resume live DASH playlist refreshes after pausing and loading DA…

gesinger:
…SH playlist loader

This particular issue was noticed when rendition switches while playing
live DASH resulted in the audio buffer no longer filling, and the content
stopping playback after finishing the currently buffered media.

In the code, the media groups module paused the audio playlist loader, which
cleared its refresh timeout, but on calling load to resume the loader, the
refresh timeout was not set, as it is in the normal playlist loader. The
added logic matches the normal playlist loader's load function.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -438,7 +438,11 @@ export default class DashPlaylistLoader extends EventTarget {
       return;
     }
 
-    this.trigger('loadedplaylist');
+    if (media && !media.endList) {
+      this.trigger('mediaupdatetimeout');
+    } else {
+      this.trigger('loadedplaylist');
+    }
   }
 
   /**
--- a/test/dash-playlist-loader.test.js
+++ b/test/dash-playlist-loader.test.js
@@ -2191,3 +2191,42 @@ QUnit.test('child loaders wait for async action before moving to HAVE_MASTER', f
   // media playlist is chosen automatically
   assert.strictEqual(childLoader.state, 'HAVE_METADATA');
 });
+
+QUnit.test('load resumes the media update timer for live playlists', function(assert) {
+  const loader = new DashPlaylistLoader('dash-live.mpd', this.fakeHls);
+
+  loader.load();
+  this.standardXHRResponse(this.requests.shift());
+  this.clock.tick(1);
+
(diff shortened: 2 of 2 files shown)


src/dash-playlist-loader.js
```diff
@@ -438,7 +438,11 @@ export default class DashPlaylistLoader extends EventTarget {
       return;
     }
 
-    this.trigger('loadedplaylist');
+    if (media && !media.endList) {
```
mjneil: Now that this is more closely matching `PlaylistLoader` it got me thinking, can we remove this function override all together and let `PlaylistLoader.load` be called since this extends that class.

The only difference between the two functions now is the above line

```
window.clearTimeout(this.minimumUpdatePeriodTimeout_);
```
This timeout is meant to only be used on the master playlist loader loading the actual mpd xml file and updating the string reference on its own timer. The the Audio playlist loader has its own `mediaupdatetimeout` we are used to that just re-parses the current xml string.

Small investigation found that this timeout only gets started once, in `onClientServerClockSync_`, which only happens in `start`. (This is a fix for another time, but I think this should be resyncing client server clock every time the mpd is refreshed, I believe thats what shaka and dash-js do). 

This means that if `masterPlaylistLoader_.load` is called at any time, refreshing of the mpd will stop. I've found only one place this happens, in `blacklistCurrentPlaylist ` https://github.com/videojs/http-streaming/blob/1.x/src/master-playlist-controller.js#L800

I tested this easily by disabling all representations except one, and then forcing an error for the last representation. (I just made chrome network go "offline" but you could force 404 from charles or numerous other things), and it stopped refreshing the mpd and eventually leads to player timeout. 

I think a check similar to `onClientServerClockSync_` https://github.com/videojs/http-streaming/blob/1.x/src/dash-playlist-loader.js#L637 would be needed here to kick start that process again. The exact placement in relation to these other triggers I'm not 100% sure on

gesinger: After looking into it, I think we may want to separate the issues into different PRs. This case seems more critical at the moment, and we should resolve the issue you've pointed out soon, but I think separating the PRs makes sense.

In addition, another timing related bug was found in examining the issue you found that should be another PR.

PR #744, opened 2020-02-19
Take devicePixelRatio into account

EhsanCh:
## Description
Currently player dimensions is not determined correctly and cause lower quality playlist to be played.
This fix calculate real player dimensions on devices that have devicePixelRatio greater than 1 (retina displays and most mobile devices).

## Specific Changes proposed
multiply player width and height by devicePixelRatio .

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-selectors.js
+++ b/src/playlist-selectors.js
@@ -121,9 +121,9 @@ export const comparePlaylistResolution = function(left, right) {
  * @param {number} playerBandwidth
  *        Current calculated bandwidth of the player
  * @param {number} playerWidth
- *        Current width of the player element
+ *        Current width of the player element (should account for the device pixel ratio)
  * @param {number} playerHeight
- *        Current height of the player element
+ *        Current height of the player element (should account for the device pixel ratio)
  * @param {boolean} limitRenditionByPlayerDimensions
  *        True if the player width and height should be used during the selection, false otherwise
  * @return {Playlist} the highest bitrate playlist less than the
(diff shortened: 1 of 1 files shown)


EhsanCh:
Because simpleSelector method is getting w/h from its parameters, So we should account for the device pixel where we get dimensions from player, and simpleSelector should rely on values it get. 
Also we can send custom w/h to simpleSelector if we leave it as a base method.
Anyway i'm not very familiar with your design structure. if simpleSelector is not a base function we can remove w/h from method parameters and calculate w/h inside it.

gkatsev:
`simpleSelector` doesn't have access to the player, plus, for testing it makes it easy if the width/height is passed in. It probably doesn't make a lot of difference whether the devicePixelRatio is used since this method is only used internally. If we don't move it inside, can a comment be added to the docs of `simpleSelector` that the width/height that are passed in should account for the device pixel ratio?

gkatsev:
So, in another conversation, I just talked myself out of doing it inside `simpleSelector`. `simpleSelector` gets everything it needs passed in, so, unless we were to pass `devicePixelRatio` in to `simpleSelector`, we should leave `simpleSelector` simple.

gkatsev:
Thanks @EhsanCh

gkatsev:
Not sure it'd worth having tests here because we don't test `lastBandwidthSelector` directly right now and it'll require mocking out `window`.

EhsanCh:
@gkatsev Can you please merge this pull request?

PR #745, opened 2020-02-19
fix: only reset syncController if we passed a discontinuity

gesinger:
See https://github.com/videojs/http-streaming/pull/502

Added some modifications to make testing easier and better organize the code, but let me know if you disagree with those changes.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -121,6 +121,9 @@ const timingInfoPropertyForMedia = (mediaType) => `${mediaType}TimingInfo`;
  *        The estimated segment start
  * @param {TimeRange[]} buffered
  *        The loader's buffer
+ * @param {boolean} overrideCheck
+ *        If true, no checks are made to see if the timestamp offset value should be set,
+ *        but sets it directly to a value.
  *
  * @return {number|null}
  *         Either a number representing a new timestamp offset, or null if the segment is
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -842,43 +882,16 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      segmentTimeline: segmentInfo.timeline,
+      currentTimeline: this.currentTimeline_,
+      startOfSegment: segmentInfo.startOfSegment,
+      buffered
+    });
 
     this.loadSegment_(segmentInfo);
   }
 
-  startsBeforeSourceBufferTimestampOffset(segmentInfo) {
```
gkatsev: so, deleting this block is basically equivalent to the change in #502?

test/segment-loader.test.js
```diff
@@ -1865,6 +1931,213 @@ QUnit.module('SegmentLoader', function(hooks) {
...
+        );
+        assert.equal(
+          transmuxerTimestampOffsets.length,
+          2,
+          'another transmuxer timestamp offset'
+        );
+      });
+    });
+
+    QUnit.test('sets timestamp offset on timeline changes but not if segment start is early', function(assert) {
```
gkatsev: this is so much less readable than https://github.com/videojs/http-streaming/pull/502/files#diff-4efdae9dc0d9508ec02d8c6707ae5df8R445 but idk if there's really anything we can about that.

gkatsev: Maybe it can be annotated with some comments about what's happening?

gkatsev:
Also, verified that the stream plays correctly with the latest changes.

PR #754, opened 2020-03-04
fix(segment-loader): resetEverything should remove through Infinity

gkatsev:
In DASH it's very common to provide the captions as one file rather than
having it be segmented VTT. It's possible to do the same in HLS, though,
it isn't common.

When we switch between tracks, we remove all the cues for the
duration of the video because the cues are no longer needed. Also, as we
re-download segments, the associated cues should be re-added.
Unfortunately, this the case of a single "segment", we don't get more
segments and the cues don't get re-added. Therefore, for non-live when
we have one segment, we shouldn't remove the cues.

This fixes the case both for turning the track on and off but also for
seeking.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -775,7 +775,9 @@ export default class SegmentLoader extends videojs.EventTarget {
    * operation is complete
    */
   remove(start, end, done = () => {}) {
-    // clamp end to duration if we need to remove everything
+    // clamp end to duration if we need to remove everything.
+    // This is due to a browser bug that causes issues if we remove to Infinity.
+    // videojs/videojs-contrib-hls#1225
     if (end === Infinity) {
       end = this.duration_();
     }
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -721,7 +721,11 @@ export default class SegmentLoader extends videojs.EventTarget {
       video: true
     };
     this.resetLoader();
-    this.remove(0, this.duration_(), done);
+
+    // remove from 0, the earliest point, to Infinity, to signify removal of everything.
+    // VTT Segment Loader doesn't need to do anything but in the regular SegmentLoader,
+    // we then clamp the value to duration if necessary.
+    this.remove(0, Infinity, done);
 
     // clears fmp4 captions
     if (this.captionParser_) {
@@ -771,6 +775,11 @@ export default class SegmentLoader extends videojs.EventTarget {
    * operation is complete
    */
   remove(start, end, done = () => {}) {
+    // clamp end to duration if we need to remove everything
+    if (end === Infinity) {
+      end = this.duration_();
+    }
+
     if (!this.sourceUpdater_ || !this.startingMedia_) {
       // nothing to remove if we haven't processed any media
       return;
--- a/src/vtt-segment-loader.js
+++ b/src/vtt-segment-loader.js
@@ -154,6 +154,20 @@ export default class VTTSegmentLoader extends SegmentLoader {
    * @param {number} end - the end time of the region to remove from the buffer
    */
   remove(start, end) {
+    // If we only have "one" segment we likely have the all the captions in a single file,
+    // which is a common configuration for DASH and a possible one for HLS.
+    // It's possible that we have cues that end after the video duration, therefore,
+    // when removing from beginning to duration_(), we want to remove *all* the cues,
+    // and not just those through duration_().
+    // If we won't remove all the cues, checkBuffer_ will assume we're all set and we won't
+    // fill the buffer again with the missing cues.
+    if (this.playlist_ && this.playlist_.endList === true) {
+      const lastCue = this.subtitlesTrack_.cues[this.subtitlesTrack_.cues.length - 1];
+
+      if (end === this.duration_() && lastCue && lastCue.endTime > end) {
+        end = lastCue.endTime;
+      }
+    }
     removeCuesFromTrack(start, end, this.subtitlesTrack_);
   }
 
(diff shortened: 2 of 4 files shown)


src/vtt-segment-loader.js
```diff
@@ -154,6 +154,15 @@ export default class VTTSegmentLoader extends SegmentLoader {
    * @param {number} end - the end time of the region to remove from the buffer
    */
   remove(start, end) {
+    // if we only have "one" segment we likely have the entire caption file in a single
+    // and in that case we shouldn't remove cues because they won't get re-added later on.
+    // Also, if we're live, we probably want to remove them still as more segments could be added.
+    if (this.playlist_ &&
+      this.playlist_.endList &&
```
brandonocasey: Judging by the comment should this be `!this.playlist_.endList`?

gkatsev: We want to skip removal of cues if we are VOD (`endList`) and we have 1 segment. This is an occasion where explicitly checking `endList === true` might be warranted.

brandonocasey: your right :doh:

src/vtt-segment-loader.js
```diff
@@ -154,6 +154,19 @@ export default class VTTSegmentLoader extends SegmentLoader {
    * @param {number} end - the end time of the region to remove from the buffer
    */
   remove(start, end) {
+    // if we only have "one" segment we likely have the entire caption file in a single.
```
mjneil: sentence feels incomplete `in a single.`?

src/vtt-segment-loader.js
```diff
@@ -154,6 +154,19 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
+    // It's possible that we have cues that end after the video duration, therefore,
+    // when removing from beginning to duration_(), we want to remove *all* the cues,
+    // and not just those through duration_().
+    // If we won't remove all the cues, checkBuffer_ will assume we're all set and we won't
+    // fill the buffer again with the missing cues.
+    if (this.playlist_ && this.playlist_.endList === true) {
+      const lastCue = this.subtitlesTrack_.cues[this.subtitlesTrack_.cues.length - 1];
+
+      if (end === this.duration_() && lastCue) {
```
mjneil: We should also check that `lastCue.endTime > end`

gkatsev: good call

src/vtt-segment-loader.js
```diff
@@ -154,6 +154,20 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
   remove(start, end) {
+    // If we only have "one" segment we likely have the all the captions in a single file,
+    // which is a common configuration for DASH and a possible one for HLS.
+    // It's possible that we have cues that end after the video duration, therefore,
+    // when removing from beginning to duration_(), we want to remove *all* the cues,
+    // and not just those through duration_().
+    // If we won't remove all the cues, checkBuffer_ will assume we're all set and we won't
+    // fill the buffer again with the missing cues.
+    if (this.playlist_ && this.playlist_.endList === true) {
+      const lastCue = this.subtitlesTrack_.cues[this.subtitlesTrack_.cues.length - 1];
```
brandonocasey: Could we just check if `end === this.duration_()` and if so just set `end` to `Infinity`? I assume that this was done for the `remove` call that happens in `SegmentLoader#resetEverything` is there any reason not to remove `0` to `Infinity` there rather than `this.duration_()`?

src/vtt-segment-loader.js
```diff
@@ -154,6 +154,20 @@ export default class VTTSegmentLoader extends SegmentLoader {
    * @param {number} end - the end time of the region to remove from the buffer
    */
   remove(start, end) {
+    // If we only have "one" segment we likely have the all the captions in a single file,
+    // which is a common configuration for DASH and a possible one for HLS.
+    // It's possible that we have cues that end after the video duration, therefore,
+    // when removing from beginning to duration_(), we want to remove *all* the cues,
+    // and not just those through duration_().
+    // If we won't remove all the cues, checkBuffer_ will assume we're all set and we won't
```
gesinger: If we don't*

src/vtt-segment-loader.js
```diff
@@ -154,6 +154,20 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
+    // fill the buffer again with the missing cues.
+    if (this.playlist_ && this.playlist_.endList === true) {
```
gesinger: I think the `=== true` can be removed

src/vtt-segment-loader.js
```diff
@@ -154,6 +154,20 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
+      const lastCue = this.subtitlesTrack_.cues[this.subtitlesTrack_.cues.length - 1];
+
+      if (end === this.duration_() && lastCue && lastCue.endTime > end) {
```
gesinger: Will the end always be exactly the duration? We should be basing the duration on information that we have, which may be limited (based on manifest info or on parsed segments), so we may even be shorter or longer than the reported cue times.

gkatsev: Yes because the call here happens from `resetEverything()` which calls `remove(start, this.duration_())`

src/vtt-segment-loader.js
gkatsev: yeah, that's probably true (pun intended). I was thinking that this benefits from the explicit check but `endList` gets set through our parser, so, might be ok without the explicit check.

src/segment-loader.js
```diff
@@ -771,6 +775,11 @@ export default class SegmentLoader extends videojs.EventTarget {
    * operation is complete
    */
   remove(start, end, done = () => {}) {
+    // clamp end to duration if we need to remove everything
+    if (end === Infinity) {
```
gkatsev: this needs to be clamped because of https://github.com/videojs/videojs-contrib-hls/pull/1225

brandonocasey: I would make this a comment so we can remove in the future!

PR #762, opened 2020-03-12
fix: better codec handling

brandonocasey:
## Description
Depends on https://github.com/videojs/vhs-utils/pull/10

Ultimately this pull request solves issues multiple issues:
1. We can now answer what codecs the muxer supports
2. We can rely on muxer support for ts segments
3. We can rely on browser support for fmp4 segments
4. We can rely on muxer and browser support separetly

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1272,7 +1272,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
     if (!audioSupported || !videoSupported) {
       const message = `${mainStartingMedia.isFmp4 ? 'browser' : 'muxer'}` +
         ` does not support ${mainStartingMedia.isFmp4 ? 'fmp4' : 'ts'} codec(s): ` +
-        `${!audioSupported ? codecs.audio : ''} ${!videoSupported ? codecs.video : ''}`;
+        `${!audioSupported ? codecs.audio : ''}${!videoSupported ? codecs.video : ''}`;
 
       // reset startingMedia_ for playlists blacklisted
       this.mainSegmentLoader_.startingMedia_ = void 0;
@@ -1344,6 +1344,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
     const codecCount = Object.keys(codecs).length;
 
     this.master().playlists.forEach((variant) => {
+      // skip variants that are already blacklisted forever
+      if (variant.excludeUntil === Infinity) {
+        return;
+      }
       // TODO: should we really assume codecs?
       let variantCodecs = {};
       let variantCodecCount = 2;
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -318,26 +318,15 @@ export class MasterPlaylistController extends videojs.EventTarget {
         // playlist's `segments` list is already available, a media playlist won't be
         // requested, and loadedplaylist won't fire again, so the playlist handler must be
         // called on its own here.
-        if (this.sourceType_ !== 'vhs-json' && !this.initialMedia_.segments) {
+        const haveJsonSource = this.sourceType_ === 'vhs-json' && this.initialMedia_.segments;
+
+        if (!haveJsonSource) {
           return;
         }
         updatedPlaylist = this.initialMedia_;
       }
 
-      const codecs = updatedPlaylist.attributes.CODECS;
-      const segments = updatedPlaylist.segments || [];
-
-      // For non-fmp4 segments we also need to blacklist based on what the
-      // muxer supports. As non-fmp4 goes through the muxer
-      // but we have to wait for a playlist to download before
-      // we can know if it is fmp4 or not, so we can only really blacklist just
-      // before a playlist would get choosen
-      if (codecs && !segments[0].map && !muxerSupportsCodec(codecs)) {
-        this.blacklistCurrentPlaylist({
-          playlist: updatedPlaylist,
-          message: `muxer does not support codec ${codecs}`,
-          noWarn: true
-        }, Infinity);
+      if (this.blacklistUnsupportedMuxerCodec(updatedPlaylist)) {
         return;
       }
 
@@ -928,14 +917,37 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // Select a new playlist
     const nextPlaylist = this.selectPlaylist();
+    const logFn = error.internal ? this.logger_ : videojs.log.warn;
 
-    if (!error.noWarn) {
-      videojs.log.warn('Problem encountered with the current playlist.' +
-                     (error.message ? ' ' + error.message : '') +
-                     ' Switching to another playlist.');
-    }
+    logFn(`${(error.internal ? 'Internal problem' : 'Problem')} encountered with the current playlist.` +
+      `${error.message ? ` ${error.message}` : '' } Switching to another playlist.`);
 
     return this.masterPlaylistLoader_.media(nextPlaylist, isFinalRendition);
+  }
+
+  /**
+   * Segments that are not mp4 are not transmuxed, therfore, playlists that
+   * contain segments which mux.js does not support should be blacklisted. This can only
+   * happen at time of playlist selection, as for some playlists, that is the only time
+   * when the container format for the segments is known.
+   *
+   * @return {boolean}
+   *         Wether the blacklist happened or not
+   */
+  blacklistUnsupportedMuxerCodec(playlist = this.masterPlaylistLoader_.media()) {
+    const codecs = playlist.attributes.CODECS;
+    const segments = playlist.segments || [];
+
+    if (codecs && !segments[0].map && !muxerSupportsCodec(codecs)) {
+      this.blacklistCurrentPlaylist({
+        playlist,
+        message: `muxer does not support codec ${codecs}`,
+        internal: true
+      }, Infinity);
+      return true;
+    }
+
+    return false;
   }
 
   /**
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -295,10 +295,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
       let updatedPlaylist = this.masterPlaylistLoader_.media();
 
       if (!updatedPlaylist) {
-        // blacklist any variants that are not supported by the browser before selecting
-        // an initial media as the playlist selectors do not consider browser support
-        this.excludeUnsupportedVariants_();
-
         let selectedMedia;
 
         if (this.enableLowInitialPlaylist) {
@@ -324,10 +320,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
           return;
         }
         updatedPlaylist = this.initialMedia_;
-      }
-
-      if (this.blacklistUnsupportedMuxerCodec(updatedPlaylist)) {
-        return;
       }
 
       this.handleUpdatedMediaPlaylist(updatedPlaylist);
@@ -879,9 +871,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return;
     }
 
-    const isFinalRendition =
-      this.masterPlaylistLoader_.master.playlists.filter(isEnabled).length === 1;
     const playlists = this.masterPlaylistLoader_.master.playlists;
+    const enabledPlaylists = playlists.filter(isEnabled);
+    const isFinalRendition = enabledPlaylists.length === 1 && enabledPlaylists[0] === currentPlaylist;
 
     if (playlists.length === 1) {
       // Never blacklisting this playlist because it's the only playlist
@@ -923,54 +915,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
       `${error.message ? ` ${error.message}` : '' } Switching to another playlist.`);
 
     return this.masterPlaylistLoader_.media(nextPlaylist, isFinalRendition);
-  }
-
-  /**
-   * Segments that are not mp4 are not transmuxed, therefore, playlists that
-   * contain segments which mux.js does not support should be blacklisted. This can only
-   * happen at time of playlist selection, as for some playlists, that is the only time
-   * when the container format for the segments is known.
-   *
-   * @param {Object} playlist
-   *        The playlist object that we should check codecs on for blacklisting.
-   *
-   * @return {boolean}
-   *         Whether the blacklist happened or not
-   */
-  blacklistUnsupportedMuxerCodec(playlist = this.masterPlaylistLoader_.media()) {
-    const codecs = playlist.attributes.CODECS;
-    const segments = playlist.segments || [];
-
-    /**
-     * NOTE: The following check is a shorthand until a larger change can be
-     * taken on. This code will probably have issues with ts playlists that contain
-     * EXT-X-MAP, aka the map property on a segment. See the TODO below for more
-     * information on the change that is needed to get ts playlists with EXT-X-MAP
-     * working.
-     *
-     * TODO:
-     * Refactor blacklisting by codec so that it happens after
-     * we download enough of a segment to know whether it is fmp4 or ts.
-     * * If it is fmp4 we check if the browser supports the codec and blacklist
-     *   when it does not, as fmp4 is not transmuxed.
-     * * if it is ts we check if the muxer supports the codec and blacklist
-     *   when it does not, as ts segments will be transmuxed.
-     *
-     * We should probably do this by adding a segmentType handler to mediaSegmentRequest
-     * that calls back with the current segment type and potentially the codec it uses.
-     * Then we check the appropriate support matrix to determine if the playlist should
-     * be blacklisted.
-     */
-    if (codecs && !segments[0].map && !muxerSupportsCodec(codecs)) {
-      this.blacklistCurrentPlaylist({
-        playlist,
-        message: `muxer does not support codec ${codecs}`,
-        internal: true
-      }, Infinity);
-      return true;
-    }
-
-    return false;
   }
 
   /**
@@ -1319,6 +1263,25 @@ export class MasterPlaylistController extends videojs.EventTarget {
       }
     }
 
+    // fmp4 relies on browser support, while ts relies on muxer support
+    const supportFunction = mainStartingMedia.isFmp4 ? browserSupportsCodec : muxerSupportsCodec;
+    const audioSupported = !codecs.hasOwnProperty('audio') ? true : supportFunction(codecs.audio);
+    const videoSupported = !codecs.hasOwnProperty('video') ? true : supportFunction(codecs.video);
+
+    // make sure both audio and video codecs are supported, if
+    if (!audioSupported || !videoSupported) {
+      const message = `${mainStartingMedia.isFmp4 ? 'browser' : 'muxer'}` +
+        ` does not support ${mainStartingMedia.isFmp4 ? 'fmp4' : 'ts'} codec(s): ` +
+        `${!audioSupported ? codecs.audio : ''} ${!videoSupported ? codecs.video : ''}`;
+
+      this.blacklistCurrentPlaylist({
+        playlist: media,
+        message,
+        internal: true
+      }, Infinity);
+      return;
+    }
+
     if (!codecs.video && !codecs.audio) {
       const error = 'Failed to create SourceBuffers. No compatible SourceBuffer ' +
         'configuration for the variant stream:' + media.resolvedUri;
@@ -1348,18 +1311,15 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return;
     }
 
-    this.excludeIncompatibleVariants_(media);
-  }
+    const codecString = [codecs.video, codecs.audio].filter(Boolean).join(',');
 
-  /**
-   * Blacklists playlists with codecs that are unsupported by the browser.
-   */
-  excludeUnsupportedVariants_() {
-    this.master().playlists.forEach(variant => {
-      if (variant.attributes.CODECS && !browserSupportsCodec(variant.attributes.CODECS)) {
-        variant.excludeUntil = Infinity;
-      }
-    });
+    // TODO:
+    // blacklisting incompatible renditions will have to change
+    // once we add support for `changeType` on source buffers.
+    // We will have to not blacklist any rendition until we try to
+    // switch to it and learn that it is incompatible and if it is compatible
+    // we `changeType` on the sourceBuffer.
+    this.excludeIncompatibleVariants_(codecString);
   }
 
   /**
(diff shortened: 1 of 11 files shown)


src/master-playlist-controller.js
```diff
@@ -273,9 +273,26 @@ export class MasterPlaylistController extends videojs.EventTarget {
         // playlist's `segments` list is already available, a media playlist won't be
         // requested, and loadedplaylist won't fire again, so the playlist handler must be
         // called on its own here.
-        if (this.sourceType_ === 'vhs-json' && this.initialMedia_.segments) {
-          this.handleUpdatedMediaPlaylist(this.initialMedia_);
+        if (this.sourceType_ !== 'vhs-json' && !this.initialMedia_.segments) {
+          return;
         }
+        updatedPlaylist = this.initialMedia_;
```
brandonocasey: We have to check muxer support for codecs, and this code was just doing the same thing as the code below.

src/master-playlist-controller.js
```diff
@@ -872,9 +889,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // Select a new playlist
     const nextPlaylist = this.selectPlaylist();
 
-    videojs.log.warn('Problem encountered with the current playlist.' +
+    if (!error.noWarn) {
```
brandonocasey: We have to blacklist playlists that are not supported by codec. I don't think this would be a valid warning to log.

src/master-playlist-controller.js
```diff
@@ -1261,10 +1280,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
    */
   excludeUnsupportedVariants_() {
     this.master().playlists.forEach(variant => {
-      if (variant.attributes.CODECS &&
-          window.MediaSource &&
-          window.MediaSource.isTypeSupported &&
-          !window.MediaSource.isTypeSupported(`video/mp4; codecs="${mapLegacyAvcCodecs(variant.attributes.CODECS)}"`)) {
+      if (variant.attributes.CODECS && !browserSupportsCodec(variant.attributes.CODECS)) {
```
brandonocasey: Use the `browserSupportsCodec` function from vhs utils.

src/master-playlist-controller.js
```diff
@@ -1286,34 +1302,36 @@ export class MasterPlaylistController extends videojs.EventTarget {
    */
   excludeIncompatibleVariants_(media) {
     let codecCount = 2;
-    let videoCodec = null;
-    let codecs;
+    let codecs = {video: {}};
 
     if (media.attributes.CODECS) {
       codecs = parseCodecs(media.attributes.CODECS);
```
brandonocasey: parseCodec has a slightly different return value now.

src/master-playlist-controller.js
```diff
@@ -1286,34 +1302,36 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      // the video codec cannot change
+      if (variantCodecs.video && codecs.video &&
+        variantCodecs.video.type !== codecs.video.type) {
         variant.excludeUntil = Infinity;
       }
 
-      // if h.264 is specified on the current playlist, some flavor of
-      // it must be specified on all compatible variants
-      if (variantCodecs.videoCodec !== videoCodec) {
+      // the audio codec cannot change
```
brandonocasey: previously we ignored the audio codec all together, seemed wrong to me.

src/source-updater.js
```diff
@@ -225,15 +226,19 @@ export default class SourceUpdater extends videojs.EventTarget {
     }
 
     if (codecs.audio) {
-      this.audioBuffer = this.mediaSource.addSourceBuffer(`audio/mp4;codecs="${codecs.audio}"`);
+      const mime = getMimeForCodec(codecs.audio);
```
brandonocasey: This function should automatically get the video mime type merged with codes. This should allow us to support whatever the browser does.

src/util/codecs.js
```diff
@@ -74,23 +74,28 @@ export const codecsForPlaylist = function(master, media) {
...
   // HLS with multiple-audio tracks must always get an audio codec.
   // Put another way, there is no way to have a video-only multiple-audio HLS!
-  if (isMaat(master, media) && !codecInfo.audioProfile) {
+  if (isMaat(master, media) && !codecInfo.audioCodec) {
     if (!isMuxed(master, media)) {
       // It is possible for codecs to be specified on the audio media group playlist but
       // not on the rendition playlist. This is mostly the case for DASH, where audio and
       // video are always separate (and separately specified).
-      codecInfo.audioProfile = audioProfileFromDefault(master, mediaAttributes.AUDIO);
+      const defaultCodecs = codecsFromDefault(master, mediaAttributes.AUDIO);
```
brandonocasey: this function uses `parseCodecs` and the function name itself was changed to be more generic.

src/util/codecs.js
```diff
@@ -99,25 +104,3 @@ export const codecsForPlaylist = function(master, media) {
...
 };
-
-/*
- * Check if a codec string refers to an audio codec.
- *
- * @param {String} codec codec string to check
- * @return {Boolean} if this is an audio codec
- * @private
- */
-export const isAudioCodec = function(codec) {
```
brandonocasey: both of these functions were moved to vhs-utils.

src/master-playlist-controller.js
gesinger: I think you're right, but it may be safer to isolate that change to a separate PR. Unless we know that browsers can't switch audio codecs.

brandonocasey: This was one of the failures I encountered on edge. We setup source buffers with `avc` video  and `aac`. Then we try to use a playlist that has `avc` video and `ac-3` audio because we didn't blacklist as we were not changing types.

src/master-playlist-controller.js
```diff
@@ -273,9 +273,26 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+
+      const codecs = updatedPlaylist.attributes.CODECS;
+      const segments = updatedPlaylist.segments || [];
+
+      // For non-fmp4 segments we also need to blacklist based on what the
+      // muxer supports. As non-fmp4 goes through the muxer
+      // but we have to wait for a playlist to download before
+      // we can know if it is fmp4 or not, so we can only really blacklist just
+      // before a playlist would get choosen
+      if (codecs && !segments[0].map && !muxerSupportsCodec(codecs)) {
```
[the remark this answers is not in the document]
brandonocasey: I wonder if we should just leave this as is for now. I think that scenario will only effect mixed ts/fmp4 playlists, right? 

Reasons:
Doing this now will make this pull request much more complex and we already rely on very simplistic blacklisting anyway.

For instance this is what I think we would have to do:
We will have to add a handler to `mediaSegmentRequest` that is called with enough segment data to determine the type of segment we are dealing with. Then we will have to check if the browser supports that codec if it is fmp4 or check if the muxer will support that codec if it is ts. Finally we will have to send an event up to the master playlist loader to blacklist this playlist and then abort loading of the remainder of the media segment based on that.

src/master-playlist-controller.js
gesinger: 👍 all good then

src/master-playlist-controller.js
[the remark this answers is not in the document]
gesinger: The scenario can also be for TS playlists alone, see https://github.com/videojs/http-streaming/issues/151

Part of TBA was aiming to address the issue of assuming map meant fmp4. Though that has to be fully tested to see if it was fixed everywhere. 

If we want to go ahead with the PR without doing the more complicated approach, we can add a TODO. I think the case is rare enough that it might not be too big of an issue.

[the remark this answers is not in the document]
brandonocasey: I will add a TODO and work on it after this pull request.

src/master-playlist-controller.js
```diff
@@ -871,14 +877,53 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      `${error.message ? ` ${error.message}` : '' } Switching to another playlist.`);
 
     return this.masterPlaylistLoader_.media(nextPlaylist, isFinalRendition);
   }
 
+  /**
+   * TODO:
+   * Refactor blacklisting by codec so that it happens after
+   * we download enough of a segment to know wether it is fmp4 or ts.
+   * * If it is fmp4 we check if the browser supports the codec and blacklist
```
gesinger: Should the TODO above be below this bulleted set of comments?

src/master-playlist-controller.js
```diff
@@ -871,14 +877,53 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+   * * if it is ts we check if the muxer supports the codec and blacklist
+   *   when it does not, as ts segments will be transmuxed.
+   *
+   * We should probably do this by adding a segmentType handler to mediaSegmentRequest
+   * that calls back with the current segment type and potentially the codec it uses.
+   * Then we check the appropriate support matrix to determine if the playlist should
+   * be blacklisted.
+   */
+  /**
+   * Segments that are not mp4 are not transmuxed, therfore, playlists that
```
gesinger: therefore*

src/master-playlist-controller.js
```diff
@@ -871,14 +877,53 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+   * contain segments which mux.js does not support should be blacklisted. This can only
+   * happen at time of playlist selection, as for some playlists, that is the only time
+   * when the container format for the segments is known.
+   *
+   * @return {boolean}
+   *         Wether the blacklist happened or not
```
gesinger: Whether*

src/master-playlist-controller.js
```diff
@@ -871,14 +877,53 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+   * that calls back with the current segment type and potentially the codec it uses.
+   * Then we check the appropriate support matrix to determine if the playlist should
+   * be blacklisted.
+   */
+  /**
+   * Segments that are not mp4 are not transmuxed, therfore, playlists that
+   * contain segments which mux.js does not support should be blacklisted. This can only
+   * happen at time of playlist selection, as for some playlists, that is the only time
+   * when the container format for the segments is known.
+   *
```
gesinger: I think this is missing `@param`

src/master-playlist-controller.js
gesinger: Actually, maybe it is best to put it right over where we check for `.map`, and mention that we use that check as a shorthand until the larger change is made (but that it may impact TS playlists with EXT-X-MAP tags)

src/master-playlist-controller.js
```diff
@@ -833,9 +831,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return;
     }
 
-    const isFinalRendition =
-      this.masterPlaylistLoader_.master.playlists.filter(isEnabled).length === 1;
     const playlists = this.masterPlaylistLoader_.master.playlists;
+    const enabledPlaylists = playlists.filter(isEnabled);
+    const isFinalRendition = enabledPlaylists.length === 1 && enabledPlaylists[0] === currentPlaylist;
```
brandonocasey, line 836: This function was probably broken for two playlist situations where one playlist gets blacklisted and the other doesn't. This is because we did not check if the last enabled playlist is the playlist we are about to blacklist. Example

You have two playlists `a` and `b` . `a` is enabled, now we are going to blacklist `a` and enable `b`. `enabledPlaylists` will contain only `b`. This means that `a` which is being blacklisted is **NOT** the final rendition.

src/master-playlist-controller.js
```diff
@@ -1224,6 +1222,25 @@ export class MasterPlaylistController extends videojs.EventTarget {
       }
     }
 
+    // fmp4 relies on browser support, while ts relies on muxer support
```
brandonocasey, line 1225: Originally I was doing this by playlist but after looking at the tests and testing more videos it was entirely inadequate. We have to check for support when we get all `trackinfo` just before we try to create source buffers.

src/media-segment-request.js
```diff
@@ -367,6 +367,7 @@ const handleSegmentBytes = ({
     const {tracks} = segment.map;
 
     const trackInfo = {
+      isFmp4: true,
```
brandonocasey, line 370: This is how we determine if we are working with fmp4 or ts all the way up in master playlist loader.

src/segment-loader.js
```diff
@@ -598,10 +598,10 @@ export default class SegmentLoader extends videojs.EventTarget {
     if (!newPlaylist) {
       return;
     }
-
     const oldPlaylist = this.playlist_;
     const segmentInfo = this.pendingSegment_;
 
+    this.startingMedia_ = void 0;
```
brandonocasey, line 604: On playlist change we have to reset starting media, as the previous media may have been blacklisted and caused this change, so it's startingMedia_ was invalid.

src/master-playlist-controller.js
```diff
@@ -1325,38 +1339,44 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-      let variantCodecs = {
-        codecCount: 2,
-        videoCodec: null
-      };
+    this.master().playlists.forEach((variant) => {
+      // skip variants that are already blacklisted forever
+      if (variant.excludeUntil === Infinity) {
+        return;
+      }
+      // TODO: should we really assume codecs?
```
gesinger, line 1351: Is this asking if we should assume that there are two codecs in a playlist without codec info?

src/master-playlist-controller.js
```diff
@@ -1325,38 +1339,44 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-      // it must be specified on all compatible variants
-      if (variantCodecs.videoCodec !== videoCodec) {
+      // the audio codec cannot change
+      if (variantCodecs.audio && codecs.audio &&
+        variantCodecs.audio.type.toLowerCase() !== codecs.audio.type.toLowerCase()) {
         variant.excludeUntil = Infinity;
       }
+
+      if (variant.excludeUntil === Infinity) {
+        this.logger_(`blacklisting ${variant.id} as variant codecs ${variant.attributes.CODECS} are incompatible with ${codecString}`);
```
gesinger, line 1378: This might read a bit funny if there is no CODECS attribute. Maybe we should have a special case log for that?

test/videojs-http-streaming.test.js
```diff
@@ -1446,6 +1454,206 @@ QUnit.test('does not blacklist compatible AAC codec strings', function(assert) {
...
+    // selected playlist
+    '#EXT-X-STREAM-INF:BANDWIDTH=1,CODECS="avc1.4d400d,mp4a.40.2"\n' +
+    'media.m3u8\n' +
+    // compatible with selected playlist
+    '#EXT-X-STREAM-INF:BANDWIDTH=1,CODECS="avc1.4d400d,mp4a.40.2"\n' +
+    'media1.m3u8\n' +
+    // incompatible by audio codec difference
+    '#EXT-X-STREAM-INF:BANDWIDTH=1,CODECS="avc1.4d400d,ac-3"\n' +
+    'media2.m3u8\n' +
+    // incompable by video codec difference
```
gesinger, line 1478: incompatible

test/videojs-http-streaming.test.js
```diff
@@ -1446,6 +1454,206 @@ QUnit.test('does not blacklist compatible AAC codec strings', function(assert) {
...
+
+  openMediaSource(this.player, this.clock);
+
+  const playlistString =
+    '#EXTM3U\n' +
+    // selected playlist
+    '#EXT-X-STREAM-INF:BANDWIDTH=1,CODECS="avc1.4d400d,mp4a.40.2"\n' +
+    'media.m3u8\n' +
+    // compatible with selected playlist
+    '#EXT-X-STREAM-INF:BANDWIDTH=1,CODECS="avc1.4d400d,mp4a.40.2"\n' +
```
gesinger, line 1473: Might be good to have another compatible codec that isn't exact

src/master-playlist-controller.js
```diff
@@ -911,10 +909,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // Select a new playlist
     const nextPlaylist = this.selectPlaylist();
+    const logFn = error.internal ? this.logger_ : videojs.log.warn;
 
-    videojs.log.warn('Problem encountered with the current playlist.' +
-                     (error.message ? ' ' + error.message : '') +
-                     ' Switching to another playlist.');
+    logFn(`${(error.internal ? 'Internal problem' : 'Problem')} encountered with the current playlist.` +
+      `${error.message ? ` ${error.message}` : '' } Switching to another playlist.`);
```
gesinger, line 915: I think we may want a period after the message.

src/master-playlist-controller.js
```diff
@@ -1265,6 +1263,28 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
 
+    // fmp4 relies on browser support, while ts relies on muxer support
+    const supportFunction = mainStartingMedia.isFmp4 ? browserSupportsCodec : muxerSupportsCodec;
+    const audioSupported = !codecs.hasOwnProperty('audio') ? true : supportFunction(codecs.audio);
+    const videoSupported = !codecs.hasOwnProperty('video') ? true : supportFunction(codecs.video);
+
+    // make sure both audio and video codecs are supported, if
+    if (!audioSupported || !videoSupported) {
+      const message = `${mainStartingMedia.isFmp4 ? 'browser' : 'muxer'}` +
+        ` does not support ${mainStartingMedia.isFmp4 ? 'fmp4' : 'ts'} codec(s): ` +
```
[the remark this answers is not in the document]
brandonocasey: I will remove that.

src/master-playlist-controller.js
brandonocasey: yeah

src/master-playlist-controller.js
brandonocasey: yeah

src/master-playlist-controller.js
```diff
@@ -1265,6 +1263,28 @@ export class MasterPlaylistController extends videojs.EventTarget {
       }
     }
 
+    // fmp4 relies on browser support, while ts relies on muxer support
+    const supportFunction = mainStartingMedia.isFmp4 ? browserSupportsCodec : muxerSupportsCodec;
+    const audioSupported = !codecs.hasOwnProperty('audio') ? true : supportFunction(codecs.audio);
```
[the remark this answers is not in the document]
brandonocasey: This part of the code works a bit weird. We don't want to check if we support a codec if we don't have one in the `codecs` object. If we do have it we use the return value of `supports` to see if it can actually be played. The suggested change would mean that any codec existence would mean that we support it.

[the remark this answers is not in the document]
gkatsev: Ah, yeah, it's not equivalent but then I'm still a bit confused on what the original code does has audio not being present would make `audioSupported` set to `true`.

```js
var o = { audio: 1}
> !o.hasOwnProperty('audio') ? true : 1
< 1
> o.hasOwnProperty('audio') || 1
< true
> !o.hasOwnProperty('video') ? true : 1
< true
> o.hasOwnProperty('video') || 1
< 1
```

src/master-playlist-controller.js
gesinger: Is it possible for us to run into this case?

src/master-playlist-controller.js
gesinger: We may want to be careful in those cases. I'm skeptical of any manifests that present wrong information or don't follow the spec. We can support them, but the support will come at a cost of possible wasted bandwidth for playlists that follow the spec.

src/master-playlist-controller.js
```diff
@@ -1265,6 +1263,28 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    const videoSupported = !codecs.hasOwnProperty('video') ? true : supportFunction(codecs.video);
+
+    // make sure both audio and video codecs are supported, if
+    if (!audioSupported || !videoSupported) {
```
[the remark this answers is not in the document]
gesinger: Ah, nevermind, the `audioSupported` line takes into account audio codecs being missing. I was thinking the case of video only, but it seems good 👍

test/videojs-http-streaming.test.js
```diff
@@ -1475,15 +1475,18 @@ QUnit.test('blacklists incompatible playlists by codec', function(assert) {
     // incompatible by audio codec difference
     '#EXT-X-STREAM-INF:BANDWIDTH=1,CODECS="avc1.4d400d,ac-3"\n' +
     'media2.m3u8\n' +
-    // incompable by video codec difference
+    // incompatble by video codec difference
```
gesinger, line 1478: incompatible

src/master-playlist-controller.js
```diff
@@ -1348,9 +1351,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
       if (variant.excludeUntil === Infinity) {
         return;
       }
-      // TODO: should we really assume codecs?
+      // TODO: should we really assume codecs
```
gesinger, line 1354: It may be worth expanding on this comment. Maybe something like.

```
TODO decide whether two codecs should be assumed
Right now, for playlists that don't specify codecs, VHS assumes that there are two (one for audio and one for video). Although this is often the case, this may lead to broken behavior if the playlist only has one codec. It may be better in the future to decide at time of segment download how many tracks there are and determine the proper codecs. This will come at a cost of potentially more bandwidth, but will be a more robust approach than the assumption here.

File: test/master-playlist-controller.test.js [hunk 28]
```diff
@@ -1308,6 +1332,85 @@ QUnit.test('blacklists switching from video-only playlists to video+audio', func
...
+  openMediaSource(this.player, this.clock);
+
+  this.player.tech_.hls.bandwidth = 1;
+
+  // master
+  this.requests.shift()
+    .respond(
+      200, null,
+      '#EXTM3U\n' +
+             '#EXT-X-STREAM-INF:BANDWIDTH=1,CODECS="avc1.4d400d,mp4a.40.5"\n' +
```
gesinger, line 1353: Indentation

src/master-playlist-controller.js
gkatsev: maybe, the ones I've seen have generally worked fine once we added the audio aac codec into the manifest.

src/master-playlist-controller.js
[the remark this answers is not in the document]
brandonocasey: Yeah `audioSupportedOrNoAudio` would have been a better name

brandonocasey:
The only other thing was the `blacklistCurrentPlayilst` confusion but that is document in #800

gkatsev:
Cool, I'll go through the PR tomorrow and make issues and what not from the TODOs.

PR #770, opened 2020-03-16
test(vtt-segment-loader): reset error log count

gkatsev:
The previous commit passed tests in PR but failing in the branch https://travis-ci.org/github/videojs/http-streaming/builds/663197151?utm_source=github_status&utm_medium=notification. This is likely because an error log count wasn't being reset in this test.


gkatsev:
This isn't the issue.

PR #772, opened 2020-03-16
fix: Remove current sourceBuffer audio on disable

brandonocasey:
## Description
When switching to alternative audio and turning off audio on the main segment loader we are not removing the original audio data from the main segment loader. This means that we start adding new audio onto the old audio causing all kinds of issues.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -354,6 +354,9 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.audioDisabled_ = !enable;
     if (enable) {
       this.appendInitSegment_.audio = true;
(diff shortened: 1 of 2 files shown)


gesinger:
Actually, one case we may want to look into. If we're switching renditions from a muxed rendition to a demuxed rendition, this would remove only audio from the buffer, at which point we might have a gap.

brandonocasey:
Added a test. If you get some time to test this pull request in that scenario that would be great. From my testing that scenario seems to work.

gesinger:
Created some content to test and wasn't able to find any problems. 👍

Issue #378, opened 2019-01-16
HLS video is not resuming after network issues


nicklive:
## Description

Hello, I have an issue with video.js+http-streaming, that I can consistently reproduce on basicly all hls streams. Video.js is not resuming live stream after network issues (slow connection and resuming to normal).

Here [reduced test case](http://jsfiddle.net/h1rfL8mk/1/) but for reproducing you need to throttle your connection with chrome devtools (please see _Steps to reproduce_).

## Steps to reproduce

1. Open [link](http://jsfiddle.net/h1rfL8mk/1/) 
2. Wait for loading example and video start
3. Open chrome devtools set throttling for _100 kb/s_ 
4. Wait about 50 seconds (or a bit more)
5. After set throttling to _online_

### Expected
Logically after **Steps to reproduce** video should resume back to live

### Actual
video.js+http-streaming try to load infinity times one segment and aborting it, video is not resuming

### Error output
Error output contains a lot of

```
video.js:142 VIDEOJS: ERROR: DOMException: Failed to set the 'duration' property on 'MediaSource': The 'updating' attribute is true on one or more of this MediaSource's SourceBuffers.
    at HtmlMediaSource.addSeekableRange_ (https://unpkg.com/video.js@7.4.1/dist/video.js:49141:44)
    at addSeekableRange (https://unpkg.com/video.js@7.4.1/dist/video.js:55039:36)
    at PlaylistLoader.<anonymous> (https://unpkg.com/video.js@7.4.1/dist/video.js:55054:15)
    at PlaylistLoader.data.dispatcher (https://unpkg.com/video.js@7.4.1/dist/video.js:1818:33)
    at trigger (https://unpkg.com/video.js@7.4.1/dist/video.js:1954:27)
    at PlaylistLoader.EventTarget.trigger (https://unpkg.com/video.js@7.4.1/dist/video.js:2432:5)
    at PlaylistLoader.haveMetadata (https://unpkg.com/video.js@7.4.1/dist/video.js:39151:14)
    at https://unpkg.com/video.js@7.4.1/dist/video.js:39089:17
    at Object.callback (https://unpkg.com/video.js@7.4.1/dist/video.js:40086:9)
    at cbOnce (https://unpkg.com/video.js@7.4.1/dist/video.js:7220:17)
```

## Additional Information
There are 3 segments in samlpe hls playlist, but I tried 30 segments in playlist and problem repeats but after more delay (2 min) in 4 step.

### videojs-http-streaming version
[...]

PR #777, opened 2020-03-18
Revert "fix: Use middleware and a wrapped function for seeking instead of relying on unreliable 'seeking' events (#161)"

gkatsev:
This reverts commit 6c68761b5d06d951d7ed20cbd5bb63f324ad2c47.

This helps address #378 and  videojs/video.js#6444.

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -29,8 +29,6 @@ import {
   comparePlaylistResolution
 } from './playlist-selectors.js';
 import { version } from '../package.json';
-// import needed to register middleware
-import './middleware-set-current-time';
 import { isAudioCodec, isVideoCodec, parseContentType } from './util/codecs';
 
 const Hls = {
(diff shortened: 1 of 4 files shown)


src/videojs-http-streaming.js
```diff
@@ -508,12 +509,9 @@ class HlsHandler extends Component {
     this.options_.tech = this.tech_;
     this.options_.externHls = Hls;
     this.options_.sourceType = simpleTypeFromSourceType(type);
-    // Whenever we seek internally, we should update both the tech and call our own
-    // setCurrentTime function. This is needed because "seeking" events aren't always
-    // reliable. External seeks (via the player object) are handled via middleware.
+    // Whenever we seek internally, we should update the tech
     this.options_.seekTo = (time) => {
```
gesinger: Is `seekTo` still used anywhere?

gkatsev: Yes, it's currently used inside of seekToProgramTime and related functions.

gesinger: Should we then just leave all of the other uses of `seekTo`?

gkatsev: We could, we specifically need `seekTo` for seekToProgramTime because STPT doesn't have access to anything else like the tech or what not.

gesinger: Might be worth keeping it consistent then. Also hides the use of tech a little more.

PR #778, opened 2020-03-20
fix: audio and video syncing for demuxed content with discontinuities

gesinger:
In order to properly sync audio and video for demuxed content when there are discontinuities, one of the streams must follow the timing of the other. For VHS, video was chosen (historically) to be the source of truth for media timing, and audio was manipulated to fit the media timing.

In order to properly fit audio to the video timeline, audio must know the timing information of video before it is appended, and video must wait for audio to cross a discontinuity before it crosses, so as not to set the timestamp offset or other associated timing information for audio before audio is done with appends along its current timeline.

As such, a TimelineChangeController object was added, in order to facilitate discontinuity crossings where audio and video streams each wait until the other is ready to cross before they make the move over to a new timeline. This allows both to remain in sync, as they cross a discontinuity together, and keep appropriate video timing information for each timeline.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -158,6 +158,81 @@ export const timestampOffsetForSegment = ({
  * Returns whether or not the loader should wait for a timeline change from the timeline
  * change controller before processing the segment.
  *
+ * Primary timing in VHS goes by video. This is different from most media players, as
+ * audio is more often used as the primary timing source. For the foreseeable future, VHS
+ * will continue to use video as the primary timing source, due to the current logic and
+ * expectations built around it.
+
+ * Since the timing follows video, in order to maintain sync, the video loader is
+ * responsible for setting both audio and video source buffer timestamp offsets.
+ *
+ * Setting different values for audio and video source buffers could lead to
+ * desyncing. The following examples demonstrate some of the situations where this
+ * distinction is important. Note that all of these cases involve demuxed content. When
+ * content is muxed, the audio and video are packaged together, therefore syncing
+ * separate media playlists is not an issue.
+ *
+ * CASE 1: Audio prepares to load a new timeline before video:
+ *
+ * Timeline:       0                 1
+ * Audio Segments: 0 1 2 3 4 5 DISCO 6 7 8 9
+ * Audio Loader:                     ^
+ * Video Segments: 0 1 2 3 4 5 DISCO 6 7 8 9
+ * Video Loader              ^
+ *
+ * In the above example, the audio loader is preparing to load the 6th segment, the first
+ * after a discontinuity, while the video loader is still loading the 5th segment, before
+ * the discontinuity.
+ *
+ * If the audio loader goes ahead and loads and appends the 6th segment before the video
+ * loader crosses the discontinuity, then when appended, the 6th audio segment will use
+ * the timestamp offset from timeline 0. This will likely lead to desyncing. In addition,
+ * the audio loader must provide the audioAppendStart value to trim the content in the
+ * transmuxer, and that value relies on the audio timestamp offset. Since the audio
+ * timestamp offset is set by the video (main) loader, the audio loader shouldn't load the
+ * segment until that value is provided.
+ *
+ * CASE 2: Video prepares to load a new timeline before audio:
+ *
+ * Timeline:       0                 1
+ * Audio Segments: 0 1 2 3 4 5 DISCO 6 7 8 9
+ * Audio Loader:             ^
+ * Video Segments: 0 1 2 3 4 5 DISCO 6 7 8 9
+ * Video Loader                      ^
+ *
+ * In the above example, the video loader is preparing to load the 6th segment, the first
+ * after a discontinuity, while the audio loader is still loading the 5th segment, before
+ * the discontinuity.
+ *
+ * If the video loader goes ahead and loads and appends the 6th segment, then once the
+ * segment is loaded and processed, both the video and audio timestamp offsets will be
+ * set, since video is used as the primary timing source. This is to ensure content lines
+ * up appropriately, as any modifications to the video timing are reflected by audio when
+ * the video loader sets the audio and video timestamp offsets to the same value. However,
+ * setting the timestamp offset for audio before audio has had a chance to change
+ * timelines will likely lead to desyncing, as the audio loader will append segment 5 with
+ * a timestamp intended to apply to segments from timeline 1 rather than timeline 0.
+ *
+ * CASE 3: When seeking, audio prepares to load a new timeline before video
+ *
+ * Timeline:       0                 1
+ * Audio Segments: 0 1 2 3 4 5 DISCO 6 7 8 9
+ * Audio Loader:           ^
+ * Video Segments: 0 1 2 3 4 5 DISCO 6 7 8 9
+ * Video Loader            ^
+ *
+ * In the above example, both audio and video loaders are loading segments from timeline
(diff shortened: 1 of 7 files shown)


src/segment-loader.js
```diff
@@ -154,6 +154,159 @@ export const timestampOffsetForSegment = ({
...
+    // * main hasn't had a timeline change yet (thus has not loaded its first segment)
+    // * main hasn't yet changed to the timeline audio is looking to load
+    return !lastMainTimelineChange || lastMainTimelineChange.to !== segmentTimeline;
+  }
+
+  // The main loader only needs to wait for timeline changes if there's demuxed audio.
+  // Otherwise, there's nothing to wait for, since audio would be muxed into the main
+  // loader's segments (or the content is audio/video only and handled by the main
+  // loader).
+  if (loaderType === 'main' && audioDisabled) {
```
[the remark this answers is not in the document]
gesinger: It will not. It's only disabled when an alt audio track is enabled.

src/segment-loader.js
```diff
@@ -154,6 +154,159 @@ export const timestampOffsetForSegment = ({
...
+  currentTimeline,
+  segmentTimeline,
+  loaderType,
+  audioDisabled
+}) => {
+  if (currentTimeline === segmentTimeline) {
+    return false;
+  }
+
+  if (loaderType === 'audio') {
```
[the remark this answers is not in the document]
gesinger: Yup

src/segment-loader.js
```diff
@@ -154,6 +154,159 @@ export const timestampOffsetForSegment = ({
   return buffered.length ? buffered.end(buffered.length - 1) : startOfSegment;
 };
 
+/**
+ * Returns whether or not the loader should wait for a timeline change from the timeline
```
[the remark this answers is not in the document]
gesinger: I can definitely put it in a doc and add a link, but my only concern is anything not in the code will quickly become out-of-date/not considered. Still open to it though.

test/segment-loader.test.js
```diff
@@ -215,6 +216,173 @@ QUnit.test('returns value when overrideCheck is true', function(assert) {
...
+QUnit.module('shouldWaitForTimelineChange');
+
+QUnit.test('should not wait if timelines are the same', function(assert) {
+  assert.notOk(
+    shouldWaitForTimelineChange({ currentTimeline: 1, segmentTimeline: 1 }),
+    'should not wait'
+  );
+});
+
+QUnit.test('should wait if audio and no main timeline change', function(assert) {
```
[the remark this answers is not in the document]
gesinger: I was trying to compress the text, but you're right, clarity over line length 👍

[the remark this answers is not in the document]
gesinger: I think yours may also be shorter too, so thanks!

src/segment-loader.js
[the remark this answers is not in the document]
brandonocasey: we decided to leave it here

src/segment-loader.js
```diff
@@ -154,6 +154,159 @@ export const timestampOffsetForSegment = ({
...
+ * @param {('main'|'audio')} loaderType
+ *        The loader type
+ * @param {boolean} audioDisabled
+ *        Whether the audio is disabled for the loader
+ *
+ * @return {boolean}
+ *         Whether the loader should wait for a timeline change from the timeline change
+ *         controller before processing the segment
+ */
+export const shouldWaitForTimelineChange = ({
```
gkatsev: does this account for something like the audio missing a segment? It would previously cause lots of av sync issues and something that squarebracket was working on. If not, we should remember to circle back to this case later on, we don't need to solve it in this PR.
I think the example would end up looking kind of like:
```

 * Timeline:       0                 1
 * Audio Segments:     2 3 4 5 DISCO 6 7 8 9
 * Audio Loader:           ^
 * Video Segments: 0 1 2 3 4 5 DISCO 6 7 8 9
 * Video Loader            ^
```

gkatsev: should this method just be on the `timelineChangeController`?
so you could do something like
```
const shouldWait = timelineChangeController.shouldWaitForTimelineChange({
  currentTimeline,
  segmentTimeline,
  loaderType: 'main'
});
```

src/segment-loader.js
```diff
@@ -154,6 +154,159 @@ export const timestampOffsetForSegment = ({
...
+ * not be aligned perfectly, even for matching segment indexes.
+ *
+ * @param {Object} timelinechangeController
+ * @param {number} currentTimeline
+ *        The timeline currently being followed by the loader
+ * @param {number} segmentTimeline
+ *        The timeline of the segment being loaded
+ * @param {('main'|'audio')} loaderType
+ *        The loader type
+ * @param {boolean} audioDisabled
```
gkatsev: so this will likely only be set in the case of the main loaded when an alternate audio track is enabled?
Might be worth adding it here.

src/segment-loader.js
gesinger: Do we have any example content for it? And would this just be for the live case where the playlists are updating out-of-sync?

src/segment-loader.js
gesinger: Yeah, and good call, will add a note about it.

src/segment-loader.js
gkatsev: I don't think we have one but I think squarebracket said we can take a demuxed stream and delete the first segment from the audio playlist but leave the video playlist alone to demonstrate the issue.

src/timeline-change-controller.js
```diff
@@ -0,0 +1,46 @@
...
+export default class TimelineChangeController extends videojs.EventTarget {
+  constructor() {
+    super();
+
+    this.pendingTimelineChanges_ = {};
+    this.lastTimelineChanges_ = {};
+  }
+
+  clearPendingTimelineChange(type) {
+    delete this.pendingTimelineChanges_[type];
```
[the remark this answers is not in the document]
gesinger: Definitely can. Any reason for the preference, or is it just more standard?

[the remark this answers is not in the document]
gkatsev: Historically, delete causes performance issues since it changes the internal data structure where-as setting to null doesn't. I do wonder how much of an issue it still is. Might better more the more used a thing is.

src/segment-loader.js
gesinger: Tested, and TBA alone fixed the issue.

gesinger: I thought about it a bit, and I think that it would be best to keep it out of the TimelineChangeController. Right now, TimelineChangeController is a pretty simple class without any logic in it around loaders or streaming decisions. That logic is primarily contained within the segment loaders, so I figure it is best to keep it contained there. Eventually it might be best to even move this function out of the segment loader module, along with some other logic that ties loaders together, into a separate module, to start shrinking the module itself, but for now I figure it's best to prevent any of the logic leaking.

It also makes testing easier to have it as function not attached to a class.

But I'm also open to hearing arguments for moving it.

gkatsev: Main argument for moving it is that this method directly deals with the timeline change and is given the controller. We already have the controller in the call site and the controller already has all the other info, so, it seems reasonable to ask it for whether it should change.
However, it is true, without that method there, the controller does stay simple. Though, the complexity doesn't go away it's just elsewhere.
I'm OK with just leaving as is.

PR #779, opened 2020-03-23
fix: null check return value of selectPlaylist

gkatsev:


src/master-playlist-controller.js
```diff
@@ -429,15 +429,16 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-      if (!currentPlaylist.endList ||
-          // For the same reason as LIVE, we ignore the low water line when the VOD
-          // duration is below the max potential low water line
-          this.duration() < Config.MAX_BUFFER_LOW_WATER_LINE ||
-          // we want to switch down to lower resolutions quickly to continue playback, but
-          nextPlaylist.attributes.BANDWIDTH < currentPlaylist.attributes.BANDWIDTH ||
-          // ensure we have some buffer before we switch up to prevent us running out of
-          // buffer while loading a higher rendition.
-          forwardBuffer >= bufferLowWaterLine) {
+      if (nextPlaylist &&
```
brandonocasey: perhaps this should just be a "return" if statement:

```js
if (!nextPlaylist || !nextPlaylist.attributes || !currentPlaylist || currentPlaylist.attributes) {
  return
}

Comment (gkatsev) [reply src/master-playlist-controller.js]: could be, though, I think we'd still want to trigger the `bandwidthupdate` event and it depends on what order we want to do that event and the media call.
If order doesn't matter, we can move the trigger above the check.

File: src/master-playlist-controller.js [hunk 2]
```diff
@@ -429,15 +429,16 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+          (!currentPlaylist.endList ||
```
brandonocasey: It might even be better to switch this giant if statement up into `return` if statements to make it more readable/understandable.

```js
// If the playlist is live, then we want to not take low water line into account.
// This is because in LIVE, the player plays 3 segments from the end of the	      
// playlist, and if `BUFFER_LOW_WATER_LINE` is greater than the duration available	 
// in those segments, a viewer will never experience a rendition up-switch.
if (currentPlaylist.endList) {
  return;
}

// For the same reason as LIVE, we ignore the low water line when the VOD
// duration is below the max potential low water line
if (this.duration() > Config.MAX_BUFFER_LOW_WATER_LINE) {
  return;
}

// we want to switch down to lower resolutions quickly to continue playback, but
if (nextPlaylist.attributes.BANDWIDTH > currentPlaylist.attributes.BANDWIDTH) {
  return;
}

// ensure we have some buffer before we switch up to prevent us running out of
if (forwardBuffer < bufferLowWaterLine) {
  return
}

Comment (gesinger) [reply src/master-playlist-controller.js]: I like this, but I think @gkatsev makes a good point about re-triggering the `bandwidthupdate` event in the same order as before, just so as not to inadvertently break anyone.

I think a good way of doing this would be to create a new function that just takes in the parameters to check for if we `shouldSwitchToMedia`, then we can have the best of both worlds and a much easier function to test.

File: src/master-playlist-controller.js (hunk 1 above)
Comment (gesinger) [inline src/master-playlist-controller.js]: It may be worth having a message being logged somewhere (maybe even as a warning) in this case. Will make debugging easier if we ever run into it.

Comment (brandonocasey) [reply src/master-playlist-controller.js]: that is true, I missed the `bandwidthupdate` triggering at the bottom, perhaps we should just have two event handlers and the second one should trigger the event, seems like we will always trigger it currently.

**Comment by gkatsev** [comment]
note https://github.com/videojs/http-streaming/pull/779/commits/70ad31499707b6798fa29d1d3a46018026362ecc

File: test/master-playlist-controller.test.js [hunk 3]
```diff
@@ -1283,6 +1283,31 @@ QUnit.test('selects a playlist after main/combined segment downloads', function(
...
+
+  // "downloaded" a segment
+  this.masterPlaylistController.mainSegmentLoader_.trigger('bandwidthupdate');
+  assert.strictEqual(calls, 2, 'selects after the initial segment');
+
+  // and another
+  this.masterPlaylistController.mainSegmentLoader_.trigger('bandwidthupdate');
+  assert.strictEqual(calls, 3, 'selects after additional segments');
+  // verify stats
+  assert.equal(this.player.tech_.hls.stats.bandwidth, 4194304, 'default bandwidth');
```
gesinger: I think we can remove this.

test/master-playlist-controller.test.js
```diff
@@ -1283,6 +1283,31 @@ QUnit.test('selects a playlist after main/combined segment downloads', function(
...
+  this.masterPlaylistController.selectPlaylist = () => {
+    calls++;
+    return null;
+  };
+  this.masterPlaylistController.mediaSource.trigger('sourceopen');
+
+  // master
+  this.standardXHRResponse(this.requests.shift());
+  // media
+  this.standardXHRResponse(this.requests.shift());
```
gesinger: Is there a different initial playlist selection going on? I'd imagine the overriden `selectPlaylist` would return `null` for the media.

gkatsev: So, it's loading the media manifest. I think what's happening is that since selectPlaylist is returning null, we don't end up switching to the media that's provided by selectPlaylist and we continue with the current select item.
Looking at the test above `selects a playlist after main/combined segment downloads`, changing it to return `playlist[1]` instead of `playlist[0]`, when inspecting this request, we get show it loading media1.m3u8, but in this case we still end up loading media.m3u8, which I believe is the expected behavior.

test/master-playlist-controller.test.js
gkatsev: I kept it because it's literally a copy and paste from the above test but I guess we can remove it.

test/master-playlist-controller.test.js
gesinger: Yeah, forgot about this line: https://github.com/videojs/http-streaming/blob/1.x/src/playlist-loader.js#L577

PR #781, opened 2020-03-24
fix: null check return value of selectPlaylist

gkatsev:
This is #779 but against master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,7 @@ test/dist/
 
 # this file is generated and shouldn't be checked in
 src/decrypter-worker.worker.js
+src/mse/transmuxer-worker.worker.js
 src/transmuxer-worker.worker.js
 
 .eslintcache
-.yo-rc.json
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -38,7 +38,7 @@ test/dist/
 
 # this file is generated and shouldn't be checked in
 src/decrypter-worker.worker.js
+src/mse/transmuxer-worker.worker.js
 src/transmuxer-worker.worker.js
 
 .eslintcache
-.yo-rc.json
```
gkatsev: this is already above

.gitignore
```diff
@@ -38,7 +38,7 @@ test/dist/
 
 # this file is generated and shouldn't be checked in
 src/decrypter-worker.worker.js
+src/mse/transmuxer-worker.worker.js
```
gkatsev: so that when we switch between 1.x and master branch, the worker file is always ignored.

gkatsev:
I'm not quite sure what the travis integration is being funky. One of the builds passed, re-running the other one.

gkatsev:
Both travis builds are passing. Merging.

PR #806, opened 2020-04-09
fix: update media references in DASH playlist loader to ensure they are from master

gesinger:
Previously, some media references in the DASH playlist loader weren't properly updated,
meaning that they were references from the old master. This broke blacklisting, where the
old playlist was blacklisted, but the playlist selector, using the newer playlist
references, could select the same playlist (but a new reference).

## Requirements Checklist
  - [ ] Docs/guides updated


brandonocasey:
could be go to push the test change first to show the failure and then push the commit to fix it. Code looks good though.

PR #808, opened 2020-04-09
chore(demo): demo type examples, ie 11 support, hit enter in type/url

brandonocasey:
1. I found myself hitting enter in the url box or type box a lot. 
2. I often look for the exact type of dash/hls to copy paste.
3. ie 11 only half works as it does not support `URLSearchParams` I added a polyfill of sorts.

### DIFF
--- a/index.html
+++ b/index.html
@@ -71,7 +71,7 @@
   <h3>Load a URL</h3>
   <label>Url:</label>
   <input id=url type=url>
-  <label>Type: (uses url extension if blank)</label>
+  <label>Type: (uses url extension if blank) usually application/x-mpegURL or application/dash+xml)</label>
   <input id=type type=text>
   <button id=load-url type=button>Load</button>
   <h3>Load a Source</h3>
--- a/index.html
+++ b/index.html
@@ -71,7 +71,7 @@
   <h3>Load a URL</h3>
   <label>Url:</label>
   <input id=url type=url>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -71,7 +71,7 @@ <h3>Options</h3>
   <h3>Load a URL</h3>
   <label>Url:</label>
   <input id=url type=url>
-  <label>Type: (uses url extension if blank)</label>
+  <label>Type: (uses url extension if blank) usually application/x-mpegURL or application/dash+xml)</label>
```
gesinger, line 74:
```suggestion
  <label>Type: (uses url extension if blank, usually application/x-mpegURL or application/dash+xml)</label>
```

scripts/index-demo-page.js
```diff
@@ -71,21 +71,34 @@
     window.history.replaceState({}, 'vhs demo', query);
   };
 
-  var loadState = function() {
-    var params = {get: function(param) {
-      return null;
-    }};
+  window.URLSearchParams = window.URLSearchParams || function(searchString) {
```
gesinger, line 74: Is this a polyfill from somewhere, or just crafted?

brandonocasey: All of it vaguely comes from stackoverflow, but it's not really a copy paste since the stackoverflow answer was wrong and overly complex.

PR #814, opened 2020-04-13
chore(demo): new sources, generate selection list from a JSON file

gkatsev:
Move sources into a JSON file that we can use to populate the list. Eventually, we could have the features for each item be in the JSON file and then have a better UI on the demo page that would allow filtering by those features. Though, currently, the features are only used for the live and low-latency streams to move them into the live and low latency opt groups.

### DIFF
--- a/scripts/index-demo-page.js
+++ b/scripts/index-demo-page.js
@@ -10,18 +10,18 @@
   // get the sources list squared away
   var xhr = new window.XMLHttpRequest();
 
-  xhr.addEventListener('load', () => {
+  xhr.addEventListener('load', function() {
     var sources = JSON.parse(xhr.responseText);
 
-    sources.forEach((source) => {
+    sources.forEach(function(source) {
       const option = document.createElement('option');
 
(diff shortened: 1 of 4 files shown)


scripts/index-demo-page.js
```diff
@@ -1,6 +1,34 @@
...
+/* eslint-disable vars-on-top, no-var, object-shorthand, no-console */
 (function(window) {
+
+  const hlsOptGroup = document.querySelector('[label="hls"]');
+  const dashOptGroup = document.querySelector('[label="dash"]');
+  const liveOptGroup = document.querySelector('[label="live"]');
+  const llliveOptGroup = document.querySelector('[label="low latency live"]');
+
+  // get the sources list squared away
+  window.fetch('./scripts/sources.json')
```
gesinger, line 11: Will we lose IE11 support for the test page with this?

gkatsev: Maybe, probably.

PR #811, opened 2020-04-10
fix: use codecsForPlaylist to support dash

brandonocasey:
## Description
Now that less codecs are being blacklisted it became apparent that we need to use `codecsForPlaylist` rather than `attributes.CODECS` in order to support dashes weird codec problems.

PR #815, opened 2020-04-13
feat: detect webm segments and blacklist playlist instead of throwing an exception

gesinger:
We may want to classify this as a fix, but WebM detection is new, so kept as a feature for now.

Requires: https://github.com/videojs/http-streaming/pull/811 (will update once that PR is merged)

## Requirements Checklist
  - [X] Docs/guides updated


gkatsev:
We don't support webm because all our code does mp4 probing which won't work on webm? And we can't easily just  put the webm into the source buffer without doing similar probing in the webm?

gesinger:
Yeah. We'll have to start probing webm to be consistent, or create an alternate code path to do direct appends without considering timing info.

test/util/codecs.test.js
```diff
@@ -372,3 +375,27 @@ QUnit.test('parses codecs regardless of codec order', function(assert) {
     'parses audio first'
   );
 });
+
+QUnit.module('isLikelyWebmData');
```
brandonocasey, line 379: could be good to add tests using actual mp4/webm init segments here too.

PR #818, opened 2020-04-15
fix: Prevent switching to alternative audio causing infinite buffering

brandonocasey:
## Description
Switching to alternative audio right now sometimes causes us to infinite buffer and the duration of the video doubles. This seems to be a fix for it. Tested using `Bipbop - Muxed TS with 1 alt Audio, 5 captions`


brandonocasey:
closing this as it doesn't do anything

PR #820, opened 2020-04-15
chore: update muxjs and allow 0 index for demo selectedIndex

brandonocasey:
## Description
1. Update `mux.js` to the most recent version. 
1. `selectedIndex` can be `0` so we need to check if `selectIndex` is a number before setting it to `-1` or else if the first dropdown source is selected, ie 0, we will set the dropdown to -1.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,12 +14,12 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.8.6",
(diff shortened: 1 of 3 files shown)


gkatsev:
Test is failing because of Video.js 7.8 fix/issue?

brandonocasey:
yeah, will revert the package-lock update and only update mux.js

gkatsev:
I'll make a Video.js patch release as well.

PR #824, opened 2020-04-16
fix: use xhr to check container type of sidx and blacklist if anything but mp4

brandonocasey:
A pull request to work-around webm content that we don't currently support. This pull request checks the container format of the sidx segment and blacklists the playlist if it is not mp4.
 
Depends on https://github.com/videojs/vhs-utils/pull/12


src/util/codecs.js
```diff
@@ -121,3 +122,52 @@ export const isLikelyWebmData = (bytes) => {
     (bytes[2] & 0xFF) === 0xDF &&
     (bytes[3] & 0xFF) === 0xA3;
 };
+
+export const isLikelyMp4Data = (bytes) => {
```
brandonocasey, line 126: Long ago in tba this function was called `isLikelyFmp4Data` but I had to change it to the current iteration because it also matched regular mp4 data. Now I am bringing it back as I think it fits our use case here.

brandonocasey:
This pull request needs:
1. Test fixes
2. cleanup the sidx request portion of the code as its very hard to follow

A Future pull request could:
Use this new function to determine the container type without downloading the whole segment in mediaSegmentRequest. That way we can blacklist after only downloading 8 bytes plus a playlist rather than an entire segment, or even two segments.

src/segment-loader.js
```diff
@@ -992,7 +992,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       this.sourceUpdater_.removeAudio(start, end, removeFinished);
     }
 
-    if (this.loaderType_ === 'main' && this.startingMedia_.hasVideo) {
+    if (this.loaderType_ === 'main' && this.startingMedia_ && this.startingMedia_.hasVideo) {
```
brandonocasey, line 995: protect against undefined `this.startingMedia_` as that is the state it starts in.

src/util/codecs.js
```diff
@@ -105,19 +105,3 @@ export const isLikelyFmp4Data = (bytes) => {
...
- * Check to see if the bytes are part of a WebM media file.
- *
- * @param {Uint8Array} bytes
- *        The starting bytes of the file, generally from the map, or media initialization
- *        section, of the file.
- * @return {Boolean}
- *         Whether the bytes likely come from a WebM media file.
- * @see https://en.wikipedia.org/wiki/List_of_file_signatures
- */
-export const isLikelyWebmData = (bytes) => {
```
brandonocasey, line 118: moved to vhs-utils.

test/dash-playlist-loader.test.js
```diff
@@ -19,6 +19,7 @@ import {
 import '../src/videojs-http-streaming';
 import testDataManifests from 'create-test-data!manifests';
 import { sidx as sidxResponse } from 'create-test-data!segments';
+import {mp4VideoInit as mp4VideoInitSegment} from 'create-test-data!segments';
```
brandonocasey, line 22: Now we have to supply a legit init segment for sidx requests or else we blacklist the playlist. Had to update tests to do that.

src/dash-playlist-loader.js
```diff
@@ -393,7 +418,8 @@ export default class DashPlaylistLoader extends EventTarget {
       sidxInfo: playlist.sidx
     };
 
-    this.request = requestSidx_(
+    this.request = requestSidx_.call(
+      this, startingState,
```
gkatsev, line 422:
```suggestion
      this,
      startingState,
```

src/dash-playlist-loader.js
```diff
@@ -306,7 +345,7 @@ export default class DashPlaylistLoader extends EventTarget {
         return doneFn(master, null);
       }
 
-      const bytes = new Uint8Array(request.response);
+      const bytes = toUint8(request.response);
```
brandonocasey, line 348: Since we potentially re-use bytes and they can already be a uin8 at this point, we use the helper from vhs-utils which only creates a new uint8 if needed.

src/xhr.js
```diff
@@ -16,6 +16,45 @@ const {
   mergeOptions
 } = videojs;
 
+const callbackWrapper = function(request, error, response, callback) {
```
brandonocasey, line 19: This code is unchanged, I just moved it into its own function so that container-request can use it.

test/dash-playlist-loader.test.js
```diff
@@ -471,7 +472,7 @@ QUnit.test('requestSidx_: creates an XHR request for a sidx range', function(ass
   const sidxInfo = {
     resolvedUri: 'sidx.mp4',
     byterange: {
-      offset: 0,
+      offset: 10,
```
brandonocasey, line 475: We don't want to re-use bytes in this test (from the container request) so I set the offset so that it doesn't happen.

test/test-helpers.js
```diff
@@ -372,31 +373,29 @@ export const standardXHRResponse = function(request, data) {
...
-    response = data.buffer;
-  } else if (!data) {
-    // A placeholder is used for some old tests. This may be a good target to clean up in
-    // the future.
-    response = new Uint8Array(1024).buffer;
-  } else {
-    response = data;
+  if (isTypedArray(response)) {
+    // a string for container requests or a buffer for non-container requests
+    response = isContainerRequest ? bytesToString(response) : response.buffer;
```
brandonocasey, line 398: container requests are `byte` strings, so that we can work-around the fact that xhr does not add to `request.response` until the request is done, it only adds to `request.responseText`, which throws an error when accesed if the `responseType` is `arraybuffer` like normal.

test/util/codecs.test.js
```diff
@@ -379,30 +374,3 @@ QUnit.test('parses codecs regardless of codec order', function(assert) {
     'parses audio first'
   );
 });
-
-QUnit.module('isLikelyWebmData');
```
brandonocasey, line 383: All this moved to vhs-utils

src/dash-playlist-loader.js
```diff
@@ -204,7 +206,44 @@ export const requestSidx_ = (sidxRange, playlist, xhr, options, finishProcessing
...
+      return finishProcessingFn(err, request);
+    }
+
+    if (!container || container !== 'mp4') {
+      this.error = {
+        status: request.status,
+        message: `Unsupported ${container || 'unknown'} container type for sidx segment at URL: ${request.uri}`,
+        response: request.response,
+        playlist,
+        internal: true,
```
brandonocasey, line 220: Are we ok with this being an internal error? Seems like a feature we lack rather than an actual error to me, but if we want to warn in the console that could be fine.

src/media-segment-request.js
```diff
@@ -168,11 +168,14 @@ const handleInitSegmentResponse =
...
-  if (isLikelyWebmData(segment.map.bytes)) {
+  const type = detectContainerForBytes(segment.map.bytes);
+
+  if (type !== 'mp4') {
     return finishProcessingFn({
       status: request.status,
-      message: 'Found unsupported WebM initialization segment at URL: ' + request.uri,
+      message: `Found unsupported ${type || 'unknown'} container for initialization segment at URL: ${request.uri}`,
       code: REQUEST_ERRORS.FAILURE,
+      internal: true,
```
brandonocasey, line 178: Are we ok with this being an internal error? Seems like a feature we lack rather than an actual error to me, but if we want to warn in the console that could be fine.

src/dash-playlist-loader.js
gkatsev: I think we can just leave it as is for now, we probably want to rethink how errors are handled to be able to better allow for people to handle things as they want.

src/media-segment-request.js
```diff
@@ -168,11 +168,14 @@ const handleInitSegmentResponse =
 
   segment.map.bytes = new Uint8Array(request.response);
 
-  if (isLikelyWebmData(segment.map.bytes)) {
+  const type = detectContainerForBytes(segment.map.bytes);
+
+  if (type !== 'mp4') {
```
gesinger, line 173: A transport stream can still have a media initialization section, thus, an EXT-X-MAP tag: https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-3.2

src/segment-loader.js
```diff
@@ -2317,8 +2317,8 @@ export default class SegmentLoader extends videojs.EventTarget {
     // Although transmuxing is done, appends may not yet be finished. Throw a marker
     // on each queue this loader is responsible for to ensure that the appends are
     // complete.
-    const waitForVideo = this.loaderType_ === 'main' && this.startingMedia_.hasVideo;
-    const waitForAudio = !this.audioDisabled_ && this.startingMedia_.hasAudio;
+    const waitForVideo = this.loaderType_ === 'main' && this.startingMedia_ && this.startingMedia_.hasVideo;
```
gesinger, line 2320: Is there a case where we're trying to perform this operation (and the check in updateTimingInfoEnd) without having startingMedia?

src/dash-playlist-loader.js
```diff
@@ -204,7 +206,44 @@ export const requestSidx_ = (sidxRange, playlist, xhr, options, finishProcessing
...
   });
 
-  return xhr(sidxRequestOptions, finishProcessingFn);
+  return containerRequest(sidxInfo.uri, xhr, (err, request, container, bytes) => {
+    if (err) {
+      return finishProcessingFn(err, request);
+    }
+
+    if (!container || container !== 'mp4') {
+      this.error = {
```
gesinger, line 215: Because this is a standalone function, even though it can be `call`ed with a context, I think it would be preferable to pass in an error callback handler or use some other system than to use `this`.

src/util/container-request.js
```diff
@@ -0,0 +1,51 @@
...
+    request.abort();
+    const type = detectContainerForBytes(bytes);
+
+    cb(null, request, type, bytes);
+  };
+
+  const options = {
+    uri,
+    beforeSend(request) {
+      request.overrideMimeType('text/plain; charset=x-user-defined');
```
gesinger, line 41: Would be worth adding the comment used in the partial downloads elsewhere here to clarify why we're doing the override.

src/util/container-request.js
```diff
@@ -0,0 +1,51 @@
...
+    if (!id3Offset) {
+      id3Offset = getId3Offset(bytes);
+    }
+
+    // not enough data to determine type yet
+    if (bytes.length < id3Offset + 2) {
+      return;
+    }
+    request.abort();
+    const type = detectContainerForBytes(bytes);
```
gesinger, line 33: If there's no id3 offset, and bytes.length is 3, would that be enough bytes to determine the container in all cases?

src/util/container-request.js
```diff
@@ -0,0 +1,51 @@
...
+  let id3Offset;
+
+  const progressListener = function(error, request) {
+    if (error) {
+      return cb(error, request, '', bytes);
+    }
+    const currentLength = bytes && bytes.length || 0;
+    const contentPart = request.responseText.substring(currentLength, request.responseText.length);
+    const newBytes = toUint8(stringToBytes(contentPart, true));
+    const temp = new Uint8Array(currentLength + newBytes.length);
```
gesinger, line 16: Might be good to use a more specific name, or specify why a temp var is being used.

test/test-helpers.js
```diff
@@ -372,31 +373,29 @@ export const standardXHRResponse = function(request, data) {
     manifestName = request.url;
   }
 
+  const isContainerRequest = request.mimeTypeOverride === 'text/plain; charset=x-user-defined';
```
gesinger, line 376: Since this could be for partial data downloads in any case, we may want to rename to something like `isPartialRequest`

src/media-segment-request.js
brandonocasey: Won't that mean that the map is mp4 though?

src/segment-loader.js
brandonocasey: I only really see it when we are going to player error in some way, or when we try to load something that we shouldn't be, like loading a webm in the muxer.

src/util/container-request.js
brandonocasey: Only mp3 and aac should have id3 offset and they only need two bytes after the id3 data to determine the type

src/media-segment-request.js
gesinger: The map would be a transport stream too, just with PAT and PMT

src/segment-loader.js
gesinger: Would it be better to fail quickly with an exception there? Since it can otherwise lead to undefined behavior, and will help to surface issues where we might be trying to do something that we shouldn't be doing.

src/media-segment-request.js
brandonocasey: The problem with that though is that we won't be able to parse it as we assume that the map is mp4 below. I will add a TODO comment specifying that we should also handle ts segment maps.

src/util/container-request.js
brandonocasey: added a comment

src/segment-loader.js
brandonocasey: OK I triggered an error for this case, probably a good idea to blacklist such a playlist and try other in this case too.

src/segment-loader.js
```diff
@@ -2314,11 +2314,19 @@ export default class SegmentLoader extends videojs.EventTarget {
   }
 
   waitForAppendsToComplete_(segmentInfo) {
+    if (!this.startingMedia_) {
+      this.error({
+        message: 'No starting media returned, we dont support this media format.',
```
gesinger, line 2319:
```suggestion
        message: 'No starting media returned, likely due to an unsupported media format.',
```

src/dash-playlist-loader.js
```diff
@@ -189,7 +191,7 @@ export const filterChangedSidxMappings = (masterXml, srcUrl, clientOffset, oldSi
 };
 
 // exported for testing
-export const requestSidx_ = (sidxRange, playlist, xhr, options, finishProcessingFn) => {
+export const requestSidx_ = function(loader, sidxRange, playlist, xhr, options, finishProcessingFn) {
```
gesinger, line 194:
```suggestion
export const requestSidx_ = (loader, sidxRange, playlist, xhr, options, finishProcessingFn) => {
```

src/media-segment-request.js
```diff
@@ -168,11 +168,16 @@ const handleInitSegmentResponse =
 
   segment.map.bytes = new Uint8Array(request.response);
 
-  if (isLikelyWebmData(segment.map.bytes)) {
+  const type = detectContainerForBytes(segment.map.bytes);
+
+  // TODO: We should also handle ts init segments here, but we
+  // only know how to parse ts maps at the moment
```
gesinger, line 174:
```suggestion
  // only know how to parse mp4 maps at the moment
```

src/segment-loader.js
```diff
@@ -2440,7 +2448,7 @@ export default class SegmentLoader extends videojs.EventTarget {
   updateTimingInfoEnd_(segmentInfo) {
     segmentInfo.timingInfo = segmentInfo.timingInfo || {};
     const useVideoTimingInfo =
-      this.loaderType_ === 'main' && this.startingMedia_.hasVideo;
+      this.loaderType_ === 'main' && this.startingMedia_ && this.startingMedia_.hasVideo;
```
gesinger, line 2451: Since the check for startingMedia happens earlier, I think we can change this back

```suggestion
      this.loaderType_ === 'main' && this.startingMedia_.hasVideo;
```

src/util/container-request.js
gesinger: I'm probably missing something, but I guess what I'm asking is, if we're trying to detect the container of a video mp4 segment, and we need 4 bytes for that detection, but we got 3 bytes on progress, will this return no detected type and the playlist get blacklisted?

test/test-helpers.js
```diff
@@ -372,31 +373,29 @@ export const standardXHRResponse = function(request, data) {
...
-  if (isTypedBuffer) {
-    response = data.buffer;
-  } else if (!data) {
-    // A placeholder is used for some old tests. This may be a good target to clean up in
-    // the future.
-    response = new Uint8Array(1024).buffer;
-  } else {
-    response = data;
+  if (isTypedArray(response)) {
+    // a string for container requests or a buffer for non-container requests
```
gesinger, line 397:
```suggestion
    // a string for partial requests or a buffer for non-partial requests
```

src/util/container-request.js
```diff
@@ -0,0 +1,58 @@
...
+    beforeSend(request) {
+      // this forces the browser to pass the bytes to us unprocessed
+      request.overrideMimeType('text/plain; charset=x-user-defined');
+      request.addEventListener('progress', function(event) {
+        return callbackWrapper(request, null, {statusCode: request.status}, progressListener);
+      });
+    }
+  };
+
+  return xhr(options, () => {});
```
gesinger, line 55: Will this prevent normal (non progress) XHR completions? (If so, we may have to look into IE11, or worry about other browser compatibilities for progress: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequestEventTarget/onprogress#Browser_compatibility .)

src/util/container-request.js
brandonocasey: I think you may have a point her, going to refactor this a bit.

src/util/container-request.js
brandonocasey: ie 11 on windows 8 and 10 does support progress events, but I updated the code anyway to run the final callback if the progress requests did not finish the request.

brandonocasey:
I think I have all the comments handled here at one point, but I realized that we want another helper in vhs-utils, so we will have to get that in before we can finish this pr up. https://github.com/videojs/vhs-utils/pull/13

src/util/container-request.js
```diff
@@ -1,58 +1,85 @@
...
+  return;
+};
+
 const containerRequest = (uri, xhr, cb) => {
-  let bytes;
+  let bytes = [];
   let id3Offset;
+  let finished = false;
+
+  const callback = function(err, req, type, _bytes) {
```
gesinger, line 19: Minor, but I find it a bit hard to read when there are two params that are essentially the same name, but have different spellings (e.g., `cb` and `callback`). Since this one serves the purpose of finishing the request, maybe we can name it something along the lines of `endRequestAndCallback`.

gkatsev: `requestCompletedCallback` here
and on line 14 `cb` to `containerRequestCallback`

brandonocasey: Done!

PR #828, opened 2020-04-22
chore(demo): add more sources, add keySystems and put DRMed sources in an optgroup

gkatsev:


brandonocasey:
Seems to work, would it be helpful to have another section. Maybe we should have sections for dash drm, live drm, hls drm etc. I also think that the keysystems textarea needs to be a lot bigger

gkatsev:
Yeah, there's more work here, like adding a couple of sources. Originally opened it because it didn't work locally and needed help verifying it should work. It should start working for me once #824 is merged.

PR #829, opened 2020-04-27
feat: Use playback watcher to watch when progress isn't being made

brandonocasey:
## Description
Tested using the issue from #823. I also tested that this code does not blacklist playlists with large content on slow network connections. IE throttle network to 10mb/s in chrome and select the 4k boat source.

## Testing
1. `npm install mux.js@5.5.3` and use `https://d2zihajmogu5jn.cloudfront.net/bipbop-advanced/bipbop_16x9_variant.m3u8`. Skip to the middle of the video and change audio tracks. Before we would infinite download audio, now we blacklist until everything stops.
2. comment out `this.shouldSaveSegmentTimingInfo = false` in `vtt-segment-loader.js`. Run the axiom live stream, not that the text track is excluded after some time as it isn't making any progress.
3. Test with a playlist that has gaps to verify that nothing is excluded. 
4. Tested with a playlist that has one vtt for the entire video and make sure nothing is excluded.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -143,6 +143,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
       timeout: null
     };
 
+    this.on('error', this.pauseLoading);
+
     this.mediaTypes_ = createMediaTypes();
 
     this.mediaSource = new window.MediaSource();
@@ -881,10 +883,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
     const enabledPlaylists = playlists.filter(isEnabled);
     const isFinalRendition = enabledPlaylists.length === 1 && enabledPlaylists[0] === currentPlaylist;
 
-    if (playlists.length === 1) {
-      // Never blacklisting this playlist because it's the only playlist
-      videojs.log.warn('Problem encountered with the current ' +
-                       'playlist. Trying again since it is the only playlist.');
+    // Don't blacklist the only playlist unless it was blacklisted
+    // forever
+    if (playlists.length === 1 && blacklistDuration !== Infinity) {
+      videojs.log.warn(`Problem encountered with playlist ${currentPlaylist.id}. ` +
+                       'Trying again since it is the only playlist.');
 
       this.tech_.trigger('retryplaylist');
       return this.masterPlaylistLoader_.load(isFinalRendition);
@@ -895,17 +898,30 @@ export class MasterPlaylistController extends videojs.EventTarget {
       // it, instead of erring the player or retrying this playlist, clear out the current
       // blacklist. This allows other playlists to be attempted in case any have been
       // fixed.
-      videojs.log.warn('Removing all playlists from the blacklist because the last ' +
-                       'rendition is about to be blacklisted.');
+      let reincluded = false;
+
       playlists.forEach((playlist) => {
-        if (playlist.excludeUntil !== Infinity) {
+        // skip current playlist which is about to be blacklisted
+        if (playlist === currentPlaylist) {
+          return;
+        }
+        const excludeUntil = playlist.excludeUntil;
+
+        // a playlist cannot be reincluded if it wasn't excluded to begin with.
+        if (typeof excludeUntil !== 'undefined' && excludeUntil !== Infinity) {
+          reincluded = true;
           delete playlist.excludeUntil;
         }
       });
-      // Technically we are retrying a playlist, in that we are simply retrying a previous
-      // playlist. This is needed for users relying on the retryplaylist event to catch a
-      // case where the player might be stuck and looping through "dead" playlists.
-      this.tech_.trigger('retryplaylist');
+
+      if (reincluded) {
+        videojs.log.warn('Removing other playlists from the exclusion list because the last ' +
+                         'rendition is about to be excluded.');
+        // Technically we are retrying a playlist, in that we are simply retrying a previous
+        // playlist. This is needed for users relying on the retryplaylist event to catch a
+        // case where the player might be stuck and looping through "dead" playlists.
+        this.tech_.trigger('retryplaylist');
+      }
     }
 
     // Blacklist this playlist
@@ -915,19 +931,26 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // Select a new playlist
     const nextPlaylist = this.selectPlaylist();
+
+    if (!nextPlaylist) {
+      this.error = 'Playback cannot continue. No available working or supported playlists.';
+      this.trigger('error');
+      return;
+    }
     const logFn = error.internal ? this.logger_ : videojs.log.warn;
     const errorMessage = error.message ? (' ' + error.message) : '';
 
-    logFn(`${(error.internal ? 'Internal problem' : 'Problem')} encountered with the current playlist.` +
-      `${errorMessage} Switching to another playlist.`);
+    logFn(`${(error.internal ? 'Internal problem' : 'Problem')} encountered with playlist ${currentPlaylist.id}.` +
+      `${errorMessage} Switching to playlist ${nextPlaylist.id}.`);
 
     return this.masterPlaylistLoader_.media(nextPlaylist, isFinalRendition);
   }
 
   /**
-   * Pause all segment loaders
+   * Pause all segment/playlist loaders
    */
   pauseLoading() {
+    // pause all segment loaders
     this.mainSegmentLoader_.pause();
     if (this.mediaTypes_.AUDIO.activePlaylistLoader) {
       this.audioSegmentLoader_.pause();
@@ -935,6 +958,14 @@ export class MasterPlaylistController extends videojs.EventTarget {
     if (this.mediaTypes_.SUBTITLES.activePlaylistLoader) {
       this.subtitleSegmentLoader_.pause();
     }
+
+    // pause all playlist loaders
+    this.masterPlaylistLoader_.pause();
+    Object.keys(this.mediaTypes_).forEach((type) => {
+      if (this.mediaTypes_[type].activePlaylistLoader) {
+        this.mediaTypes_[type].activePlaylistLoader.pause();
+      }
+    });
   }
 
   /**
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -233,7 +233,7 @@ export default class PlaybackWatcher {
     // TODO: should we exclude audio tracks rather than main tracks
     // when type is audio?
     mpc.blacklistCurrentPlaylist({
-      message: `Infinite ${type} segment downloading detected.`
+      message: `Excessive ${type} segment downloading detected.`
     }, Infinity);
   }
 
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -11,6 +11,7 @@
 import window from 'global/window';
 import * as Ranges from './ranges';
 import logger from './util/logger';
+import videojs from 'video.js';
 
 // Set of events that reset the playback-watcher time check logic and clear the timeout
 const timerCancelEvents = [
@@ -71,6 +72,7 @@ export default class PlaybackWatcher {
    * @param {Object} options an object that includes the tech and settings
    */
   constructor(options) {
+    this.masterPlaylistController_ = options.masterPlaylistController;
     this.tech_ = options.tech;
     this.seekable = options.seekable;
     this.seekTo = options.seekTo;
@@ -90,6 +92,22 @@ export default class PlaybackWatcher {
     const cancelTimerHandler = () => this.cancelTimer_();
     const fixesBadSeeksHandler = () => this.fixesBadSeeks_();
 
+    const mpc = this.masterPlaylistController_;
+
+    const loaderTypes = ['main', 'subtitle', 'audio'];
+    const loaderChecks = {};
+
+    loaderTypes.forEach((type) => {
+      loaderChecks[type] = {
+        reset: () => this.resetSegmentDownloads_(type),
+        updateend: () => this.checkSegmentDownloads_(type)
+      };
+
+      mpc[`${type}SegmentLoader_`].on('updateend', loaderChecks[type].updateend);
+      mpc[`${type}SegmentLoader_`].on('playlist-update', loaderChecks[type].reset);
+      this.tech_.on(['seeked', 'seeking'], loaderChecks[type].reset);
+    });
+
     this.tech_.on('seekablechanged', fixesBadSeeksHandler);
     this.tech_.on('waiting', waitingHandler);
     this.tech_.on(timerCancelEvents, cancelTimerHandler);
@@ -102,6 +120,12 @@ export default class PlaybackWatcher {
       this.tech_.off('waiting', waitingHandler);
       this.tech_.off(timerCancelEvents, cancelTimerHandler);
       this.tech_.off('canplay', canPlayHandler);
+
+      loaderTypes.forEach((type) => {
+        mpc[`${type}SegmentLoader_`].off('updateend', loaderChecks[type].updateend);
+        mpc[`${type}SegmentLoader_`].off('playlist-update', loaderChecks[type].reset);
+        this.tech_.off(['seeked', 'seeking'], loaderChecks[type].reset);
+      });
       if (this.checkCurrentTimeTimeout_) {
         window.clearTimeout(this.checkCurrentTimeTimeout_);
       }
@@ -124,6 +148,75 @@ export default class PlaybackWatcher {
     // 42 = 24 fps // 250 is what Webkit uses // FF uses 15
     this.checkCurrentTimeTimeout_ =
       window.setTimeout(this.monitorCurrentTime_.bind(this), 250);
+  }
+
+  /**
+   * Reset stalled download stats for a specific type of loader
+   *
+   * @param {string} type
+   *        Whether to check the audio or main segment loader.
+   */
+  resetSegmentDownloads_(type) {
+    const loader = this.masterPlaylistController_[`${type}SegmentLoader_`];
+
+    if (this[`${type}StalledDownloads_`] > 0) {
+      this.logger_(`reseting stalled downloads for ${type} loader`);
+    }
+    this[`${type}StalledDownloads_`] = 0;
+    this[`${type}Buffered_`] = loader.buffered_();
+  }
+
+  /**
+   * Checks on every main/audio segment `appended` to see
+   * if segment appends are making progress. If they are not
+   * and we are still downloading bytes. We blacklist the playlist.
+   *
+   * @param {string} type
+   *        Whether to check the audio or main segment loader.
+   */
+  checkSegmentDownloads_(type) {
+    const mpc = this.masterPlaylistController_;
+    const loader = mpc[`${type}SegmentLoader_`];
+    const buffered = loader.buffered_();
+    const isBufferedDifferent = Ranges.isRangeDifferent(this[`${type}Buffered_`], buffered);
+
+    this[`${type}Buffered_`] = buffered;
+
+    // if another watcher is going to fix the issue or
+    // the buffered value for this loader changed
+    // appends are working
+    if (isBufferedDifferent) {
+      this.resetSegmentDownloads_(type);
+      return;
+    }
+
+    this[`${type}StalledDownloads_`]++;
+
+    this.logger_(`found stalled download #${this[`${type}StalledDownloads_`]} for ${type} loader`);
+    // We will technically get past this on the fourth bad append
+    // rather than the third. As the first will almost always cause
+    // buffered to change which means that StalledDownloads_ will
+    // not be incremented
+    if (this[`${type}StalledDownloads_`] < 3) {
+      return;
+    }
+
+    this.logger_(`${type} loader download exclusion`);
+    this.resetSegmentDownloads_(type);
+    this.tech_.trigger({type: 'usage', name: `vhs-${type}-download-exclusion`});
+
+    if (type !== 'subtitle') {
+      mpc.blacklistCurrentPlaylist({
+        message: `Infinite ${type} segment downloading detected.`
+      }, Infinity);
+    } else {
+      const track = loader.track();
+      const label = track.label || track.language || 'Unknown';
+
+      videojs.log.warn(`Text track "${label}" is not working correctly. It will be disabled and excluded.`);
+      track.mode = 'disabled';
+      this.tech_.textTracks().removeTrack(track);
+    }
   }
 
   /**
@@ -351,6 +444,8 @@ export default class PlaybackWatcher {
 
       this.logger_(`Stopped at ${currentTime}, setting timer for ${difference}, seeking ` +
         `to ${nextRange.start(0)}`);
+
+      this.cancelTimer_();
 
       this.timer_ = setTimeout(
         this.skipTheGap_.bind(this),
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -103,8 +103,15 @@ export default class PlaybackWatcher {
         updateend: () => this.checkSegmentDownloads_(type)
       };
 
-      mpc[`${type}SegmentLoader_`].on('updateend', loaderChecks[type].updateend);
-      mpc[`${type}SegmentLoader_`].on('playlist-update', loaderChecks[type].reset);
+      mpc[`${type}SegmentLoader_`].on('appendsdone', loaderChecks[type].updateend);
+      // If a rendition switch happens during a playback stall where the buffer
+      // isn't changing we want to reset. We cannot assume that the new rendition
+      // will also be stalled, until after new appends.
+      mpc[`${type}SegmentLoader_`].on('playlistupdate', loaderChecks[type].reset);
+      // Playback stalls should not be detected right after seeking.
+      // This prevents one segment playlists (single vtt or single segment content)
+      // from being detected as stalling. As the buffer will not change in those cases, since
+      // the buffer is the entire video duration.
       this.tech_.on(['seeked', 'seeking'], loaderChecks[type].reset);
     });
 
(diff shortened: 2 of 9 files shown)


src/master-playlist-controller.js
```diff
@@ -875,40 +875,55 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-      // playlist. This is needed for users relying on the retryplaylist event to catch a
-      // case where the player might be stuck and looping through "dead" playlists.
-      this.tech_.trigger('retryplaylist');
-    }
 
-    // Blacklist this playlist
-    currentPlaylist.excludeUntil = Date.now() + (blacklistDuration * 1000);
-    this.tech_.trigger('blacklistplaylist');
-    this.tech_.trigger({type: 'usage', name: 'hls-rendition-blacklisted'});
+      if (unblacklisted) {
```
brandonocasey, line 907: Only log and trigger if a playlist was unblacklisted, otherwise we aren't actually retrying anything.

src/master-playlist-controller.js
```diff
@@ -875,40 +875,55 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+        // playlist. This is needed for users relying on the retryplaylist event to catch a
+        // case where the player might be stuck and looping through "dead" playlists.
+        this.tech_.trigger('retryplaylist');
+      }
+    }
 
     // Select a new playlist
     const nextPlaylist = this.selectPlaylist();
+
+    if (!nextPlaylist) {
```
brandonocasey, line 920: It's possible for `selectPlaylist` to return nothing. We don't handle that case now because we make sure to always unblacklist before coming here. Now we don't do that because some playlists that cause playback to stall **should** stay blacklisted.

src/master-playlist-controller.js
```diff
@@ -143,6 +143,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
       timeout: null
     };
 
+    this.on('error', () => {
```
brandonocasey, line 146: We can take this change out if it seems too big, but IMO we should be pausing all loading when we cause a player error.

src/master-playlist-controller.js
```diff
@@ -875,10 +880,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
     const enabledPlaylists = playlists.filter(isEnabled);
     const isFinalRendition = enabledPlaylists.length === 1 && enabledPlaylists[0] === currentPlaylist;
 
-    if (playlists.length === 1) {
-      // Never blacklisting this playlist because it's the only playlist
-      videojs.log.warn('Problem encountered with the current ' +
-                       'playlist. Trying again since it is the only playlist.');
+    // Don't blacklist the only playlist unless it was blacklisted
+    // forever
+    if (playlists.length === 1 && blacklistDuration !== Infinity) {
```
brandonocasey, line 885: I don't think we should retry a playlist, even if it is the only one, when it is blacklisted with `Infinity`. Without this change, media with a single playlist cannot be stopped from looping downloads.

src/master-playlist-controller.js
```diff
@@ -889,17 +895,29 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+
       playlists.forEach((playlist) => {
-        if (playlist.excludeUntil !== Infinity) {
+        // skip current playlist which is about to be blacklisted
+        if (playlist === currentPlaylist) {
+          return;
+        }
+        const excludeUntil = playlist.excludeUntil;
+
+        if (typeof excludeUntil !== 'undefined' && excludeUntil !== Infinity) {
```
brandonocasey, line 907: a playlist is not unblacklisted if it wasn't blacklisted to begin with.

src/master-playlist-controller.js
```diff
@@ -889,17 +895,29 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
       // blacklist. This allows other playlists to be attempted in case any have been
       // fixed.
-      videojs.log.warn('Removing all playlists from the blacklist because the last ' +
-                       'rendition is about to be blacklisted.');
+      let unblacklisted = false;
+
       playlists.forEach((playlist) => {
-        if (playlist.excludeUntil !== Infinity) {
+        // skip current playlist which is about to be blacklisted
+        if (playlist === currentPlaylist) {
```
brandonocasey, line 902: The current playlist is about to be blacklisted, so it should not cause us to log that **other** playlists are going to be unblacklisted.

src/master-playlist-controller.js
```diff
@@ -909,11 +927,17 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      this.error = 'Playback cannot continue. No playlists are working.';
+      this.trigger('error');
+      return;
+    }
     const logFn = error.internal ? this.logger_ : videojs.log.warn;
     const errorMessage = error.message ? (' ' + error.message) : '';
 
-    logFn(`${(error.internal ? 'Internal problem' : 'Problem')} encountered with the current playlist.` +
-      `${errorMessage} Switching to another playlist.`);
+    logFn(`${(error.internal ? 'Internal problem' : 'Problem')} encountered with playlist ${currentPlaylist.id}.` +
```
brandonocasey, line 939: Updated this log line to be more specific.

src/master-playlist-controller.js
```diff
@@ -931,6 +955,17 @@ export class MasterPlaylistController extends videojs.EventTarget {
     }
   }
 
+  pausePlaylistLoading() {
```
brandonocasey, line 958: Added a function to pause loaders. I wonder if this should just be a part of `pauseLoading`?

src/playback-watcher.js
```diff
@@ -107,18 +118,71 @@ export default class PlaybackWatcher {
...
+    // should fire a `waiting` event in this scenario, but due to browser and tech
+    // inconsistencies. Calling `techWaiting_` here allows us to simulate
+    // responding to a native `waiting` event when the tech fails to emit one.
+
+    if (this.lastRecordedTime === currentTime && atEndBuffer) {
+      if (this.techWaiting_()) {
+        return;
+      }
+
+      const previousBytes = this.stallStartBytes_;
```
brandonocasey, line 134: The biggest part of this change is here. We check if playback is abnormally stalled, if it is we start tracking how many byte transfer changes have happened during this time. If we detect more than 3, ie 3 segments have been downloaded, but not `playing` has occurred. Then we try to blacklist something.

src/playback-watcher.js
```diff
@@ -107,18 +118,71 @@ export default class PlaybackWatcher {
...
-      // should fire a `waiting` event in this scenario, but due to browser and tech
-      // inconsistencies. Calling `techWaiting_` here allows us to simulate
-      // responding to a native `waiting` event when the tech fails to emit one.
-      return this.techWaiting_();
+      this.resetStallHandlers_();
+      this.tech_.trigger({type: 'usage', name: 'hls-infinite-download-blacklist'});
+      const AUDIO = mpc.mediaTypes_.AUDIO;
+
+      // try blacklisting audio first
+      if (AUDIO.activePlaylistLoader && Object.keys(AUDIO.tracks).length > 1) {
```
brandonocasey, line 155: First we blacklist audio, I don't think we have a good way to do this so I had to manually figure it out. We really should think about a more universal `blacklistPlaylist` which includes the ability to blacklist audio playlists at some point though.

src/playback-watcher.js
```diff
@@ -107,18 +118,71 @@ export default class PlaybackWatcher {
...
+          AUDIO.tracks[availableTracksIds[0]].enabled = true;
+          this.tech_.audioTracks().removeTrack(activeTrack);
+          videojs.log.warn(`Problem encountered with playlist ${media.id}. ` +
+            `Playback stalled on audio ${activeTrack.id}. Switching to ${availableTracksIds[0]}.`);
+          return;
+        }
+      }
+
+      // if that didn't work or can't happen blacklist main media
+      const playlist = mpc.media();
```
brandonocasey, line 175: If we cannot blacklist any more audio playlists, and playback is still stalled try to blacklist main media playlists.

brandonocasey:
Almost all the test changes right now are related to the log changes that I made in this pull request. I am going to be adding the following tests soon:
* mpc errors pause all playlist and segment loaders
* stalled playback causes audio playlists to be blacklisted when possible
* stalled playback blacklists audio playlists then main media playlists if needed
* stallled playback blacklists main media playlists if no audio playlists can be blacklisted

src/playback-watcher.js
```diff
@@ -107,18 +118,69 @@ export default class PlaybackWatcher {
...
+    // responding to a native `waiting` event when the tech fails to emit one.
+
+    if (this.lastRecordedTime === currentTime && atEndBuffer) {
+      if (this.techWaiting_()) {
+        return;
+      }
+      const mpc = this.vhs.masterPlaylistController_;
+      const previousBytes = this.stallStartBytes_;
+
+      this.stallStartBytes_ = mpc.mediaBytesTransferred_();
```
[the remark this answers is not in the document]
brandonocasey: yeah

src/master-playlist-controller.js
```diff
@@ -889,17 +895,29 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+
       playlists.forEach((playlist) => {
-        if (playlist.excludeUntil !== Infinity) {
+        // skip current playlist which is about to be blacklisted
+        if (playlist === currentPlaylist) {
+          return;
+        }
+        const excludeUntil = playlist.excludeUntil;
+
+        if (typeof excludeUntil !== 'undefined' && excludeUntil !== Infinity) {
```
gesinger, line 907: Adding your comment as a comment:

```suggestion
        // a playlist is not unblacklisted if it wasn't blacklisted to begin with.
        if (typeof excludeUntil !== 'undefined' && excludeUntil !== Infinity) {
```

src/master-playlist-controller.js
```diff
@@ -909,11 +927,17 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // Select a new playlist
     const nextPlaylist = this.selectPlaylist();
+
+    if (!nextPlaylist) {
+      this.error = 'Playback cannot continue. No playlists are working.';
```
gesinger, line 932: Since technically the playlists can be working but not supported, maybe something like

```suggestion
      this.error = 'Playback cannot continue. No available working or supported playlists.';
```

src/master-playlist-controller.js
gesinger: I think it should. If it is, then we can remove the error listener on MPC and rely on the tech error listener in `videojs-http-streaming.js`.

src/playback-watcher.js
```diff
@@ -107,18 +131,81 @@ export default class PlaybackWatcher {
...
+
+      if (previousBytes && previousBytes !== this.stallStartBytes_) {
+        this.stallByteChanges_++;
+      }
+
+      // if we have yet to be stalled for 3 segment downloads
+      // or mpc has an error, do nothing
+      if (this.stallByteChanges_ < 3 || mpc.error) {
```
gesinger, line 162: Since `mediaBytesTransferred` is a summed loader stat between audio and video loaders, I believe that this will detect infinite buffering in the event that one loader has downloaded 3 segments before the other has downloaded 1, which may be a likely case for small audio segments and large video segments.

src/playback-watcher.js
```diff
@@ -107,18 +131,81 @@ export default class PlaybackWatcher {
...
+        return;
+      }
+
+      this.resetStallHandlers_();
+      this.tech_.trigger({type: 'usage', name: 'hls-infinite-download-exclusion'});
```
gesinger, line 167: It might be worth us no longer using the `hls` prefix. I know it's consistent with prior usage events, but maybe we should start migrating them

src/playback-watcher.js
```diff
@@ -107,18 +131,81 @@ export default class PlaybackWatcher {
...
+      }
+      const mpc = this.vhs.masterPlaylistController_;
+      const previousBytes = this.stallStartBytes_;
+
+      if (!previousBytes) {
+        this.tech_.off('timeupdate', this.checkForStallFix_);
+        this.tech_.on('timeupdate', this.checkForStallFix_);
+      }
+
+      this.stallStartBytes_ = mpc.mediaBytesTransferred_();
```
gesinger, line 154: Although it's a less likely scenario, we may also want to consider the case where the VTT loader is infinitely downloading as well, as that was an issue we recently resolved, but is possible may recur.

src/playback-watcher.js
```diff
@@ -107,18 +131,81 @@ export default class PlaybackWatcher {
...
+      this.resetStallHandlers_();
+      this.tech_.trigger({type: 'usage', name: 'hls-infinite-download-exclusion'});
+
+      const playlist = mpc.media();
+      const playlists = mpc.master().playlists;
+      const enabledPlaylists = playlists.filter(isEnabled);
+      const isFinalRendition = enabledPlaylists.length === 1 && enabledPlaylists[0] === playlist;
+      const AUDIO = mpc.mediaTypes_.AUDIO;
+
+      // if this is the final playlists and there are audio tracks
```
gesinger, line 175:
```suggestion
      // if this is the final playlist and there are audio tracks
```

src/playback-watcher.js
```diff
@@ -107,18 +131,81 @@ export default class PlaybackWatcher {
...
+          .filter((id) => !AUDIO.tracks[id].enabled && !AUDIO.tracks[id].excluded);
+
+        if (availableTracksIds.length) {
+          const activeTrack = AUDIO.activeTrack();
+          const newTrack = AUDIO.tracks[availableTracksIds[0]];
+
+          activeTrack.excluded = true;
+          // if main media has audio we won't have an active loader here yet
+          if (AUDIO.activePlaylistLoader && AUDIO.activePlaylistLoader.media()) {
+            AUDIO.activePlaylistLoader.media().excludeUntil = Infinity;
```
gesinger, line 188: Why is it excluded if it doesn't have audio but is instead just a label?

src/playback-watcher.js
```diff
@@ -107,18 +131,81 @@ export default class PlaybackWatcher {
...
+
+      const playlist = mpc.media();
+      const playlists = mpc.master().playlists;
+      const enabledPlaylists = playlists.filter(isEnabled);
+      const isFinalRendition = enabledPlaylists.length === 1 && enabledPlaylists[0] === playlist;
+      const AUDIO = mpc.mediaTypes_.AUDIO;
+
+      // if this is the final playlists and there are audio tracks
+      // try to exclude those before excluding the final playlist
+      if (isFinalRendition && Object.keys(AUDIO.tracks).length > 1) {
```
gesinger, line 177: Won't this potentially switch languages (e.g., English to Spanish) or audio type (e.g., dialogue to director's commentary)?

src/playback-watcher.js
brandonocasey: I wonder if we should use the specific segment loaders `mediaBytesTransferred` then we know weather to try blacklisting audio or video 🤔

src/playback-watcher.js
brandonocasey: sure lets use `vhs`

src/videojs-http-streaming.js
```diff
@@ -511,7 +511,8 @@ class HlsHandler extends Component {
     this.masterPlaylistController_ = new MasterPlaylistController(this.options_);
     this.playbackWatcher_ = new PlaybackWatcher(videojs.mergeOptions(this.options_, {
       seekable: () => this.seekable(),
-      media: () => this.masterPlaylistController_.media()
+      media: () => this.masterPlaylistController_.media(),
+      vhs: this
```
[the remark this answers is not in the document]
brandonocasey: I'm starting to wonder why we don't just pass vhs to almost everything in the constructor for these objects. Similar to how was pass Player everywhere in video.js, regardless I will change this to masterPlaylistController.

src/playback-watcher.js
brandonocasey: It will potentially do that, perhaps we should only switch within the same language and throw an error otherwise?

src/playback-watcher.js
brandonocasey: I'm not sure I understand what you mean here.

src/playback-watcher.js
brandonocasey: Do you have a link to the pull request and the content that I can use for testing this issue? I agree that we should detect it but it may require something a bit different.

src/playback-watcher.js
gesinger: I'd be wary of switching even within the same language, just because the setup could be different for different users, and the experience could be jarring if the audio suddenly changes to something unexpected.

src/playback-watcher.js
gesinger: PR: https://github.com/videojs/http-streaming/pull/825 , stream was the Axinom live HLS stream

src/master-playlist-controller.js
```diff
@@ -915,26 +931,41 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // Select a new playlist
     const nextPlaylist = this.selectPlaylist();
+
+    if (!nextPlaylist) {
+      this.error = 'Playback cannot continue. No available working or supported playlists.';
```
gkatsev, line 936: outside the scope of this PR but we need to think about localizations from the strings in this project, particularly if they show errors to the user.

src/ranges.js
```diff
@@ -396,3 +396,44 @@ export const timeRangesToArray = (timeRanges) => {
...
+ *        the first buffered object to check
+ *
+ * @param {TimeRange} b
+ *        the second buffered object to check
+ *
+ * @return {Boolean}
+ *         Whether the time range objects differ
+ */
+
+export const isRangeDifferent = function(a, b) {
```
brandonocasey, line 413: I originally had this as a function in `playback-watcher` but it seems generic enough to pull into `ranges`.

src/segment-loader.js
```diff
@@ -833,6 +833,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     }
 
     if (!oldPlaylist || oldPlaylist.uri !== newPlaylist.uri) {
+      this.trigger('playlist-update');
```
brandonocasey, line 836: new event for detecting a playlist change from the loader alone. We need this because during playlist changes it can look like playback is stalled, when really it isn't.

src/segment-loader.js
```diff
@@ -2470,6 +2471,7 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @private
    */
   handleAppendsDone_() {
+    this.trigger('updateend');
```
brandonocasey, line 2474: I needed a definitive way to determine when an append is actually done. We have appended but that only fires in a few scenarios. I could have latched on to `updateend`  on source buffers but that was much lower level. Ultimately I think an event fits best, but i'm not sure if the name should stay `updateend`

src/playback-watcher.js
```diff
@@ -352,6 +460,8 @@ export default class PlaybackWatcher {
       this.logger_(`Stopped at ${currentTime}, setting timer for ${difference}, seeking ` +
         `to ${nextRange.start(0)}`);
 
+      this.cancelTimer_();
```
brandonocasey, line 463: Noticed that it's technically possible to have two gap skipper timers, although I would wager that it is unlikely.

src/playback-watcher.js
```diff
@@ -90,6 +92,22 @@ export default class PlaybackWatcher {
...
+
+    loaderTypes.forEach((type) => {
+      loaderChecks[type] = {
+        reset: () => this.resetSegmentDownloads_(type),
+        updateend: () => this.checkSegmentDownloads_(type)
+      };
+
+      mpc[`${type}SegmentLoader_`].on('updateend', loaderChecks[type].updateend);
+      mpc[`${type}SegmentLoader_`].on('playlist-update', loaderChecks[type].reset);
+      this.tech_.on(['seeked', 'seeking'], loaderChecks[type].reset);
```
brandonocasey, line 108: Playback stalls should not be detected right after seeking. This prevents one segment playlists (such as a single vtt file for a video) from being detected as stalling as the buffer will not change, since it will be the entire video duration.

brandonocasey, line 108: Playback stalls should not be detected right after seeking. This prevents one segment playlists (such as a single vtt file for a video) from being detected as stalling as the buffer will not change, since it will be the entire video duration.

src/playback-watcher.js
```diff
@@ -90,6 +92,22 @@ export default class PlaybackWatcher {
...
+    const loaderChecks = {};
+
+    loaderTypes.forEach((type) => {
+      loaderChecks[type] = {
+        reset: () => this.resetSegmentDownloads_(type),
+        updateend: () => this.checkSegmentDownloads_(type)
+      };
+
+      mpc[`${type}SegmentLoader_`].on('updateend', loaderChecks[type].updateend);
+      mpc[`${type}SegmentLoader_`].on('playlist-update', loaderChecks[type].reset);
```
brandonocasey, line 107: If a rendition switch happens during a playback stall where the buffer isn't changing we want to reset as we cannot assume that the new rendition will also be stalled, until after new appends.

src/playback-watcher.js
```diff
@@ -126,6 +150,75 @@ export default class PlaybackWatcher {
...
+    const loader = mpc[`${type}SegmentLoader_`];
+    const buffered = loader.buffered_();
+    const isBufferedDifferent = Ranges.isRangeDifferent(this[`${type}Buffered_`], buffered);
+
+    this[`${type}Buffered_`] = buffered;
+
+    // if another watcher is going to fix the issue or
+    // the buffered value for this loader changed
+    // appends are working
+    if (isBufferedDifferent) {
```
brandonocasey, line 188: The basic premise for this function is that `buffered` should change at least once on every three appends. If it does not then we are likely infinite downloading and our appends are having no effect.

src/playback-watcher.js
```diff
@@ -126,6 +150,75 @@ export default class PlaybackWatcher {
...
+
+    this.logger_(`${type} loader download exclusion`);
+    this.resetSegmentDownloads_(type);
+    this.tech_.trigger({type: 'usage', name: `vhs-${type}-download-exclusion`});
+
+    if (type !== 'subtitle') {
+      mpc.blacklistCurrentPlaylist({
+        message: `Infinite ${type} segment downloading detected.`
+      }, Infinity);
+    } else {
```
brandonocasey, line 212: Subtitles cannot be excluded. So we have to disable and remove the track. If the track isn't appending to the buffer correctly anyway then its not going to be showing subtitles making the track useless.

test/playback-watcher.test.js
```diff
@@ -742,6 +743,293 @@ QUnit.test('jumps to buffered content if seeking just before', function(assert)
...
+      assert.equal(track.mode, 'disabled', 'mode set to disabled now');
+      assert.equal(this.env.log.warn.callCount, 1, 'warning logged');
+      assert.equal(this.env.log.warn.args[0][0], message, 'warning message as expected');
+
+      this.env.log.warn.resetHistory();
+    }
+  });
+
+  if (type !== 'subtitle') {
+    QUnit.test(`detects ${type} appends without buffer changes and excludes many playlists`, function(assert) {
```
gesinger, line 889: Since this only covers main and audio, and uses a conditional, I think it would be better to have two separate tests.

test/playback-watcher.test.js
```diff
@@ -742,6 +743,293 @@ QUnit.test('jumps to buffered content if seeking just before', function(assert)
...
+    assert.equal(this.playbackWatcher[`${type}StalledDownloads_`], 0, '0 stalled downloads after playlist-update');
+    assert.deepEqual(this.usageEvents, {}, 'no usage events');
+
+    loader.trigger('updateend');
+    assert.equal(this.playbackWatcher[`${type}StalledDownloads_`], 1, '1st append 1 stalled downloads');
+    assert.deepEqual(this.usageEvents, {}, 'no usage events');
+
+    loader.trigger('updateend');
+    assert.equal(this.playbackWatcher[`${type}StalledDownloads_`], 2, '2nd append 2 stalled downloads');
+    assert.deepEqual(this.usageEvents, {}, 'no usage events');
```
gesinger, line 1006: Any specific reason for these?

src/playback-watcher.js
brandonocasey: I agree that there could be accessibility implications and I will add a TODO.

src/ranges.js
```diff
@@ -396,3 +396,44 @@ export const timeRangesToArray = (timeRanges) => {
 
   return timeRangesList;
 };
+
+/**
+ * Determines if two time range objects are different.
+ *
+ * @param {TimeRange} a
+ *        the first buffered object to check
```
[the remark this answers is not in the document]
brandonocasey: good catch, originally I had this object called isBufferedDifferent, but realized it was more generic. I will make the change :)

test/playback-watcher.test.js
brandonocasey: You are right we can remove them

src/videojs-http-streaming.js
```diff
@@ -517,8 +517,15 @@ class HlsHandler extends Component {
...
+      let error = this.masterPlaylistController_.error;
 
-      player.error(this.masterPlaylistController_.error);
+      if (typeof error === 'object' && !error.code) {
+        error.code = 3;
+      } else if (typeof error === 'string') {
+        error = {message: error, code: 3};
+      }
+
+      player.error(error);
```
gkatsev, line 528: 👍

test/playback-watcher.test.js
brandonocasey: If we decide to do different excluding with audio loader we can do that, but right now this would just end in a copy paste of a test.

src/playback-watcher.js
```diff
@@ -205,18 +218,23 @@ export default class PlaybackWatcher {
...
       videojs.log.warn(`Text track "${label}" is not working correctly. It will be disabled and excluded.`);
       track.mode = 'disabled';
       this.tech_.textTracks().removeTrack(track);
+      return;
     }
+
+    // TODO: should we exclude audio tracks rather than main tracks
+    // when type is audio?
+    mpc.blacklistCurrentPlaylist({
+      message: `Infinite ${type} segment downloading detected.`
```
gesinger, line 236: A bit minor, but since this can potentially be seen by users, I think we'd be better off phrasing it along the lines of "Excessive" instead of "Infinite," as it more accurately reflects what's detected.

PR #832, opened 2020-04-29
chore: better selectedIndex and use mimetype

brandonocasey:
## Description
For some of the sources in the dropdown we can't guess the mimetype, but we already provide it. We should use it in those cases. Also we need to reset `selectIndex` every time the source is changed so that it is always accurate.

### DIFF
--- a/scripts/index-demo-page.js
+++ b/scripts/index-demo-page.js
@@ -26,6 +26,10 @@
         option.setAttribute('data-key-systems', JSON.stringify(source.keySystems, null, 2));
       }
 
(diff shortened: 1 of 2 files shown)


brandonocasey:
Yeah let me add that

brandonocasey:
I tried your suggestion a few ways @gkatsev but ultimately I think it would be more frustrating to have type deleted on every edit of the url input then having to manually delete it.

gkatsev:
Though, we still detect the type by default. But we can always make that change another time if we find the need for it.

PR #841, opened 2020-05-06
feat: Support switching codecs if supported by the browser

brandonocasey:
If the browser supports [`sourceBuffer.changeType`](https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer/changeType) stop excluding playlists that have a different codec than the initial selected rendition. If we switch to a playlist that has a different codec than the current rendition change codecs using `changeType`

### DIFF
--- a/index.html
+++ b/index.html
@@ -104,12 +104,6 @@
   <script src="scripts/index-demo-page.js"></script>
   <script>
     window.startDemo(function(player) {
-
-      window.changeRep = function(i) {
-        player.vhs.representations().forEach(function(rep, index) {
-          rep.enabled(i === index);
-        });
-      };
       // do something with setup player
     });
   </script>
--- a/index.html
+++ b/index.html
@@ -105,7 +105,7 @@
   <script>
     window.startDemo(function(player) {
 
-      window.changeRep = (i) => {
+      window.changeRep = function(i) {
         player.vhs.representations().forEach(function(rep, index) {
           rep.enabled(i === index);
         });
--- a/index.html
+++ b/index.html
@@ -104,6 +104,12 @@
   <script src="scripts/index-demo-page.js"></script>
   <script>
     window.startDemo(function(player) {
+
+      window.changeRep = (i) => {
+        player.vhs.representations().forEach(function(rep, index) {
+          rep.enabled(i === index);
+        });
+      };
       // do something with setup player
     });
   </script>
--- a/scripts/index-demo-page.js
+++ b/scripts/index-demo-page.js
@@ -328,10 +328,6 @@
 
     sources.addEventListener('change', function(event) {
       var selectedOption = sources.options[sources.selectedIndex];
-
-      if (!selectedOption) {
-        return;
-      }
       var src = selectedOption.value;
 
       stateEls.url.value = src;
--- a/scripts/index-demo-page.js
+++ b/scripts/index-demo-page.js
@@ -328,6 +328,10 @@
 
     sources.addEventListener('change', function(event) {
       var selectedOption = sources.options[sources.selectedIndex];
+
+      if (!selectedOption) {
+        return;
+      }
       var src = selectedOption.value;
 
       stateEls.url.value = src;
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -553,7 +553,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       const codecs = this.getCodecsOrExclude_();
 
       // no codecs, excluded
-      if (!codecs) {
+      if (!codecs || !this.sourceUpdater_.canCodecSwitch()) {
         return;
       }
 
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1290,7 +1290,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
     }
 
     // source buffers are already created
-    if (this.sourceUpdater_.ready()) {
+    if (this.sourceUpdater_.ready() && !this.sourceUpdater_.canChangeType()) {
       return;
     }
 
@@ -1373,29 +1373,31 @@ export class MasterPlaylistController extends videojs.EventTarget {
       }
     }
 
-    try {
-      this.sourceUpdater_.createSourceBuffers(codecs);
-    } catch (e) {
-      const error = 'Failed to create SourceBuffers: ' + e;
+    if (this.sourceUpdater_.ready()) {
+      this.sourceUpdater_.changeType(codecs);
+    } else {
+      try {
+        this.sourceUpdater_.createSourceBuffers(codecs);
+      } catch (e) {
+        const error = 'Failed to create SourceBuffers: ' + e;
 
-      videojs.log.warn(error);
-      this.error = error;
-      if (this.mediaSource.readyState !== 'open') {
-        this.trigger('error');
-      } else {
-        this.sourceUpdater_.endOfStream('decode');
+        videojs.log.warn(error);
+        this.error = error;
+        if (this.mediaSource.readyState !== 'open') {
+          this.trigger('error');
+        } else {
+          this.sourceUpdater_.endOfStream('decode');
+        }
+        return;
       }
+    }
+
+    if (this.sourceUpdater_.canChangeType()) {
       return;
     }
 
     const codecString = [codecs.video, codecs.audio].filter(Boolean).join(',');
 
-    // TODO:
-    // blacklisting incompatible renditions will have to change
-    // once we add support for `changeType` on source buffers.
-    // We will have to not blacklist any rendition until we try to
-    // switch to it and learn that it is incompatible and if it is compatible
-    // we `changeType` on the sourceBuffer.
     this.excludeIncompatibleVariants_(codecString);
   }
 
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1375,8 +1375,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
       const switchMessages = [];
 
       ['video', 'audio'].forEach((type) => {
-        const newCodec = parseCodecs(this.sourceUpdater_.codecs[type] || '')[type].type;
-        const oldCodec = parseCodecs(codecs[type] || '')[type].type;
+        const newCodec = (parseCodecs(this.sourceUpdater_.codecs[type] || '')[type] || {}).type;
+        const oldCodec = (parseCodecs(codecs[type] || '')[type] || {}).type;
 
         if (newCodec && oldCodec && newCodec.toLowerCase() !== oldCodec.toLowerCase()) {
           switchMessages.push(`"${this.sourceUpdater_.codecs[type]}" -> "${codecs[type]}"`);
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1407,8 +1407,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // playlist was excluded, do nothing.
     if (!codecs) {
-      this.mainSegmentLoader_.startingMedia_ = void 0;
-      this.mainSegmentLoader_.startingMedia_ = void 0;
       return;
     }
 
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1283,23 +1283,29 @@ export class MasterPlaylistController extends videojs.EventTarget {
   }
 
   getCodecsOrExclude_() {
-    const mainMedia = this.mainSegmentLoader_.currentMedia_ || this.mainSegmentLoader_.startingMedia_ || {};
-    const audioMedia = this.audioSegmentLoader_.currentMedia_ || this.audioSegmentLoader_.startingMedia_ || {};
-    const isFmp4 = mainMedia.isFmp4 || audioMedia.isFmp4;
+    const media = {
+      main: this.mainSegmentLoader_.currentMedia_ || this.mainSegmentLoader_.startingMedia_ || {},
+      audio: this.audioSegmentLoader_.currentMedia_ || this.audioSegmentLoader_.startingMedia_ || {}
+    };
+
+    // set "main" media equal to video
+    media.video = media.main;
     const playlistCodecs = codecsForPlaylist(this.master(), this.media());
     const codecs = {};
     const usingAudioLoader = !!this.mediaTypes_.AUDIO.activePlaylistLoader;
 
-    if (mainMedia.hasVideo) {
-      codecs.video = playlistCodecs.video || mainMedia.videoCodec || DEFAULT_VIDEO_CODEC;
+    if (media.main.hasVideo) {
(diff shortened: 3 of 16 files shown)


index.html
```diff
@@ -104,6 +104,12 @@ <h3>Navigation</h3>
   <script src="scripts/index-demo-page.js"></script>
   <script>
     window.startDemo(function(player) {
+
+      window.changeRep = (i) => {
```
brandonocasey, line 108: I end up using this one a lot, if we want to remove it though we can.

scripts/index-demo-page.js
```diff
@@ -256,7 +256,7 @@
           liveui: stateEls.liveui.checked,
           html5: {
             hls: {
-              overrideNative: !window.videojs.browser.IS_SAFARI
+              overrideNative: true
```
brandonocasey, line 259: This code change is only really testable on new edge and safari as other browsers don't support many native types.

scripts/index-demo-page.js
```diff
@@ -322,6 +322,10 @@
 
     sources.addEventListener('change', function(event) {
       var selectedOption = sources.options[sources.selectedIndex];
+
+      if (!selectedOption) {
```
brandonocasey, line 326: Firefox currently throws an error on the demo page.

src/master-playlist-controller.js
```diff
@@ -589,15 +599,15 @@ export class MasterPlaylistController extends videojs.EventTarget {
    *
    * @private
    */
-  smoothQualityChange_() {
-    const media = this.selectPlaylist();
+  smoothQualityChange_(media = this.selectPlaylist()) {
```
brandonocasey, line 602: Allows us to easily change the media that is currently being used

src/master-playlist-controller.js
```diff
@@ -723,17 +731,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-    } catch (e) {
-      videojs.log.warn('Failed to create Source Buffers', e);
-      if (this.mediaSource.readyState !== 'open') {
-        this.trigger('error');
-      } else {
-        this.sourceUpdater_.endOfStream('decode');
-      }
-      return;
-    }
+    this.tryToCreateSourceBuffers_();
```
brandonocasey, line 734: We won't throw an error that can be caught by this anymore

src/media-segment-request.js
```diff
@@ -290,7 +290,8 @@ const transmuxAndNotify = ({
     if (probeResult) {
       trackInfoFn(segment, {
         hasAudio: probeResult.hasAudio,
-        hasVideo: probeResult.hasVideo
+        hasVideo: probeResult.hasVideo,
+        isMuxed
```
brandonocasey, line 294: When messing around with renditions in safari I realized that we don't pass isMuxed up for ts segments. This is a problem because ts segments can contain video/audio and be appended to a "videoBuffer"  that has both codecs.

src/segment-loader.js
```diff
@@ -450,6 +490,12 @@ export default class SegmentLoader extends videojs.EventTarget {
       }
     });
 
+    this.on('trackinfo', () => {
```
brandonocasey, line 493: Since `startingMedia_` is reset on media change now, we have to check if we have enough to append after `trackinfo`, which sets `startingMedia_`

src/rendition-mixin.js
```diff
@@ -66,6 +66,12 @@ class Representation {
 
     this.bandwidth = playlist.attributes.BANDWIDTH;
 
+    if (playlist.attributes.CODECS) {
+      this.codecs = playlist.attributes.CODECS;
+    }
```
brandonocasey, line 71: Make renditions much better to use. Now we know the playlist codecs and we can use `masterPlaylistController.fastQualityChange` or `masterPlaylistController.smoothQualityChange`

src/segment-loader.js
```diff
@@ -1448,18 +1497,21 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      return;
+    }
+
+    trackInfo = trackInfo || {};
+
     // When we have track info, determine what media types this loader is dealing with.
     // Guard against cases where we're not getting track info at all until we are
     // certain that all streams will provide it.
-    if (typeof this.startingMedia_ === 'undefined' && (trackInfo.hasAudio || trackInfo.hasVideo)) {
+    if ((trackInfo.hasVideo || trackInfo.hasAudio) && !shallowEqual(this.startingMedia_, trackInfo)) {
```
brandonocasey, line 1509: If the trackinfo actually changed fire a new trackinfo event.

src/videojs-http-streaming.js
```diff
@@ -826,6 +828,11 @@ class HlsHandler extends Component {
       delete this.tech_.hls;
     }
 
+    if (this.mediaSourceUrl_ && window.URL.revokeObjectURL) {
```
brandonocasey, line 831: This was a memory leak, we were never revoking the mediaSources url blob. See https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL and https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL

test/manifests/dash-many-codecs.mpd
```diff
@@ -3,7 +3,7 @@
  <Period>
   <AdaptationSet mimeType="video/mp4" contentType="video" subsegmentAlignment="true" subsegmentStartsWithSAP="1" par="16:9">
    <SegmentTemplate duration="120" timescale="30" media="$RepresentationID$/$RepresentationID$_$Number$.m4v" startNumber="1" initialization="$RepresentationID$/$RepresentationID$_0.m4v"/>
-   <Representation id="video_1" codecs="avc1.foo.bar" bandwidth="507246" width="320" height="180" frameRate="30" sar="1:1" scanType="progressive"/>
+   <Representation id="video_1" codecs="avc1.4d400d" bandwidth="507246" width="320" height="180" frameRate="30" sar="1:1" scanType="progressive"/>
```
brandonocasey, line 6: These have to be real codecs now, as we don't blacklist on codec change.

src/master-playlist-controller.js
```diff
@@ -536,9 +536,23 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-      this.tryToCreateSourceBuffers_();
-    });
+    const updateCodecs = () => {
+      if (!this.sourceUpdater_.ready()) {
+        return this.tryToCreateSourceBuffers_();
+      }
+
+      const codecs = this.getCodecsOrExclude_();
+
+      // no codecs, excluded
```
gesinger, line 546:
```suggestion
      // no codecs means that the playlist was excluded
```

src/segment-loader.js
```diff
@@ -1448,18 +1490,33 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      if (!shallowEqual(this.currentMedia_, trackInfo)) {
+        this.currentMedia_ = trackInfo;
+        changed = true;
+      }
 
-    if (this.checkForIllegalMediaSwitch(trackInfo)) {
-      return;
+      if (changed) {
+        this.logger_('trackinfo update', trackInfo);
+        this.trigger('trackinfo');
```
gesinger, line 1516: Would it be worth renaming to `newtrackinfo` for clarity?

src/master-playlist-controller.js
```diff
@@ -1415,24 +1452,30 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
         variant.excludeUntil = Infinity;
       }
 
-      // the video codec cannot change
-      if (variantCodecs.video && codecs.video &&
-        variantCodecs.video.type.toLowerCase() !== codecs.video.type.toLowerCase()) {
-        blacklistReasons.push(`video codec "${variantCodecs.video.type}" !== "${codecs.video.type}"`);
-        variant.excludeUntil = Infinity;
-      }
+      // only exclude playlsits by codec change, if codecs cannot switch
```
gesinger, line 1463:
```suggestion
      // only exclude playlists by codec change, if codecs cannot switch
```

src/source-updater.js
```diff
@@ -219,55 +350,149 @@ export default class SourceUpdater extends videojs.EventTarget {
...
 
-  start_() {
-    this.started_ = true;
+  /**
+   * Whether or not the codecSwitch function is supported.
+   *
+   * @return {boolean}
+   *         if codecSwitch can be called
+   */
+  canCodecSwitch() {
```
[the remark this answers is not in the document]
brandonocasey: Originally I had tried to support ie 11 type switching and audio/video to audio only switching by re-creating the mediaSource and re-adding source buffer but ultimately abandoned it since it was going to be too much work for this effort. I think that we can likely remove this function now and re-add it at another time when something like that comes up.

src/source-updater.js
```diff
@@ -102,25 +115,59 @@ const shiftQueue = (type, sourceUpdater) => {
...
+const inSourceBuffers = (mediaSource, sourceBuffer) => mediaSource && sourceBuffer &&
+  Array.prototype.indexOf.call(mediaSource.sourceBuffers, sourceBuffer) !== -1;
+
 const actions = {
   appendBuffer: (bytes, segmentInfo) => (type, sourceUpdater) => {
     const sourceBuffer = sourceUpdater[`${type}Buffer`];
 
+    // can't do anything if the media source / source buffer is null
+    // or the media source does not contain this source buffer.
+    if (!inSourceBuffers(sourceUpdater.mediaSource, sourceBuffer)) {
```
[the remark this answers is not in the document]
brandonocasey: This is possible right now if we append bad data to the source buffer, or for whatever reason the sourceBuffer is removed before this appendBuffer can happen.

src/source-updater.js
```diff
@@ -219,55 +350,149 @@ export default class SourceUpdater extends videojs.EventTarget {
...
+   * Call the changeType function on a source buffer, given the code and type.
+   *
+   * @param {string} type
+   *        The type of source buffer to call changeType on.
+   *
+   * @param {string} codec
+   *        The codec string to change type with on the source buffer.
+   */
+  changeType(type, codec) {
+    if (!this.canChangeType()) {
```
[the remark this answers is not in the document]
brandonocasey: No we always check for `this.canChangeType` before trying to call `changeType`. Do you think we should remove this line or throw an Error here?

src/master-playlist-controller.js
```diff
@@ -1245,124 +1243,161 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      media.audio.isFmp4 = (media.main.hasAudio && !media.main.isMuxed) ? media.main.isFmp4 : media.audio.isFmp4;
+    }
 
-    // Because a URI is required for EXT-X-STREAM-INF tags (therefore, there must always
-    // be a playlist, even for audio only playlists with alt audio), a segment will always
-    // be downloaded for the main segment loader, and the track info parsed from it.
-    // Therefore we must always wait for the segment loader's track info.
-    if (!mainStartingMedia || (hasAltAudio && !this.audioSegmentLoader_.startingMedia_)) {
+    // no codecs, no playback.
+    if (!codecs.audio && !codecs.video) {
```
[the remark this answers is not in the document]
brandonocasey: I will see if I can add a check to make sure we have `startingMedia` at the top of this function, so that won't be possible.

src/master-playlist-controller.js
```diff
@@ -1245,124 +1245,161 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-   */
-  tryToCreateSourceBuffers_() {
-    // media source is not ready yet
-    if (this.mediaSource.readyState !== 'open') {
-      return;
+  isMediaReady_(type) {
+    return !!this[`${type}SegmentLoader_`].startingMedia_;
+  }
+
+  getCodecsOrExclude_() {
```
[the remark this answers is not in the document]
brandonocasey: 🤔

src/master-playlist-controller.js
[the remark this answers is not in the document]
brandonocasey: added

src/source-updater.js
```diff
@@ -40,7 +53,7 @@ const shiftQueue = (type, sourceUpdater) => {
   let queueEntry = sourceUpdater.queue[queueIndex];
 
   if (queueEntry.type === 'mediaSource') {
-    if (!sourceUpdater.updating()) {
+    if (!sourceUpdater.updating() && sourceUpdater.mediaSource.readyState !== 'closed') {
```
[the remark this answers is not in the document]
brandonocasey: Yes the queue can be built up before the `mediaSource` is `open`

src/source-updater.js
```diff
@@ -147,6 +194,77 @@ const actions = {
...
+    if (!inSourceBuffers(sourceUpdater.mediaSource, sourceBuffer)) {
+      return;
+    }
+
+    sourceUpdater.logger_(`Removing ${type}Buffer with codec ${sourceUpdater.codecs[type]} from mediaSource`);
+
+    try {
+      sourceUpdater.mediaSource.removeSourceBuffer(sourceBuffer);
+    } catch (e) {
+      videojs.log.warn(`Failed to removeSourceBuffer ${type}Buffer`, e);
```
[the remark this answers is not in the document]
brandonocasey: This was mostly added to prevent errors on dispose in IE 11, as sometimes it throws errors on remove. I will see if the queue continues, but I would imagine `removeSourceBuffer` is usually the last item in the queue

src/segment-loader.js
brandonocasey: perhaps: `trackinfo` and `trackinfochange`. I don't know if we should other than to prevent breaking people that may have been using the event.

src/master-playlist-controller.js
```diff
@@ -553,7 +553,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       const codecs = this.getCodecsOrExclude_();
 
       // no codecs means that the playlist was excluded
-      if (!codecs || !this.sourceUpdater_.canChangeType()) {
+      if (!codecs) {
```
brandonocasey, line 556: We do not want to check `canChangeType` here, as we do it in `getCodecsOrExclude` and blacklist if the main part of the codec was changed, but sourcebuffers can not actually changetype.

src/master-playlist-controller.js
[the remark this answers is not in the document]
brandonocasey: Perhaps we can do this in another pull request, Most of this logic comes from `tryToCreateSourceBuffers_` and it was already in one function there.

src/master-playlist-controller.js
```diff
@@ -1350,6 +1346,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
     if (usingAudioLoader && unsupportedAudio && this.media().attributes.AUDIO) {
       const audioGroup = this.media().attributes.AUDIO;
 
+      this.mediaTypes_.AUDIO.activePlaylistLoader.pause();
```
brandonocasey, line 1349: This is important because if we wait for the masterPlaylistLoader to `mediachange` and then call `abort` then we will already have appended unsupported codecs.

PR #842, opened 2020-05-07
fix: consider whether audio is enabled for main segment loader buffered ranges

gesinger:
## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -646,9 +646,9 @@ export default class SegmentLoader extends videojs.EventTarget {
       return videojs.createTimeRanges();
     }
 
-    const { hasAudio, hasVideo } = this.startingMedia_;
-
     if (this.loaderType_ === 'main') {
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -646,13 +646,23 @@ export default class SegmentLoader extends videojs.EventTarget {
       return videojs.createTimeRanges();
     }
 
-    if (this.startingMedia_.hasAudio && this.startingMedia_.hasVideo) {
-      return this.sourceUpdater_.buffered();
-    } else if (this.startingMedia_.hasAudio) {
-      return this.sourceUpdater_.audioBuffered();
+    if (this.loaderType_ === 'main') {
```
brandonocasey, line 649: What do you think of doing something like this right above this if to prevent all the weird formatting: `const {hasAudio, hasVideo} = this.startingMedia_;`

gesinger: Good call. Updated.

PR #843, opened 2020-05-13
fix: Default overrideNative to check if browser is any Safari

ipadilla4:
Changes include setting hls _overrideNative_ default to check _IS_ANY_SAFARI_ constant. 

I removed one unit test that was failing in Chrome and Firefox due to this new default value.

### DIFF
--- a/test/videojs-http-streaming.test.js
+++ b/test/videojs-http-streaming.test.js
@@ -2839,26 +2839,6 @@ QUnit.test(
   }
 );
 
-QUnit.test('has no effect if native HLS is available', function(assert) {
-  const Html5 = videojs.getTech('Html5');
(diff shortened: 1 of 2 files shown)


test/videojs-http-streaming.test.js
```diff
@@ -2857,6 +2859,30 @@ QUnit.test('has no effect if native HLS is available', function(assert) {
...
+  const player = createPlayer();
+
+  player.src({
+    src: 'http://example.com/manifest/master.m3u8',
+    type: 'application/x-mpegURL'
+  });
+
+  this.clock.tick(1);
+
+  assert.ok(player.tech_.hls, 'did not load hls tech');
```
gesinger, line 2882:
```suggestion
  assert.ok(player.tech_.hls, 'loaded hls tech');
```

PR #849, opened 2020-05-21
fix: Use revokeObjectURL dispose for created MSE blob urls

brandonocasey:
Prevents a memory leak that we have right now, caused by not revoking urls that we create.

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -746,7 +746,9 @@ class VhsHandler extends Component {
       return;
     }
 
-    this.tech_.src(window.URL.createObjectURL(this.masterPlaylistController_.mediaSource));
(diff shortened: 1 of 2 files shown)


src/videojs-http-streaming.js
```diff
@@ -829,6 +831,11 @@ class HlsHandler extends Component {
       delete this.tech_.hls;
     }
 
+    if (this.mediaSourceUrl_ && window.URL.revokeObjectURL) {
+      window.URL.revokeObjectURL(this.mediaSourceUrl_);
```
gkatsev, line 835: Alternatively, we could've gotten the blob url via `this.tech_.src()` but I think this is fine as well.

brandonocasey:
Added a test 🤞 for ie 11...

PR #859, opened 2020-06-10
feat: add external vhs properties and deprecate hls and dash references

gesinger:
## Description
Deprecate lingering references to hls from the old videojs-contrib-hls project, as well as dash. Ensure all references use "vhs" instead.

Remove unused utils/fixtures dir.

## Requirements Checklist
  - [X] Docs/guides updated


src/videojs-http-streaming.js
```diff
@@ -350,22 +354,37 @@ class HlsHandler extends Component {
       if (!_player.hasOwnProperty('hls')) {
         Object.defineProperty(_player, 'hls', {
           get: () => {
-            videojs.log.warn('player.hls is deprecated. Use player.tech().hls instead.');
+            videojs.log.warn('player.hls is deprecated. Use player.vhs instead.');
+            tech.trigger({ type: 'usage', name: 'vhs-player-access' });
```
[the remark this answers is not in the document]
gesinger: I think player.tech().vhs is fine (and makes sense), but I'd just be a bit worried that we do have a decent number of APIs and recommended calls that are at the vhs level (rather than at the video.js level) and if we recommend using player.tech().vhs people are going to see lots of warning logs in the console:

```
VIDEOJS: WARN: Using the tech directly can be dangerous. I hope you know what you're doing.
See https://github.com/videojs/video.js/issues/2617 for more info.
```

Which might seem confusing to users (and potentially bothersome). Should we consider removing that log message in video.js?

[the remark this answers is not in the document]
gkatsev: At the very least, we should make a better landing page for that log. That currently points to the v5 related issue that recommends `IWillNotUseThisInPlugins`. We can create a page on the wiki (or on the videojs.com site) that says something along the lines of "be careful of doing stuff on the tech, pass in a value like `true` to suppress the warning".

I do think the warning is useful because a lot of people copy snippets of code without looking into the code. Hopefully, that warning is helping people think more about it.

src/xhr.js
```diff
@@ -57,14 +57,14 @@ const callbackWrapper = function(request, error, response, callback) {
...
-    // Add a default timeout for all hls requests
+    // Add a default timeout
     options = mergeOptions({
       timeout: 45e3
     }, options);
 
     // Allow an optional user-specified function to modify the option
     // object before we construct the xhr request
-    const beforeRequest = XhrFunction.beforeRequest || videojs.Hls.xhr.beforeRequest;
+    const beforeRequest = XhrFunction.beforeRequest || videojs.Vhs.xhr.beforeRequest;
```
brandonocasey, line 67: I'm not sure changing an option like this will be great for users. Perhaps we need to keep `videojs.Hls` around just for this property, and log a deprecation? We can still prioritize using `videojs.Vhs.xhr.beforeRequest`

gkatsev: `videojs.Hls` is still present.

src/videojs-http-streaming.js
```diff
@@ -113,19 +113,19 @@ const handleHlsMediaChange = function(qualityLevels, playlistLoader) {
...
   });
-  handleHlsMediaChange(qualityLevels, hls.playlists);
+  handleVhsMediaChange(qualityLevels, vhs.playlists);
 };
 
 // HLS is a source handler, not a tech. Make sure attempts to use it
 // as one do not cause exceptions.
-Hls.canPlaySource = function() {
+Vhs.canPlaySource = function() {
   return videojs.log.warn('HLS is no longer a tech. Please remove it from ' +
```
brandonocasey, line 129: Should this read `HLS`/`VHS`?

src/videojs-http-streaming.js
```diff
@@ -64,16 +64,16 @@ const Hls = {
   'BUFFER_LOW_WATER_LINE_RATE',
   'BANDWIDTH_VARIANCE'
 ].forEach((prop) => {
-  Object.defineProperty(Hls, prop, {
+  Object.defineProperty(Vhs, prop, {
```
brandonocasey, line 67: We may need to keep HLS around for these too, as they are pseudo options unfortunately.

gkatsev:
https://github.com/gesinger/http-streaming/pull/11

PR #872, opened 2020-06-19
fix: initialize EME for all playlists and PSSH values

gesinger:
## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -172,11 +172,75 @@ const emeKeySystems = (keySystemOptions, videoPlaylist, audioPlaylist) => {
   return videojs.mergeOptions(keySystemOptions, keySystemContentTypes);
 };
 
+/**
+ * @typedef {Object} KeySystems
+ *
+ * keySystems configuration for https://github.com/videojs/videojs-contrib-eme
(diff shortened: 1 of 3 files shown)


src/videojs-http-streaming.js
```diff
@@ -172,11 +172,75 @@ const emeKeySystems = (keySystemOptions, videoPlaylist, audioPlaylist) => {
...
+ * @param {Object[]} mainPlaylists
+ *        The playlists found on the master playlist object
+ */
 const setupEmeOptions = ({
   player,
   sourceKeySystems,
   media,
-  audioMedia
+  audioMedia,
+  mainPlaylists
```
gkatsev, line 243: Seems like this is a newly required attribute? Can we default it to an empty array and not break if this method is called without it?

gesinger: It shouldn't be called anywhere except from the one place we call it internally, which will always provide it.

gesinger: And from tests.

gkatsev: I thought that setupEmeOptions already existed in a released version but it doesn't, so, it doesn't matter.

PR #880, opened 2020-06-25
refactor: remove duplicate bufferIntersection code

brandonocasey:
No functional change here. This refactor removes `util/buffer` the nearly identical function in `ranges.js`. While trying to find out why we sometimes download duplicate segments I cam upon a solution that used `bufferIntersection`. Thats when I noticed that these functions are doing the same thing, and `util/buffer` is just a copy paste.


src/source-updater.js
```diff
@@ -4,7 +4,7 @@
 import videojs from 'video.js';
 import logger from './util/logger';
 import noop from './util/noop';
-import { buffered } from './util/buffer';
+import { bufferIntersection } from './ranges.js';
```
brandonocasey, line 7: Use the one in ranges

src/source-updater.js
```diff
@@ -322,7 +322,15 @@ export default class SourceUpdater extends videojs.EventTarget {
   }
 
   buffered() {
-    return buffered(this.videoBuffer, this.audioBuffer);
+    if (this.audioBuffer && !this.videoBuffer) {
```
brandonocasey, line 325: The only unique functionality in that file what using audio or video buffered, if we only had a buffer for one or the other.

test/util/buffer.test.js
```diff
@@ -1,92 +0,0 @@
-import videojs from 'video.js';
-import QUnit from 'qunit';
-import { buffered } from '../../src/util/buffer';
-import {timeRangesEqual} from '../custom-assertions.js';
-
-QUnit.module('buffer');
-
-QUnit.test('buffered returns video buffered when no audio', function(assert) {
```
brandonocasey, line 8: First two tests here moved to `sourceUpdater`, the rest moved to `ranges`.

src/source-updater.js
```diff
@@ -4,7 +4,7 @@
 import videojs from 'video.js';
 import logger from './util/logger';
 import noop from './util/noop';
-import { buffered } from './util/buffer';
```
gkatsev, line 7: sounds like we can delete buffer.js?

brandonocasey: woops, removing

PR #881, opened 2020-06-25
fix: use TIME_FUDGE_FACTOR rather than rounding by decimal digits

brandonocasey:
## Description
Currently we ran into an issue with `roundSignificantDigit` because it is rounding integer whole numbers when it seems like it was only made to round floats. This can cause us to download multiple duplicate segments on slow connections. Instead I propose that we only round by `TIME_FUDGE_FACTOR` here. This will prevent us from having to do any rounding at all.

## Why the current solution is bad:
Rounding a value of `2.111111` to `2.111112` is less than 1 milliseconds a very small change but rounding from `2.1` to `2.2` is 100ms. Both are possible with the current code.

## New solution
All numbers are rounded up or down by 33 milliseconds

 ## Possible Solutions:
1. Cap the rounding at `TIME_FUDGE_FACTOR` for all numbers, but round less for longer floats. Not sure this one really makes sense as longer floats are not necessarily more accurate.
2. Don't round for numbers less than 1 decimal digits, effectively caps rounding at 99ms, which is still very high.
3. Don't round for numbers less than 2 decimal digits, effectively caps rounding at 9ms, which might not be enough? It seems to have been working until now, or we didn't notice that it wasn't

### DIFF
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -5,6 +5,7 @@
  */
 import videojs from 'video.js';
 import window from 'global/window';
+import {TIME_FUDGE_FACTOR} from './ranges.js';
 
 const {createTimeRange} = videojs;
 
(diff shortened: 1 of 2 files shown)


gkatsev (dismissed):

gesinger (commented):
Looks good, but would be good to have a test showing the behavior.

gkatsev (approved):

PR #884, opened 2020-06-26
fix: more conservative stalled download check, better logging

brandonocasey:
This was brought up because some of our users were seeing stalled download exclusions for playlists that were not stalled.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1346,6 +1346,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
     if (usingAudioLoader && unsupportedAudio && this.media().attributes.AUDIO) {
       const audioGroup = this.media().attributes.AUDIO;
 
+      this.mediaTypes_.AUDIO.activePlaylistLoader.pause();
+      this.audioSegmentLoader_.pause();
+      this.audioSegmentLoader_.abort();
       this.master().playlists.forEach(variant => {
         const variantAudioGroup = variant.attributes && variant.attributes.AUDIO;
 
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1346,9 +1346,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
     if (usingAudioLoader && unsupportedAudio && this.media().attributes.AUDIO) {
       const audioGroup = this.media().attributes.AUDIO;
 
-      this.mediaTypes_.AUDIO.activePlaylistLoader.pause();
-      this.audioSegmentLoader_.pause();
-      this.audioSegmentLoader_.abort();
       this.master().playlists.forEach(variant => {
         const variantAudioGroup = variant.attributes && variant.attributes.AUDIO;
 
--- a/src/media-groups.js
+++ b/src/media-groups.js
@@ -736,10 +736,13 @@ export const setupMediaGroups = (settings) => {
   // DO NOT enable the default subtitle or caption track.
   // DO enable the default audio track
   const audioGroup = mediaTypes.AUDIO.activeGroup();
-  const groupId = (audioGroup.filter(group => group.default)[0] || audioGroup[0]).id;
 
-  mediaTypes.AUDIO.tracks[groupId].enabled = true;
-  mediaTypes.AUDIO.onTrackChanged();
+  if (audioGroup) {
+    const groupId = (audioGroup.filter(group => group.default)[0] || audioGroup[0]).id;
+
+    mediaTypes.AUDIO.tracks[groupId].enabled = true;
+    mediaTypes.AUDIO.onTrackChanged();
+  }
 
   masterPlaylistLoader.on('mediachange', () => {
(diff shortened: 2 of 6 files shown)


src/master-playlist-controller.js
```diff
@@ -1346,9 +1346,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
     if (usingAudioLoader && unsupportedAudio && this.media().attributes.AUDIO) {
       const audioGroup = this.media().attributes.AUDIO;
 
-      this.mediaTypes_.AUDIO.activePlaylistLoader.pause();
```
brandonocasey, line 1349: this is now handled by `mediachanging` in `media-groups.js` We already reset the main loader on `mediachanging`. This prevents bad appends from excluding the main playlist and resets the playback watcher sooner.

src/playback-watcher.js
```diff
@@ -204,16 +204,18 @@ export default class PlaybackWatcher {
...
-    // not be incremented
-    if (this[`${type}StalledDownloads_`] < 3) {
+    this.logger_(`found #${this[`${type}StalledDownloads_`]} ${type} appends that did not increase buffer (possible stalled download)`, {
+      playlistId: loader.playlist_ && loader.playlist_.id,
+      buffered: Ranges.timeRangesToArray(buffered)
+
+    });
+
+    // after 5 possibly stalled appends with no reset, exclude
+    if (this[`${type}StalledDownloads_`] < 5) {
```
brandonocasey, line 214: Wait until 5 stalled downloads.

src/segment-loader.js
```diff
@@ -2492,6 +2491,8 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
+    this.trigger('appendsdone');
```
brandonocasey, line 2494: Don't report aborted appends as appends

gesinger: Do you know in what instances we will abort appends? It seems like it would only happen on dispose of the source buffers, otherwise appends will continue even if the current segment loader process is aborted.

src/media-groups.js
```diff
@@ -744,7 +744,7 @@ export const setupMediaGroups = (settings) => {
     mediaTypes.AUDIO.onTrackChanged();
   }
 
-  masterPlaylistLoader.on('mediachange', () => {
+  masterPlaylistLoader.on(['mediachange', 'mediachanging'], () => {
```
gesinger, line 747: Should we worry about this causing double aborts in some cases?

src/segment-loader.js
brandonocasey: I think it's the request itself in this case that is aborted, rather than an actual append. I was seeing it with aggressive network throttling and changing renditions.

src/media-groups.js
brandonocasey: 🤔

src/playback-watcher.js
gkatsev: Let's use a larger value here. Would 10 be ok? Maybe 15 or even 20?

src/media-groups.js
brandonocasey: After some testing, this will cause a second abort on `mediachange` but the segmentLoader was already in an aborted state and nothing bad happens. After some more thought though I wonder if we should `stopLoaders` and `resync/reset` on `mediachanging` and then `startLoaders` on `mediachange` going to see how that works

gkatsev: Not sure where you're at, but an alternative is to skip this change for the PR for right now

brandonocasey: yeah I gave myself until the end of today to try to solve it, just about to push up a commit that takes it out and fixes the tests.

brandonocasey: This took so long because this change seemed to cause other race condition issues and debugging it is quite difficult. Ultimately I think we still want some kind of fix, but it an edge case so we can probably just document it and skip for now.

src/master-playlist-controller.js
gesinger: I think this needs to be added back since `mediachanging` was removed

src/segment-loader.js
gesinger: Might be worth us adding a test for this case.

src/media-groups.js
```diff
@@ -736,13 +736,10 @@ export const setupMediaGroups = (settings) => {
   // DO NOT enable the default subtitle or caption track.
   // DO enable the default audio track
   const audioGroup = mediaTypes.AUDIO.activeGroup();
+  const groupId = (audioGroup.filter(group => group.default)[0] || audioGroup[0]).id;
 
-  if (audioGroup) {
```
gesinger, line 740: Is there a reason to remove this check? It doesn't seem to have much of an impact since it's on setup, but it ends up calling resetEverything in the main segment loader for video only streams.

brandonocasey: Good catch, going to revert this.

PR #886, opened 2020-06-29
feat: update ABR algorithm to use moving average and switching conditionally depending on the buffer level

gkatsev:
We want to update our ABR algorithm to be more reliable. We already have a moving average bandwidth estimator, but we weren't using it, so, we're switching to use it with a value of 0.55 for the decay factor.
For the bitrate adaptation, we want to start switching conditionally depending on the buffer level. If the buffer level is high, and we're switching down, we want to stick with our current rendition. We also want to stick with the current rendition if we are switching up and the buffer level is low.


gkatsev:
What I thought would be an easy change is making things a lot trickier. First, I saw the playback watcher kick in a couple of times for "excessive downloading" (https://github.com/videojs/http-streaming/pull/884).
Second, it seems like we don't try to switch every segment load? There were definitely a couple of times when it seemed like `shouldSwitchMedia` was called after several segments were downloaded rather than for every segment.
Finally, when testing down switching, I was experimenting with restricting the bandwidth via the chrome dev tools and it seems like we would miss a down switch opportunity and then run out of time and the early abort would kick in because it could no longer get the segment in time and it would switch down a lot more aggressively than if it caught the previous down switch opportunity. This may be related to the second item.

Tests likely don't work due to the changes in the default bandwidth estimation and bitrate adaptation algorithms.

gkatsev:
I think that we need to make this behind a switch, maybe we have a separate `shouldSwitchToMedia` function that gets called depending on the switch?

index.html
```diff
@@ -73,6 +73,10 @@ <h3>Options</h3>
       <input id=partial type="checkbox">
       Handle Partial (reloads player)
     </label>
+    <label>
+      <input id=buffer-water type="checkbox">
```
brandonocasey, line 77: Adds an option to our debug page to use the new playlist selection logic.

scripts/index-demo-page.js
```diff
@@ -251,13 +251,15 @@
     stateEls.partial.addEventListener('change', function(event) {
       saveState();
 
-      window.videojs.options = window.videojs.options || {};
-      window.videojs.options.vhs = window.videojs.options.vhs || {};
-      window.videojs.options.vhs.handlePartialData = event.target.checked;
+      // reload the player and scripts
+      stateEls.minified.dispatchEvent(newEvent('change'));
```
brandonocasey, line 255: I tried to use this code for the buffer water line option and noticed that it didn't really work in all cases. So I updated them both.

scripts/index-demo-page.js
```diff
@@ -329,7 +331,7 @@
           sources.dispatchEvent(newEvent('change'));
         }
         player.on('loadedmetadata', function() {
-          if (player.vhs) {
+          if (player.tech_.vhs) {
```
brandonocasey, line 334: This was causing a deprecation warning.

src/config.js
```diff
@@ -11,5 +11,11 @@ export default {
   // How much of the buffer must be filled before we consider upswitching
   BUFFER_LOW_WATER_LINE: 0,
   MAX_BUFFER_LOW_WATER_LINE: 30,
-  BUFFER_LOW_WATER_LINE_RATE: 1
+
+  // TODO: Remove this when useBufferWaterLines is removed
+  MAX_BUFFER_LOW_WATER_LINE_NEW: 16,
```
brandonocasey, line 16: Not sure what we should call this, but in order to remain backwards compatible we needed a new config value or a hard coded value in `shouldSwitchToMedia_`. Perhaps a better name? I didn't really want to make it that much longer though as the name is already unwieldy.

src/master-playlist-controller.js
```diff
@@ -58,31 +60,61 @@ const shouldSwitchToMedia = function({
     return false;
   }
 
+  const sharedLogLine = `allowing switch ${currentPlaylist && currentPlaylist.id || 'null'} -> ${nextPlaylist.id}`;
```
brandonocasey, line 63: Added tons of logging throughout this function so that we can track what is going on by turning debug logging on.

src/master-playlist-controller.js
```diff
@@ -318,7 +352,12 @@ export class MasterPlaylistController extends videojs.EventTarget {
           selectedMedia = this.selectPlaylist();
         }
 
+        if (!selectedMedia || !this.shouldSwitchToMedia_(selectedMedia)) {
```
brandonocasey, line 355: This is the only change that isn't behind a feature switch, but I think it is something that we should have been doing. This prevents us from switching aborting a bunch of things and then re-requesting the same playlist and requesting them again.

src/master-playlist-controller.js
```diff
@@ -478,6 +517,27 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.tech_.trigger({type: 'usage', name: 'hls-playlist-cue-tags'});
     }
   }
+
+  shouldSwitchToMedia_(nextPlaylist) {
+    const currentPlaylist = this.masterPlaylistLoader_.media();
```
brandonocasey, line 522: We were calling this function the same way everywhere except for `nextPlaylist` so I just added it to the mpc object.

src/master-playlist-controller.js
```diff
@@ -487,27 +547,23 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+
+      if (this.shouldSwitchToMedia_(nextPlaylist)) {
         this.masterPlaylistLoader_.media(nextPlaylist);
       }
 
       this.tech_.trigger('bandwidthupdate');
     });
+
     this.mainSegmentLoader_.on('progress', () => {
+      if (this.bufferWaterLineSelector) {
```
brandonocasey, line 559: Sometimes, especially with low bandwidth, or when bandwidth gets slower, we do not switch fast enough by waiting for an entire segment to download.

brandonocasey: I debated using this on without the flag, but I think we want to keep the current playlist switching intact wherever possible, so I did not do it.

src/master-playlist-controller.js
```diff
@@ -542,7 +598,21 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.onEndOfStream();
     });
 
-    this.mainSegmentLoader_.on('earlyabort', () => {
+    this.mainSegmentLoader_.on('earlyabort', (event) => {
+      if (this.bufferWaterLineSelector) {
```
brandonocasey, line 602: In `segmentLoader` we used to abort the download during an `earlyabort`, but that was problematic because with the new switching algorithm at certain buffer values we do not want to force a switch. Ultimately I think that we should move most of the logic for what triggers `earlyabort` in here, but doing that in a backwards compatible way wasn't going to be possible without drastic effort.

src/segment-loader.js
```diff
@@ -1326,22 +1326,22 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @return {boolean} True if the request was aborted, false otherwise
    * @private
    */
-  abortRequestEarly_(stats) {
+  earlyAbortWhenNeeded_(stats) {
```
brandonocasey, line 1329: Since this function no longer aborts, the name needed a slight change. Now we only trigger `earlyabort` whenever we think that an `earlyabort` should happen and let master playlist controller handle a switch if it thinks that it should happen.

test/loader-common.js
```diff
@@ -340,7 +340,7 @@ export const LoaderCommonFactory = ({
     });
 
     QUnit.test(
-      'aborts request at progress events if bandwidth is too low',
+      'triggers earlyabort at progress events if bandwidth is too low',
```
brandonocasey, line 343: This test changed as earlyabort no longer aborts directly.

test/master-playlist-controller.test.js
```diff
@@ -1643,12 +1643,14 @@ QUnit.test(
...
     let endList = true;
     let duration = 100;
+    let id = 0;
 
     this.masterPlaylistController.tech_.currentTime = () => currentTime;
     this.masterPlaylistController.tech_.buffered = () => videojs.createTimeRanges(buffered);
     this.masterPlaylistController.duration = () => duration;
     this.masterPlaylistController.selectPlaylist = () => {
       return {
+        id: id++,
```
brandonocasey, line 1653: This was needed as we compare playlist ids now.

src/master-playlist-controller.js
gkatsev: So, now we exclude the current playlist, but we may not switch away from it? Should we only exclude it in this case if we do end up switching away from it?

src/config.js
gkatsev: I think  it's probably good enough.

src/segment-loader.js
```diff
@@ -1326,22 +1326,22 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @return {boolean} True if the request was aborted, false otherwise
```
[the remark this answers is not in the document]
brandonocasey: It does not, I will remove this doc

src/master-playlist-controller.js
brandonocasey: We only temporarily exclude it below so that it isn't selected as the `nextPlaylist` in `selectPlaylist`. Then we remove the exclusion and make sure that we should switch to the next playlist before actually blacklisting it and switching. Somewhat confusing, I will add some comments here.

src/segment-loader.js
```diff
@@ -1402,17 +1402,15 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-      return false;
+      return;
     }
 
     // set the bandwidth to that of the desired playlist being sure to scale by
     // BANDWIDTH_VARIANCE and add one so the playlist selector does not exclude it
     // don't trigger a bandwidthupdate as the bandwidth is artifial
     this.bandwidth =
       switchCandidate.playlist.attributes.BANDWIDTH * Config.BANDWIDTH_VARIANCE + 1;
-    this.abort();
```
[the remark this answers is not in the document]
brandonocasey: in the `earlyabort` handler in `master-playlist-controller`

src/master-playlist-controller.js
gkatsev: Oh, I totally missed that. That makes sense. Comments would definitely be helpful.

brandonocasey: Also of note is that we always blacklist (which was the previous behavior) when not using the `bufferWaterLineSelector` option.

gkatsev: Yup, that's excpted, thanks.

src/segment-loader.js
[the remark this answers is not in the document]
gkatsev: I guess I don't see it? Unless I'm not understanding things. In MPC only call to segment loader's abort is inside `mediachanging` and `setCurrentTime`. Unless `mediachanging` ends up getting called as part of that and thus we don't need to call it explicitly here?

src/master-playlist-controller.js
```diff
@@ -602,14 +602,19 @@ export class MasterPlaylistController extends videojs.EventTarget {
       if (this.bufferWaterLineSelector) {
         const currentPlaylist = this.masterPlaylistLoader_.media();
 
+        // temporarily exclude the current playlist so that we can
+        // determine the next playlist that would be selected
+        // if this playlist were to be blacklisted.
```
gkatsev, line 607:
```suggestion
        // if this playlist were to be excluded.
```

src/segment-loader.js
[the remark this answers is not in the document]
brandonocasey: abort is called when we blacklist, so in the `earlyabort` handler we call `blacklistCurrentPlaylist` which will call abort and load a new playlist.

gkatsev:
Before we merge, I think we should change the option from `bufferWaterLineSelector` to something that begins with `experimental`. Can even be `experimentalBufferWaterLineSelector`. 
I have a PR incoming to here that adds a checkbox for it in the example page.

PR #76, opened 2018-04-04
feat: Utilize option to override native on tech

OshinKaramian:
## Description
Utilize the addition of `overrideNativeAudioTracks` and `overrideNativeVideoTracks` as added in https://github.com/videojs/video.js/pull/5074.

I'm in the process of adding tests, but would like eyes on this and the above noted PR before getting too far along.

## Specific Changes proposed
This PR removes the need to set `nativeAudioTracks` and `nativeVideoTracks` to false when enabling overrideNative.

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] If necessary, more likely in a feature request than a bug fix
  - [ ] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](http://jsbin.com/nofafidabo/edit?html,output))
- [ ] Reviewed by Two Core Contributors

PR #889, opened 2020-07-06
docs: native*Tracks options are no longer needed

squarebracket:
## Description
I believe using `native*Tracks: false` is no longer needed since #76, but this crept back into the docs.

## Specific Changes proposed
Remove `native*Tracks: false` from the `overrideNative` sections of the README.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,9 +335,7 @@ var player = videojs('playerId', {
   html5: {
     vhs: {
       overrideNative: true
-    },
-    nativeAudioTracks: false,
-    nativeVideoTracks: false
+    }
   }
 });
 ```


gkatsev:
It may be necessary depending on the browser you're using maybe? Like I think that it was necessary for Legacy Edge?
@mister-ben do you remember anything about it? I think you dealt with these options "recently".

squarebracket:
@gkatsev I think you're referring to [this](https://github.com/videojs/http-streaming/pull/599#discussion_r306359774)?

squarebracket:
It seems that the line was removed from the README, but not from the code snippet.

gkatsev:
Yeah, that was probably it.

gkatsev:
This comment seems to imply it's still needed? https://github.com/videojs/http-streaming/issues/920#issuecomment-667950213

gkatsev:
Well, just noticed that Chrome 85 (85.0.4183.121) on macOS has audio tracks now which causes the native audio track to show up. The workaround right now is to enable nativeAudioTracks: false. So, I think this is still needed.

PR #893, opened 2020-07-07
fix: have reloadSourceOnError get src from player

gkatsev:
Previously, we were getting the source from tech.currentSource_. This
only gets set when a source handler like VHS is used. However, because
VHS is in Video.js by default and the plugin is global, some people are
using this plugin for other sources as well.

There isn't a reason why we should be relying on tech.currentSource_
over player.currentSource() when we have access to the player in the
plugin.

Fixes videojs/video.js#6744

### DIFF
--- a/test/reload-source-on-error.test.js
+++ b/test/reload-source-on-error.test.js
@@ -82,7 +82,7 @@ QUnit.test('triggers on player error', function(assert) {
   );
 });
 
-QUnit.test('triggers on player error, uses player.currentSource() is tech.currentSource_ is not set', function(assert) {
+QUnit.test('triggers on player error, uses player.currentSource() if tech.currentSource_ is not set', function(assert) {
   const oldsrc = this.tech.currentSource_;
   const src = {
     src: 'thisisanothersource.m3u8',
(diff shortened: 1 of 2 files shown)


test/reload-source-on-error.test.js
```diff
@@ -78,8 +78,35 @@ QUnit.test('triggers on player error', function(assert) {
   assert.deepEqual(
     this.player.src.calledWith[0],
     this.tech.currentSource_,
+    'player.src was called with tech.currentSource_'
+  );
+});
+
+QUnit.test('triggers on player error, uses player.currentSource() is tech.currentSource_ is not set', function(assert) {
```
gesinger, line 85:
```suggestion
QUnit.test('triggers on player error, uses player.currentSource() if tech.currentSource_ is not set', function(assert) {
```

Raino:
@gkatsev Do I understand correctly that You fixed this? If so, when will the new version of the player be released and will there be documentation for the parameter `reloadSourceOnError`?

gkatsev:
@Raino it's fixed. It may be a bit to get released as it's tied with a lot of other changes. In the meantime, you could grab the code from https://github.com/videojs/http-streaming/blob/main/src/reload-source-on-error.js and load it manually.

Raino:
@gkatsev How can I install this on my site? Perhaps I can download a full archive of all the scripts with the latest changes somewhere?

PR #894, opened 2020-07-08
fix: do not wait for audio appends for muxed segments

brandonocasey:
## Description
Noticed that our fmp4/ts mixed content is failing. This is happening because we fire `trackinfo` on a change when we did not before. The trackinfo that we fire for the ts segment causes us to stall palyback because we wait for `audio`, but only ever receiver `video`/`combined` back from the transmuxer. The [fix for muxed fmp4 was only reporting `hasVideo`](https://github.com/videojs/http-streaming/blob/e0e497f7ca00aa061147e4dd005d03cb802f9cec/src/media-segment-request.js#L412). The fix for ts is exactly the same, we should not report `hasAudio` if the audio and video are muxed together or else we will wait for audio appends that will never come.

### DIFF
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -289,13 +289,13 @@ const transmuxAndNotify = ({
 
     if (probeResult) {
       trackInfoFn(segment, {
-        hasAudio: !isMuxed && probeResult.hasAudio,
+        hasAudio: probeResult.hasAudio,
         hasVideo: probeResult.hasVideo,
         isMuxed
       });
       trackInfoFn = null;
 
-      if (probeResult.hasAudio) {
+      if (probeResult.hasAudio && !isMuxed) {
         audioStartFn(probeResult.audioStart);
       }
       if (probeResult.hasVideo) {
@@ -321,7 +321,6 @@ const transmuxAndNotify = ({
       if (trackInfoFn) {
         if (isMuxed) {
           trackInfo.isMuxed = true;
-          trackInfo.hasAudio = false;
         }
         trackInfoFn(segment, trackInfo);
       }
@@ -412,7 +411,6 @@ const handleSegmentBytes = ({
 
     if (tracks.video && tracks.audio) {
       trackInfo.isMuxed = true;
-      trackInfo.hasAudio = false;
     }
 
     // since we don't support appending fmp4 data on progress, we know we have the full
@@ -426,7 +424,7 @@ const handleSegmentBytes = ({
     // decoding).
     const timingInfo = mp4probe.startTime(segment.map.timescales, bytesAsUint8Array);
 
-    if (trackInfo.hasAudio) {
+    if (trackInfo.hasAudio && !trackInfo.isMuxed) {
       timingInfoFn(segment, 'audio', 'start', timingInfo);
     }
 
@@ -439,7 +437,7 @@ const handleSegmentBytes = ({
       // for it to be audio only. See `tracks.video && tracks.audio` if statement
       // above.
       // we make sure to use segment.bytes here as that
-      dataFn(segment, {data: bytes, type: trackInfo.hasAudio ? 'audio' : 'video'});
+      dataFn(segment, {data: bytes, type: trackInfo.hasAudio && !trackInfo.isMuxed ? 'audio' : 'video'});
       if (captions && captions.length) {
         captionsFn(segment, captions);
       }
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -289,7 +289,7 @@ const transmuxAndNotify = ({
 
     if (probeResult) {
       trackInfoFn(segment, {
-        hasAudio: probeResult.hasAudio,
+        hasAudio: !isMuxed && probeResult.hasAudio,
         hasVideo: probeResult.hasVideo,
         isMuxed
       });
@@ -321,6 +321,7 @@ const transmuxAndNotify = ({
       if (trackInfoFn) {
         if (isMuxed) {
           trackInfo.isMuxed = true;
+          trackInfo.hasAudio = false;
         }
         trackInfoFn(segment, trackInfo);
       }
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1690,6 +1690,7 @@ export default class SegmentLoader extends videojs.EventTarget {
         return false;
       }
 
+      // muxed content only relies on video timeing information for now.
       if (hasAudio && !this.audioDisabled_ && !isMuxed && !segmentInfo.audioTimingInfo) {
         return false;
       }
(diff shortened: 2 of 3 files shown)


src/media-segment-request.js
```diff
@@ -289,7 +289,7 @@ const transmuxAndNotify = ({
 
     if (probeResult) {
       trackInfoFn(segment, {
-        hasAudio: probeResult.hasAudio,
+        hasAudio: !isMuxed && probeResult.hasAudio,
```
gesinger, line 292: It seems a bit strange to me that we are reporting that we don't have audio even though we do. Would it be better to perform the check for isMuxed and hasAudio and separating the behavior at a higher level, so we always report what we do see in the result?

brandonocasey: It is a bit strange, I agree, but I don't know if we want to do that or not. I am open to a refactor but lets go over the reasons I have to keep it right now. 

The reason we have it setup like this is that we will only ever append to and create a `videoBuffer` in these situations as `videoBuffer` will have both codecs: `video/mp4;codecs="avc1.4d400d,mp4a.40.2"`. 
This means that our `videoBuffer` is more of a combined buffer. The segment loader, and anyone else shouldn't care that the segment has audio, since its just a big bundle of data that will only ever be treated as video.  We won't ever see any audio data and the `audioBuffer` won't ever be created.  All data will be spit out by the muxer will [reported as `combined` and changed to `video` ](https://github.com/videojs/http-streaming/blob/e0e497f7ca00aa061147e4dd005d03cb802f9cec/src/media-segment-request.js#L317)

If we want to make such a change I have the code done. I'm just not sure what is best here.

https://github.com/videojs/http-streaming/compare/fix/muxed-no-has-audio...wip/is-muxed?expand=1

gesinger: We should add in @gkatsev too, but I'm inclined to say that if we make changes like this, we should probably do it now. It's hard to read the code as `hasAudio` having the meaning that it's audio only, rather than a component of muxed.

gkatsev: are you saying to rename `hasAudio` to something like `audioOnlyMedia`?

gesinger: Probably more of what Brandon prepared in the link, that hasAudio and hasVideo remain descriptors of if the media has audio and video, and isMuxed is passed along to describe whether the content is muxed together or not.

gkatsev: Sounds good to me.

brandonocasey: done

src/segment-loader.js
```diff
@@ -637,9 +637,9 @@ export default class SegmentLoader extends videojs.EventTarget {
     }
 
     if (this.loaderType_ === 'main') {
-      const { hasAudio, hasVideo } = this.startingMedia_;
+      const { hasAudio, hasVideo, isMuxed } = this.startingMedia_;
 
-      if (hasVideo && hasAudio && !this.audioDisabled_) {
+      if (hasVideo && hasAudio && !this.audioDisabled_ && !isMuxed) {
```
gesinger, line 642: Should this check for the opposite? Since if the track has audio and video and audio is not disabled and isMuxed then we should be using the combined buffered.

src/segment-loader.js
```diff
@@ -1684,11 +1684,13 @@ export default class SegmentLoader extends videojs.EventTarget {
...
     if (!this.handlePartialData_) {
-      if (this.startingMedia_.hasVideo && !segmentInfo.videoTimingInfo) {
+      const {hasAudio, hasVideo, isMuxed} = this.startingMedia_;
+
+      if (hasVideo && !segmentInfo.videoTimingInfo) {
         return false;
       }
 
-      if (this.startingMedia_.hasAudio && !segmentInfo.audioTimingInfo) {
+      if (hasAudio && !this.audioDisabled_ && !isMuxed && !segmentInfo.audioTimingInfo) {
```
gesinger, line 1693: Might be worth adding a comment above saying that muxed content only relies on the video timing info right now.

src/segment-loader.js
```diff
@@ -2336,8 +2338,9 @@ export default class SegmentLoader extends videojs.EventTarget {
     // Although transmuxing is done, appends may not yet be finished. Throw a marker
     // on each queue this loader is responsible for to ensure that the appends are
     // complete.
-    const waitForVideo = this.loaderType_ === 'main' && this.startingMedia_.hasVideo;
-    const waitForAudio = !this.audioDisabled_ && this.startingMedia_.hasAudio;
+    const {hasAudio, hasVideo, isMuxed} = this.startingMedia_;
+    const waitForVideo = this.loaderType_ === 'main' && hasVideo;
+    const waitForAudio = !this.audioDisabled_ && hasAudio && !isMuxed;
```
gesinger, line 2343: It may be OK to skip for now, but we should probably add a TODO that I believe this may break partial support in some cases.

src/segment-loader.js
brandonocasey: So in the case of `isMuxed` we will only ever have a `videoBuffer` so we only want to check `videoBuffered`. We could forgo this check here and rely on `sourceUpdater_`. Since `sourceUpdater` already checks what buffers exist before trying to called buffered on them.

gesinger: If we have alt audio available but muxed at first (an option of demuxed), there will be both audio and video buffers (e.g., the first bipbop in the index page).

brandonocasey: Yeah so we will have to check `isMuxed` here, as in the case of `isMuxed` we only care about the `videoBuffer`.

gesinger: Ah, so even if there are both audio and video buffers, but the stream is muxed and we're not using an alt audio playlist, then we are only appending to the video buffer, but appending both audio and video? Is there a valid case then for ever using the buffered intersection in a segment loader then?

brandonocasey: Thats a good question and the answer might actually be no, maybe a TODO/issue that we need to open.

brandonocasey: We only set isMuxed when we have a mixed fmp4/ts playlist where the fmp4 part is muxed.  In that case we *have* to mux together video and audio to get it to keep playing.
Normally for muxed content we split it into separate audio and video in the muxer and append to different buffers. In the normal case we will hit these lines.

gesinger: Ah, so the isMuxed is from the result of the transmuxer, telling us whether we've muxed the content together or not. Got it, thanks!

src/segment-loader.js
```diff
@@ -1684,11 +1684,14 @@ export default class SegmentLoader extends videojs.EventTarget {
...
     if (!this.handlePartialData_) {
-      if (this.startingMedia_.hasVideo && !segmentInfo.videoTimingInfo) {
+      const {hasAudio, hasVideo, isMuxed} = this.startingMedia_;
+
+      if (hasVideo && !segmentInfo.videoTimingInfo) {
         return false;
       }
 
-      if (this.startingMedia_.hasAudio && !segmentInfo.audioTimingInfo) {
+      // muxed content only relies on video timeing information for now.
```
gesinger, line 1693:
```suggestion
      // muxed content only relies on video timing information for now.
```

PR #895, opened 2020-07-08
fix: stop alt loaders on main mediachanging to prevent append race

brandonocasey:
## Description
This is most noticeable right now on safari on the hevc fmp4 stream that we have for testing. You can reproduce it in `main` by seeking around and force changing the representation to one with another codec:

```js
vhs.representations().forEach((r, i) => r.enabled(i ===49))
vhs.representations().forEach((r, i) => r.enabled(i ===53))
vhs.representations().forEach((r, i) => r.enabled(i ===20))
```
Usually it will fail with an append error, because we change the codec and then try to append old codec data. This happens because we previously waited until `mediachange` to `stopLoaders` for `audioLoader` but for the `mainLoader` we stop right away. This causes a race condition where old audio gets appended before the media has fully changed, but after codecs have. The fix here is to stop/pause  all segment loaders on `mediachanging`

## Why didn't we see it sooner
Its not an issue to append data of the same codec to a sourcebuffer even if it isn't relevant. That and we always reset when starting to load that data append never caused an issue. Now that we switch codecs such an append will throw an error.

### DIFF
--- a/src/media-groups.js
+++ b/src/media-groups.js
@@ -105,7 +105,7 @@ export const onGroupChanged = (type, settings) => () => {
   startLoaders(activeGroup.playlistLoader, mediaType);
 };
 
-const onGroupChanging = (type, settings) => () => {
+export const onGroupChanging = (type, settings) => () => {
   const {
     segmentLoaders: {
       [type]: segmentLoader
--- a/src/media-groups.js
+++ b/src/media-groups.js
@@ -105,6 +105,17 @@ export const onGroupChanged = (type, settings) => () => {
   startLoaders(activeGroup.playlistLoader, mediaType);
 };
 
+const onGroupChanging = (type, settings) => () => {
+  const {
+    segmentLoaders: {
+      [type]: segmentLoader
+    }
+  } = settings;
+
+  segmentLoader.abort();
+  segmentLoader.pause();
+};
+
 /**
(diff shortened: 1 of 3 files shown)


src/media-groups.js
```diff
@@ -105,6 +105,17 @@ export const onGroupChanged = (type, settings) => () => {
   startLoaders(activeGroup.playlistLoader, mediaType);
 };
 
+const onGroupChanging = (type, settings) => () => {
+  const {
```
gesinger, line 109: Should this use stopLoaders? And should onGroupChanged no longer stopLoaders?

brandonocasey: Originally I tried using `stopLoaders`  but that caused an issue because we abort the current audio playlist loader request when starting playback even though we aren't changing audio playlists. I will check if `mediachange` should just stop the audio playlist loader, since the segmentLoader will be stopped.

brandonocasey: OK after testing out changing `onGroupChanged` to only stop the `activePlaylistLoader` is doesn't work. Loaders are started and stopped frequently during the start of playback and if we don't stop them in `mediachange` too it can cause playback to hang because the segment loader is trying to load, but there isn't an `activePlaylistlistLoader`.

brandonocasey: test added

PR #896, opened 2020-07-08
fix: do not try to use unsupported audio

brandonocasey:
## Description
Prevents unknown content from making it to the muxer and freezing the tab, such as ac3/ec3 audio in advanced bipbop ts.


test/media-segment-request.test.js
```diff
@@ -766,6 +767,135 @@ QUnit.test('non-TS segment will get parsed for captions on next segment request
...
+  });
+
+  mediaSegmentRequest(options);
+
+  assert.equal(this.requests.length, 1, 'there is one request');
+
+  const segmentReq = this.requests.shift();
+
+  assert.equal(segmentReq.uri, 'foo.ac3', 'segment-request');
+  this.standardXHRResponse(segmentReq, new Uint8Array([0x00, 0x00, 0x00]));
```
gesinger, line 810: Might be good for us to use an actual ac3 file here, as it should catch the case we saw with the ID3 signature (if it ever returns).

test/media-segment-request.test.js
```diff
@@ -1171,3 +1050,377 @@ QUnit.skip('id3 callback does not fire if partial data has no ID3 tags', functio
...
+      this.calls[name] = 0;
+      this.options[`${name}Fn`] = () => this.calls[name]++;
+    });
+
+  },
+  afterEach(assert) {
+    this.transmuxer = this.options.segment.transmuxer;
+    sharedHooks.afterEach.call(this, assert);
+  }
+});
```
gesinger, line 1087:
```suggestion
});

```

test/media-segment-request.test.js
```diff
@@ -1171,3 +1050,377 @@ QUnit.skip('id3 callback does not fire if partial data has no ID3 tags', functio
...
+  };
+
+  mediaSegmentRequest(this.options);
+
+  assert.equal(this.requests[0].uri, 'foo.aac', 'segment-request');
+  this.standardXHRResponse(this.requests[0], aacWithId3Segment());
+});
+
+/** TODO: fix this case in mux.js
+QUnit.test('aac without id3 will make it to the transmuxer', function(assert) {
```
gesinger, line 1233: Instead of a comment, should we do a skip with the TODO for more visibility? (and same for one further down)

test/media-segment-request.test.js
```diff
@@ -1171,3 +1050,377 @@ QUnit.skip('id3 callback does not fire if partial data has no ID3 tags', functio
...
+      id3: 0,
+      videoSegmentTimingInfo: 0
+    }, 'calls as expeceted');
+    done();
+  };
+
+  mediaSegmentRequest(this.options);
+
+  assert.equal(this.requests[0].uri, 'foo.aac', 'segment-request');
+  this.standardXHRResponse(this.requests[0], aacSegment());
```
gesinger, line 1254: Should this be something like `aacWithoutId3Segment`? (and same for one further down)

test/media-segment-request.test.js
brandonocasey: sure

brandonocasey:
nope didn't fix anything, but I found out the cause

test/media-segment-request.test.js
```diff
@@ -88,6 +92,386 @@ QUnit.module('Media Segment Request', {
       this.transmuxer.terminate();
     }
   }
+
+};
+
+QUnit.module('Media Segment Request - make it to transmuxer', {
+  beforeEach(assert) {
+    assert.timeout(50000);
```
gesinger, line 100: Should we remove the timeout now?

src/media-segment-request.js
```diff
@@ -488,6 +488,16 @@ const handleSegmentBytes = ({
     return;
   }
 
+  if (typeof segment.container === 'undefined') {
+    segment.container = detectContainerForBytes(bytesAsUint8Array);
+  }
+
+  if (segment.container !== 'ts' && segment.container !== 'aac') {
+    trackInfoFn(segment, {hasAudio: true, hasVideo: false});
```
gesinger, line 496: Why are we reporting audio and not video if the container type is unknown?

brandonocasey: Originally this was because we were working with a bad audio container, but I guess it's possible that we could get here for bad video too. I pushed a commit that makes both `hasAudio` and `hasVideo` `false` here. Needed other changes further up the chain to support that though.

test/media-segment-request.test.js
brandonocasey: removed.

PR #899, opened 2020-07-10
fix: support aac data with id3 tags with mux.js@5.6.6

brandonocasey:
## Description
Noticed while generating some aac content for tests using ffmpeg that we do not support aac content that does not have ID3 tags. This was fixed with mux.js@5.6.6 as mux.js did not recognize aac content without id3 tags as aac content.

### DIFF
--- a/test/media-segment-request.test.js
+++ b/test/media-segment-request.test.js
@@ -256,7 +256,7 @@ QUnit.test('aac with id3 will make it to the transmuxer', function(assert) {
   this.options.doneFn = () => {
     assert.deepEqual(this.calls, {
       data: 1,
-      trackInfo: 48,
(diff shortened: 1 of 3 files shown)


test/media-segment-request.test.js
```diff
@@ -256,7 +256,7 @@ QUnit.test('aac with id3 will make it to the transmuxer', function(assert) {
   this.options.doneFn = () => {
     assert.deepEqual(this.calls, {
       data: 1,
-      trackInfo: 48,
+      trackInfo: 1,
```
brandonocasey, line 259: We now only fire trackinfo on the muxer once for audio, rather than every time we fire data. The partial muxer and video stream already did this.

PR #901, opened 2020-07-13
chore(demo): add a representations selector to the demo page

brandonocasey:
## Description
While looking into issues related to codec switching and representation/playlist changes I implemented local changes to select representations using the demo ui. I think this would benefit all of us, so I cleaned it up and wrapped it into a pull request.

### DIFF
--- a/scripts/index-demo-page.js
+++ b/scripts/index-demo-page.js
@@ -1,7 +1,27 @@
 /* global window document */
 /* eslint-disable vars-on-top, no-var, object-shorthand, no-console */
 (function(window) {
+  var representationsEl = document.getElementById('representations');
 
+  representationsEl.addEventListener('change', function() {
+    var selectedIndex = representationsEl.selectedIndex;
+
+    if (!selectedIndex || selectedIndex < 1 || !window.vhs) {
+      return;
+    }
+    var selectedOption = representationsEl.options[representationsEl.selectedIndex];
+
+    if (!selectedOption) {
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


scripts/index-demo-page.js
```diff
@@ -277,10 +327,15 @@
         } else {
           sources.dispatchEvent(newEvent('change'));
         }
-        player.on('loadedmetadata', () => {
+        player.on('loadedmetadata', function() {
```
brandonocasey, line 330: As you can guess, ie 11 didn't like this

scripts/index-demo-page.js
```diff
@@ -165,6 +185,36 @@
     onload();
   };
 
+  var regenerateRepresentations = function() {
+    while (representationsEl.firstChild) {
```
brandonocasey, line 189: As things are blacklisted representations can be removed/added.

scripts/index-demo-page.js
```diff
@@ -165,6 +185,36 @@
...
+      });
+
+      if (window.mpc.media().id === rep.id) {
+        selectedIndex = i;
+      }
+
+      representationsEl.appendChild(option);
+    });
+
+    representationsEl.selectedIndex = selectedIndex;
```
brandonocasey, line 215: This does not trigger a `change` event on `representationsEl` as that only happens when not using the js api.

gkatsev:
slowly we're adding stats page features

brandonocasey:
haha yep

PR #902, opened 2020-07-13
fix: pause/abort loaders before an exclude, prevent bad appends

brandonocasey:
## Description
1. Pause loading needs to call abort as well, or else we seem to still be able to continue requesting segments after an error. All of the segments will fail to append, but we will waste a lot of bandwidth.
2. We call pauseLoading after an exclusion to prevent the excluded playlist from appending a bad segment. 

## Why didn't this come up sooner
1. When appending same-codec content appends don't really error, different codec content though does.

## Testing
1. Go to the index page in main change renditions to the hevc renditions as fast as possible and notice an append error.
2. Try the same on this branch

### DIFF
--- a/scripts/index-demo-page.js
+++ b/scripts/index-demo-page.js
@@ -18,9 +18,10 @@
     var id = selectedOption.value;
 
     window.vhs.representations().forEach(function(rep) {
-      rep.enabled(rep.id === id);
+      rep.playlist.disabled = rep.id !== id;
     });
 
+    window.mpc.smoothQualityChange_();
   });
   var hlsOptGroup = document.querySelector('[label="hls"]');
   var dashOptGroup = document.querySelector('[label="dash"]');
@@ -331,9 +332,7 @@
           if (player.vhs) {
             window.vhs = player.tech_.vhs;
             window.mpc = player.tech_.vhs.masterPlaylistController_;
-            window.mpc.masterPlaylistLoader_.on('mediachange', function() {
-              regenerateRepresentations();
-            });
+            window.mpc.masterPlaylistLoader_.on('mediachange', regenerateRepresentations);
             regenerateRepresentations();
 
           } else {
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -996,8 +996,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
    * @param {string} filter
    *        Filter loaders that should call fnNames using a string. Can be:
    *        * all - run on all loaders
-   *        * playlist - run on all playlist loaders
-   *        * segment - run on all segment loaders
    *        * audio - run on all audio loaders
    *        * subtitle - run on all subtitle loaders
    *        * main - run on the main/master loaders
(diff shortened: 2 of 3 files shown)


src/master-playlist-controller.js
```diff
@@ -956,6 +956,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.trigger('error');
       return;
     }
+
+    this.pauseLoading();
```
gesinger, line 960: Are we sure we want to abort and pause all loaders if we're blacklisting one?

brandonocasey: Good question. I think that we may be able to narrow when/what we pauseLoading during a blacklist. For instance this function can be used to blacklist things other than the current playlist and we should only pauseLoading if we are blacklisting the current playlist. It's also possible that we should only stop the audio loaders when the audio group is actually changing. Let me see what I can do.

brandonocasey: Updated the code to do the following:
1. Check if the audio/subtitle group changed, if it did then we know that we need to pause loading.
2. Verify that `currentPlaylist` is actually the `media`, as its possible that it isn't

src/master-playlist-controller.js
```diff
@@ -957,55 +961,79 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
    * or both types of loaders.
    *
+   * @param {string} filter
+   *        Filter loaders that should call fnNames using a string. Can be:
+   *        * all - run on all loaders
+   *        * playlist - run on all playlist loaders
+   *        * segment - run on all segment loaders
+   *        * audio - run on all audio loaders
+   *        * subtitle - run on all subtitle loaders
+   *        * video/main - run on the main/master loaders
```
gesinger, line 1003: Is it worth having video as an option if we only use this internally?

brandonocasey: probably not.

brandonocasey:
Added tests for everything

scripts/index-demo-page.js
```diff
@@ -18,9 +18,10 @@
     var id = selectedOption.value;
 
     window.vhs.representations().forEach(function(rep) {
-      rep.enabled(rep.id === id);
+      rep.playlist.disabled = rep.id !== id;
```
brandonocasey, line 21: The original implementation changes playlists every time enabled is changed. Now we only change once when we call `smoothQualityChange_` below this. Since all playlists other than the one we want are disabled, the one we want is selected and changed to.

src/master-playlist-controller.js
```diff
@@ -956,35 +960,97 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+   *        * subtitle - run on all subtitle loaders
+   *        * main - run on the main/master loaders
+   *
+   * @param {Array|string} fnNames
+   *        A string or array of function names to call.
+   */
+  delegateLoaders_(filter, fnNames) {
+    const loaders = [];
+
+    if (filter !== 'segment') {
```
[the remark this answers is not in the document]
brandonocasey: We could, but I have a few refactors with those options in mind that I didn't really want to include in this pull request. IE switching `load`, `setCurrentTime`, and a few other places that need some testing to use `segment`. Perhaps I can reimplement them at that time.

brandonocasey:
Added tests and removed segment/playlist options @gesinger

PR #903, opened 2020-07-13
feat: exclude all incompatable browser/muxer codecs

brandonocasey:
## Description
Brings back code from 1.x where we exclude incompatible codecs. This time though we exclude by browser/muxer, rather than just browser.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -304,7 +304,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       let updatedPlaylist = this.masterPlaylistLoader_.media();
 
       if (!updatedPlaylist) {
-        // blacklist any variants that are not supported by the browser before selecting
+        // exclude any variants that are not supported by the browser before selecting
         // an initial media as the playlist selectors do not consider browser support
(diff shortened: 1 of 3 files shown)


src/master-playlist-controller.js
```diff
@@ -304,6 +304,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
       let updatedPlaylist = this.masterPlaylistLoader_.media();
 
       if (!updatedPlaylist) {
+        // blacklist any variants that are not supported by the browser before selecting
```
gkatsev, line 307:
```suggestion
        // exclude any variants that are not supported by the browser before selecting
```

src/master-playlist-controller.js
```diff
@@ -1436,6 +1440,23 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.excludeIncompatibleVariants_(codecString);
   }
 
+  /**
+   * Blacklists playlists with codecs that are unsupported by the muxer and browser.
```
gkatsev, line 1444:
```suggestion
   * Excludes playlists with codecs that are unsupported by the muxer and browser.
```

PR #904, opened 2020-07-13
fix: do not use remove source buffer on ie 11

brandonocasey:
## Description
It seems like ie 11 likes to error when using removeSourceBuffer, let's just say that it isn't supported and not use it.

### DIFF
--- a/src/source-updater.js
+++ b/src/source-updater.js
@@ -413,7 +413,7 @@ export default class SourceUpdater extends videojs.EventTarget {
    *          if removeSourceBuffer can be called.
    */
   canRemoveSourceBuffer() {
-    return !videojs.IE_VERSION && window.MediaSource &&
+    return !videojs.browser.IE_VERSION && window.MediaSource &&
       window.MediaSource.prototype &&
       typeof window.MediaSource.prototype.removeSourceBuffer === 'function';
   }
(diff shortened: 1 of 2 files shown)


src/source-updater.js
```diff
@@ -413,7 +413,7 @@ export default class SourceUpdater extends videojs.EventTarget {
    *          if removeSourceBuffer can be called.
    */
   canRemoveSourceBuffer() {
-    return window.MediaSource &&
+    return !videojs.browser.IE_VERSION && window.MediaSource &&
```
gesinger, line 416: Might be worth a comment

src/source-updater.js
```diff
@@ -413,7 +413,10 @@ export default class SourceUpdater extends videojs.EventTarget {
    *          if removeSourceBuffer can be called.
    */
   canRemoveSourceBuffer() {
-    return window.MediaSource &&
+    // IE reports that is supports removeSourceBuffer, but often throws
```
gesinger, line 416:
```suggestion
    // IE reports that it supports removeSourceBuffer, but often throws
```

test/segment-loader.test.js
```diff
@@ -833,7 +807,8 @@ QUnit.module('SegmentLoader', function(hooks) {
           done();
```
gesinger, line 807: It won't let me write on the other line, but may be worth updating the test message above.

PR #911, opened 2020-07-17
feat: expose canChangeType on the VHS property

gkatsev:


src/source-updater.js
```diff
@@ -428,12 +428,23 @@ export default class SourceUpdater extends videojs.EventTarget {
...
   }
 
+  /**
+   * Whether or not the changeType function is supported
+   * on our SourceBuffers.
+   *
+   * @return {boolean}
+   *         if changeType can be called.
+   */
+  canChangeType() {
```
gesinger, line 444: Is it worth having this too vs just using the static function?

gkatsev: This method is used internally and makes its usage slightly easier.

PR #913, opened 2020-07-20
fix: always append init segment after trackinfo change

brandonocasey:
This pull request does three things:

1. Always append an init segment after a `trackinfo` update. We do this because if the `trackinfo` causes a `changeType` we **have** to append an init segment or chrome will media error decode.
2. Check for individual segment trackinfo before appending and allow appending after trackinfo
3. Check for an abort after a trackinfo event is triggered, in the case where the type we are changing to is not supported.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1914,7 +1914,7 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     this.sourceUpdater_.appendBuffer({segmentInfo, type, bytes}, (error) => {
       if (error) {
-        this.error(`append of ${bytes.length} failed for ${type} data of segment #${segmentInfo.mediaIndex} for playlist ${segmentInfo.playlist.id}`);
+        this.error(`${type} append of ${bytes.length}b failed for segment #${segmentInfo.mediaIndex} in playlist ${segmentInfo.playlist.id}`);
         // If an append errors, we can't recover.
         // (see https://w3c.github.io/media-source/#sourcebuffer-append-error).
         // Trigger a special error so that it can be handled separately from normal,
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1453,11 +1453,30 @@ export default class SegmentLoader extends videojs.EventTarget {
     // Guard against cases where we're not getting track info at all until we are
     // certain that all streams will provide it.
     if (!shallowEqual(this.startingMedia_, trackInfo)) {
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -1895,7 +1914,7 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     this.sourceUpdater_.appendBuffer({segmentInfo, type, bytes}, (error) => {
       if (error) {
-        this.error(`appenderror for ${type} append with ${bytes.length} bytes`);
+        this.error(`append of ${bytes.length} failed for ${type} data of segment #${segmentInfo.mediaIndex} for playlist ${segmentInfo.playlist.id}`);
```
brandonocasey, line 1917: more descriptive error message about the failure

src/segment-loader.js
```diff
@@ -1895,7 +1914,7 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     this.sourceUpdater_.appendBuffer({segmentInfo, type, bytes}, (error) => {
       if (error) {
-        this.error(`appenderror for ${type} append with ${bytes.length} bytes`);
+        this.error(`${type} append of ${bytes.length}b failed for segment #${segmentInfo.mediaIndex} in playlist ${segmentInfo.playlist.id}`);
```
gkatsev, line 1917: 👍

gesinger: Double 👍

src/segment-loader.js
```diff
@@ -1453,11 +1453,30 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        audio: true,
+        video: true
+      };
+
       this.startingMedia_ = trackInfo;
       this.logger_('trackinfo update', trackInfo);
       this.trigger('trackinfo');
     }
 
+    // trackinfo may cause an abort
```
gesinger, line 1466: Minor: but might be worth mentioning one of the example cases

src/segment-loader.js
```diff
@@ -1678,7 +1697,7 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     const segmentInfo = this.pendingSegment_;
 
-    if (!segmentInfo || !this.startingMedia_) {
+    if (!segmentInfo || !segmentInfo.trackInfo) {
       // no segment to append any data for
```
gesinger, line 1701: Might be worth an update to this comment

PR #915, opened 2020-07-23
chore: fix tears of steal playready on the demo page

brandonocasey:
We put `com.widevine.alpha` instead of `com.microsoft.playready`


gkatsev:
Works for me.

PR #921, opened 2020-07-30
chore: playlist selector logging

brandonocasey:
## Description
I was using some logging for debugging the selector, but I think that it would be more broadly useful so I added it.


gkatsev:
Tests are failing due to the extra logging

PR #926, opened 2020-08-07
feat: support MPD.Location

gkatsev:
Depends on https://github.com/videojs/mpd-parser/pull/102
With it, I can see that we properly update the srcUrl internally based on the location, for example with a stream like https://livesim.dashif.org/livesim/startrel_-20/stoprel_20/timeoffset_0/testpic_2s/Manifest.mpd

TODO
- [x] Add some tests
- [x] Make sure that a more streams with Location that don't also do a 302 redirect on the main manifest function properly
- [x] update mpd-parser

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -678,7 +678,7 @@ export default class DashPlaylistLoader extends EventTarget {
   updateMainManifest_(updatedManifest) {
     this.master = updatedManifest;
 
-    // if locations isn't set or is an empty array, exits early
+    // if locations isn't set or is an empty array, exit early
     if (!this.master.locations || !this.master.locations.length) {
       return;
     }
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -629,12 +629,12 @@ export default class DashPlaylistLoader extends EventTarget {
     this.mediaRequest_ = null;
 
     if (!this.masterPlaylistLoader_) {
-      this.master = parseMasterXml({
+      this.updateMaster_(parseMasterXml({
         masterXml: this.masterXml_,
         srcUrl: this.srcUrl,
         clientOffset: this.clientOffset_,
         sidxMapping: this.sidxMapping_
-      });
+      }));
       // We have the master playlist at this point, so
       // trigger this to allow MasterPlaylistController
       // to make an initial playlist selection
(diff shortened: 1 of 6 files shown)


gkatsev:
Are we supposed to use Location as the new baseUrl as well, if a baseUrl isn't provided?
At the very least I think with the above stream and https://livesim.dashif.org/livesim/startrel_-60/stoprel_-20/timeoffset_0/testpic_2s/Manifest.mpd, we need to update the baseUrl based on the 302 redirect.

gkatsev:
Seems like handleManifestRedirect seems to make those streams work without requiring this change, so, we'll want to get another stream to test that we actually make subsequent requests appropriately.

gkatsev:
OK, I've confirmed this does the right thing and requests the updates from the MPD.Location. I did so by grabbing https://livesim.dashif.org/livesim/startrel_-60/stoprel_-20/timeoffset_0/testpic_2s/Manifest.mpd locally, adding an MPD.BaseURL with the same url as MPD.Location. I also added a minimumUpdatePeriod so that we request updates to the manifest. I can confirm that when MPD.Location is present, it'll make requests there instead of localhost.

gkatsev:
Tests will start passing once mpd-parser is updated and released.

src/dash-playlist-loader.js
```diff
@@ -670,6 +670,26 @@ export default class DashPlaylistLoader extends EventTarget {
...
+  /**
+   * Given a new manifest, update our pointer to it and update the srcUrl based on the location elements of the manifest, if they exist.
+   *
+   * @param {Object} updatedManifest the manifest to update to
+   */
+  updateMainManifest_(updatedManifest) {
+    this.master = updatedManifest;
+
+    // if locations isn't set or is an empty array, exist early
+    if (!this.master.locations || this.master.locations && !this.master.locations.length) {
```
brandonocasey, line 682: should this just be `!this.master.locations || !this.master.locastions.length`

src/dash-playlist-loader.js
```diff
@@ -670,6 +670,26 @@ export default class DashPlaylistLoader extends EventTarget {
...
 
+  /**
+   * Given a new manifest, update our pointer to it and update the srcUrl based on the location elements of the manifest, if they exist.
+   *
+   * @param {Object} updatedManifest the manifest to update to
+   */
+  updateMainManifest_(updatedManifest) {
+    this.master = updatedManifest;
+
+    // if locations isn't set or is an empty array, exist early
```
brandonocasey, line 681:
```suggestion
    // if locations isn't set or is an empty array, exits early
```

src/dash-playlist-loader.js
gkatsev: Let me check. I thought I tried this and it was showing an error but I may have tried it with `&&`?

gkatsev: yeah, I think I had `&&` locally. with `||` seems to work.

src/dash-playlist-loader.js
```diff
@@ -678,8 +678,8 @@ export default class DashPlaylistLoader extends EventTarget {
   updateMainManifest_(updatedManifest) {
     this.master = updatedManifest;
 
-    // if locations isn't set or is an empty array, exist early
-    if (!this.master.locations || this.master.locations && !this.master.locations.length) {
+    // if locations isn't set or is an empty array, exits early
```
gesinger, line 681:
```suggestion
    // if locations isn't set or is an empty array, exit early
```

PR #928, opened 2020-08-12
chore: update m3u8-parser to v4.4.3

gesinger:
## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6534,9 +6534,9 @@
       }
     },
     "m3u8-parser": {
-      "version": "4.4.2",
-      "resolved": "https://registry.npmjs.org/m3u8-parser/-/m3u8-parser-4.4.2.tgz",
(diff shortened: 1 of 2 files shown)


AwokeKnowing:
Thanks for looking into this!  We really need videojs with these updates as our FairPlay support depends on them.  what kind of eta would we be looking at until a videojs release?  or how would we go about getting a pre-release with this fix?  we use the release version of videojs. is it likely to be released by next week, or would we need to look into building it, or is there some 'pre release builds'?

AwokeKnowing:
@gkatsev  any update on this?  would feel much more comfortable using a 'custom/pre-realease videojs' if at least the fix is officially merged in http-streaming

gkatsev:
This is definitely going to make it in to a release.

PR #929, opened 2020-08-14
fix: use playlist NAME when available as its ID

gkatsev:
This is particularly used in DASH as the `NAME` is set from the Representation's `ID` which is specified to be unique per period and that across periods, the representations with the same ID should be functionally equivalent.
This is important because in DASH, across periods, it's possible for Representations to be re-ordered, and we want to make sure that the reference doesn't get messed up when the period updates.
It still sets up a reference to the playlist via the old name but if the ordering is changed it'll reference the incorrect playlist as it does now.

TODO
- [ ] verify specifically the change in ordering
- [x] update code if necessary
- [x] add tests if necessary

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -43,7 +43,7 @@ export const parseMasterXml = ({ masterXml, srcUrl, clientOffset, sidxMapping })
     sidxMapping
   });
 
-  addPropertiesToMaster(master, srcUrl, 'dash');
+  addPropertiesToMaster(master, srcUrl);
 
   return master;
 };
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -43,7 +43,7 @@ export const parseMasterXml = ({ masterXml, srcUrl, clientOffset, sidxMapping })
     sidxMapping
   });
 
-  addPropertiesToMaster(master, srcUrl, 'dash');
+  addPropertiesToMaster(master, srcUrl, true);
 
   return master;
 };
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -43,7 +43,7 @@ export const parseMasterXml = ({ masterXml, srcUrl, clientOffset, sidxMapping })
     sidxMapping
   });
 
-  addPropertiesToMaster(master, srcUrl);
+  addPropertiesToMaster(master, srcUrl, 'dash');
 
   return master;
 };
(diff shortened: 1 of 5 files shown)


gkatsev:
Added a test and briefly reverting the previous changes to show that the test fails. Once travis shows the failure I'll bring the changes back.

brandonocasey (approved):

src/manifest.js
```diff
@@ -100,17 +100,27 @@ export const setupMediaPlaylist = ({ playlist, uri, id }) => {
...
+    }
 
     setupMediaPlaylist({
       playlist,
-      id: createPlaylistID(i, playlist.uri)
+      id
     });
     playlist.resolvedUri = resolveUrl(master.uri, playlist.uri);
+    // make sure that if a DASH is used, the old "createId" id is also available
+    master.playlists[createId] = playlist;
```
gesinger, line 123: What do we keep this ID around for for DASH?

src/manifest.js
```diff
@@ -100,17 +100,27 @@ export const setupMediaPlaylist = ({ playlist, uri, id }) => {
  * @param {Object} master
  *        The master playlist
  */
-export const setupMediaPlaylists = (master) => {
+export const setupMediaPlaylists = (master, type) => {
```
gesinger, line 103: I think this function shouldn't consider the type itself, as that logic may belong more in the appropriate loader. Instead we can have something like `useNameForId` or something like that. JSDoc can also use an update.

src/manifest.js
```diff
@@ -100,17 +100,27 @@ export const setupMediaPlaylist = ({ playlist, uri, id }) => {
...
   let i = master.playlists.length;
 
   while (i--) {
     const playlist = master.playlists[i];
+    const createId = createPlaylistID(i, playlist.uri);
```
gesinger, line 108: Minor, but might read better as `createdId`

src/manifest.js
gkatsev: For backwards compatibility.

src/dash-playlist-loader.js
```diff
@@ -43,7 +43,7 @@ export const parseMasterXml = ({ masterXml, srcUrl, clientOffset, sidxMapping })
     sidxMapping
   });
 
-  addPropertiesToMaster(master, srcUrl);
+  addPropertiesToMaster(master, srcUrl, 'dash');
```
brandonocasey, line 46: I think this should be `true` instead of `'dash'`

gkatsev: oops, I made the change locally but apparently forgot to save.

brandonocasey (approved):

PR #934, opened 2020-08-20
Request variable missing in function call

psanchezg:
## Description
When setting the option handleManifestRedirects to true, this call fails because the req variable in the function definition is null

## Specific Changes proposed
Please, add the xhr variable at the function call.

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -194,7 +194,7 @@ export const filterChangedSidxMappings = (masterXml, srcUrl, clientOffset, oldSi
 export const requestSidx_ = (loader, sidxRange, playlist, xhr, options, finishProcessingFn) => {
   const sidxInfo = {
     // resolve the segment URL relative to the playlist
-    uri: resolveManifestRedirect(options.handleManifestRedirects, sidxRange.resolvedUri),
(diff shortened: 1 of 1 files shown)


src/dash-playlist-loader.js
```diff
@@ -194,7 +194,7 @@ export const filterChangedSidxMappings = (masterXml, srcUrl, clientOffset, oldSi
 export const requestSidx_ = (loader, sidxRange, playlist, xhr, options, finishProcessingFn) => {
   const sidxInfo = {
     // resolve the segment URL relative to the playlist
-    uri: resolveManifestRedirect(options.handleManifestRedirects, sidxRange.resolvedUri),
+    uri: resolveManifestRedirect(options.handleManifestRedirects, sidxRange.resolvedUri, xhr),
```
gkatsev, line 197: I believe the `xhr` here isn't actually the request but just a reference to the `xhr` method as defined in `xhr.js`.

gkatsev, line 197: also, `xhr` in here isn't the request that `resolveManifestRedirect` expects but the actual method we use to make XHRs.

gkatsev:
Since https://github.com/videojs/http-streaming/pull/927 is in, it takes care of this as well.

Issue #931, opened 2020-08-17
Player stalls when mediaIndex is less than 0 at live mode within segment-loader.js


suggestedfixes:
This occurs when the computed mediaIndex is negative in checkBuffer_ in segment-loader.js
https://github.com/videojs/http-streaming/blob/main/src/segment-loader.js#L1189
As a result, the player stalls. This is likely due to server-side expiring items not currently on the playlist.
Proposed fix:
```
if (mediaIndex !== null) {
  mediaIndex = Math.max(0, mediaIndex)
  ...
}
```

PR #936, opened 2020-08-21
rfc: Handle a stall case when moving window moves to fast, and mediaIndex …

suggestedfixes:
…becomes negative

## Description
Please describe the change as necessary.
If it's a bug fix, please link the issue that it fixes or describe the bug in as much detail.
Link to bug
https://github.com/videojs/http-streaming/issues/931
In summary, when the HLS window moves too fast, the mediaIndex could become negative.
The segment-loader ceases to fillBuffer_() because segmentInfo is null.

## Specific Changes proposed
At generateSegmentInfo_() in segment-loader.js, allow negative mediaIndex to go through, and return the segmentInfo of the first item on the playlist.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1261,7 +1261,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       return null;
     }
     // if media index is negative, it means the segment loader is trying to
-    // load an item expired (abs(mediaIndex) * segment duration) ago. So it might 
+    // load an item expired (abs(mediaIndex) * segment duration) ago. So it might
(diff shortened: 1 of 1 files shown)


gkatsev:
Hey, thanks for the PR. This is definitely an issue that we should investigate however this isn't the place to fix it. It may be that we want to clamp it in the method where we modify the mediaIndex, but we probably want to investigate it a bit more to figure out why this happens and see if we can put a better fix in place. I've re-opened #931 and I put it on our backlog to look at.

PR #937, opened 2020-08-22
rfc: Improve computing buffered time in segment-loader

suggestedfixes:
## Description
The current checkBuffer_() in segment-loader.js tends to over-estimate the amount buffered.
For example, if we have buffered:
[0 => 21, 79 => 80]
At currentTime == 20, checkBuffer_() would think it has buffered 80 - 20 == 60 seconds worth of data.
But the reality is, we have only buffered 2 seconds worth of data.

## Specific Changes proposed
Adding up individual intervals in buffered

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -109,21 +109,22 @@ const segmentInfoString = (segmentInfo) => {
 };
 
 /**
- * Returns the buffered time beyond current time 
+ * Returns the buffered time beyond current time
  *
  * @param {TimeRange[]} buffered
  *        The loader's buffer
- * @param {number} currentTime 
- *        The player's currentTime 
+ * @param {number} currentTime
+ *        The player's currentTime
  *
  * @return {number}
- *        A number representing the amount buffered 
+ *        A number representing the amount buffered
  */
 const bufferedTimeBeyondCurrentTime = (buffered, currentTime) => {
   if (!buffered || !buffered.length || isNaN(currentTime)) {
     return 0;
   }
   let totalBufferedTime = 0;
+
   for (let i = 0; i < buffered.length; i++) {
     // case where currentTime falls within the buffered interval
     if (buffered.start(i) <= currentTime && currentTime <= buffered.end(i)) {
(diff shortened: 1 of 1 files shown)


suggestedfixes:
@gkatsev 
Comments?
Tips on Unit Test?

gkatsev:
Do you have an example of when this occurs? Theoretically, during seeks, which is the likeliest scenario for multiple buffered regions, we clear the buffer.
We've looked at this briefly and the issue you've identified is definitely something that needs fixing. However, it may be more complicated that just changing `checkBuffer`. Specifically, if the media index thinks we're at the 79-80 buffered region, but we're actually at the 0-21 region, even if `checkBuffer` checks correctly, we may still end up downloading at the wrong place which will cause more issues.

suggestedfixes:
We have playlists consists of a lot of discontinuity markers and large gaps between timestamps playing at live mode.
In other words, we are historical motion data streaming at live mode.

For intervals that are continuous, the buffering works perfectly. Somehow, the irregularity in timestamps and discontinuity caused multiple intervals to buffer and incorrect buffered time calculated. On rare occasions, the checkBuffer_ will return a null, and cause the player to stall indefinitely.

suggestedfixes:
@gkatsev
"Specifically, if the media index thinks we're at the 79-80 buffered region, but we're actually at the 0-21 region, even if checkBuffer checks correctly, we may still end up downloading at the wrong place which will cause more issues."

So in theory, if there are discontinuities when streaming intervals of discontinuous timestamps, [0 => 21, 79 => 80, ...] should not occur, what instead should happen is the amount buffered should be [0 => 22]? We should aim to figure out the place where this merge should have occurred?

gkatsev:
Yeah, discontinuities should theoretically not show visible buffer gaps and playback should be smooth.

gkatsev:
I'm going to close this PR as we won't be merging it as is. We should probably create an issue to track this, though.

PR #938, opened 2020-08-24
fix: Throw a player error when trying to play DRM content without eme

brandonocasey:
## Description
I run into issues where I am trying to play drm content but do not have the eme plugin. This causes the video buffer to increase but playback will never start because drm was never initialized. I propose that we trigger a player error in this case. If we don't want to go that far though we can just make this a warning.

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -243,13 +243,12 @@ const setupEmeOptions = ({
   audioMedia,
   mainPlaylists
 }) => {
-  if (!player.eme) {
-    return;
-  }
-
   const sourceOptions = emeKeySystems(sourceKeySystems, media, audioMedia);
 
-  if (!sourceOptions) {
+  if (!player.eme || !sourceOptions) {
+    if (sourceOptions) {
+      player.error({code: 5, message: 'DRM encrypted source cannot be decrypted without DRM plugin'});
+    }
     return;
   }
 
@@ -825,7 +824,8 @@ class VhsHandler extends Component {
 
       setupEmeOptions({
         player: this.player_,
-        sourceKeySystems: this.source_.keySystems,
+        // eslint-disable-next-line
+        sourceKeySystems: this.source_.keySystems || this.source_.key_systems,
         media: this.playlists.media(),
         audioMedia: audioPlaylistLoader && audioPlaylistLoader.media(),
         mainPlaylists: this.playlists.master.playlists
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -247,8 +247,7 @@ const setupEmeOptions = ({
 
   if (!player.eme || !sourceOptions) {
     if (sourceOptions) {
-      videojs.log.error('DRM encrypted source cannot be decrypted without a DRM plugin');
-      player.error({code: 5});
+      videojs.log.warn('DRM encrypted source cannot be decrypted without a DRM plugin');
     }
     return;
   }
@@ -825,8 +824,7 @@ class VhsHandler extends Component {
 
       setupEmeOptions({
         player: this.player_,
-        // eslint-disable-next-line
-        sourceKeySystems: this.source_.keySystems || this.source_.key_systems,
+        sourceKeySystems: this.source_.keySystems,
         media: this.playlists.media(),
         audioMedia: audioPlaylistLoader && audioPlaylistLoader.media(),
         mainPlaylists: this.playlists.master.playlists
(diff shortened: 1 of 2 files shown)


src/videojs-http-streaming.js
```diff
@@ -825,7 +824,8 @@ class VhsHandler extends Component {
 
       setupEmeOptions({
         player: this.player_,
-        sourceKeySystems: this.source_.keySystems,
+        // eslint-disable-next-line
+        sourceKeySystems: this.source_.keySystems || this.source_.key_systems,
```
brandonocasey, line 828: eme appears to clone `key_systems` into `keySystems`, but without the eme plugin that won't happen.

src/videojs-http-streaming.js
```diff
@@ -136,7 +136,7 @@ const emeKeySystems = (keySystemOptions, videoPlaylist, audioPlaylist) => {
     audio: audioPlaylist && audioPlaylist.attributes && audioPlaylist.attributes.CODECS
   };
 
-  if (!codecs.audio && codecs.video.split(',').length > 1) {
```
brandonocasey, line 139: we can currently try to call split on a codec value that doesn't exist.

src/videojs-http-streaming.js
gkatsev: I think we can ignore `key_systems` here as contrib-eme only expects `keySystems`.

src/videojs-http-streaming.js
```diff
@@ -243,13 +243,12 @@ const setupEmeOptions = ({
...
-  if (!player.eme) {
-    return;
-  }
-
   const sourceOptions = emeKeySystems(sourceKeySystems, media, audioMedia);
 
-  if (!sourceOptions) {
+  if (!player.eme || !sourceOptions) {
+    if (sourceOptions) {
+      player.error({code: 5, message: 'DRM encrypted source cannot be decrypted without DRM plugin'});
```
gkatsev, line 250: we probably don't want to change the message and just log this in the console. Since this message is user-facing, it'll mean we'll have to wordsmith it and then translate it.

src/videojs-http-streaming.js
brandonocasey: Without eme though we won't pass any options in and it won't seem like a drm source at all. That means that the check I added in `setupEmeOptions` won't happen.

gkatsev (dismissed):

test/videojs-http-streaming.test.js
```diff
@@ -5765,16 +5765,38 @@ QUnit.module('setupEmeOptions', {
...
+  let logWarn;
+  const origWarn = videojs.log.warn;
+
+  videojs.log.warn = (line) => {
+    logWarn = line;
+  };
+
   setupEmeOptions({ player, sourceKeySystems, media, audioMedia, mainPlaylists });
 
   assert.ok(true, 'no exception');
```
gesinger, line 5796:
```suggestion
```

test/videojs-http-streaming.test.js
```diff
@@ -5765,16 +5765,38 @@ QUnit.module('setupEmeOptions', {
...
+  const media = {};
+  const audioMedia = {};
+  const mainPlaylists = [];
+
+  setupEmeOptions({ player, sourceKeySystems, media, audioMedia, mainPlaylists });
+
+  assert.ok(true, 'no exception');
+});
+
+QUnit.test('player and log error if no eme and we have key systems', function(assert) {
```
gesinger, line 5780:
```suggestion
QUnit.test('log error if no eme and we have key systems', function(assert) {
```

src/videojs-http-streaming.js
```diff
@@ -243,13 +243,12 @@ const setupEmeOptions = ({
...
   mainPlaylists
 }) => {
-  if (!player.eme) {
-    return;
-  }
-
   const sourceOptions = emeKeySystems(sourceKeySystems, media, audioMedia);
 
-  if (!sourceOptions) {
+  if (!player.eme || !sourceOptions) {
```
gesinger, line 248: Instead of running through `emeKeySystems ` should we just check for `sourceKeySystems`?

gesinger (dismissed):
Minor comment, but looks good.

src/videojs-http-streaming.js
gkatsev: wouldn't checking sourceKeySystems mean that configuring eme directly with key system options not work again?

gesinger: If it's configured directly we don't have to worry about configuring it here.

And checking sourceOptions here is the same as checking sourceKeySystems, since if sourceKeysSytems is falsey its returned from emeKeySystems.

gkatsev (approved):

PR #942, opened 2020-09-03
feat: Update minimumUpdatePeriod handling

alex-barstow:
## Description
This is part of a set of upcoming changes to add support for live DASH playback. Specifically, this PR differentiates the handling of 2 cases which were formerly conflated:

1. The `MPD@minimumUpdatePeriod` attribute has a _value of 0_, indicating that the MPD has no validity after the moment it was retrieved.
2. The `MPD@minimumUpdatePeriod` attribute is _absent_, indicating the MPD has infinite validity and will never be updated

## Specific Changes proposed
- Update [`mpd-parser`](https://github.com/videojs/mpd-parser/pull/103) to allow us to identify case 2
- Make sure `updateMaster()` checks if a manifest's `minimumUpdatePeriod` value changes, and return the new manifest if so.
- Add handling for case 1: minimumUpdateperiod is 0 
  - Continue refreshing manifests every `targetDuration`, per existing [`TODO`](https://github.com/videojs/http-streaming/compare/update-minimumUpdatePeriod-handling?expand=1#diff-ca1029eb342915b65f2b261194fef835L660-L665)
- Add handling for case 2: minimumUpdatePeriod is absent
  - Do not trigger a `'minimumUpdatePeriod'` event as there is no need to refresh the manifest
- Add/update tests

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -94,6 +94,10 @@ export const updateMaster = (oldMaster, newMaster) => {
     }
   });
 
+  if (newMaster.minimumUpdatePeriod !== oldMaster.minimumUpdatePeriod) {
+    noChanges = false;
+  }
+
   if (noChanges) {
     return null;
   }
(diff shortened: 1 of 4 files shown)


gkatsev (dismissed):

src/dash-playlist-loader.js
```diff
@@ -766,10 +764,11 @@ export default class DashPlaylistLoader extends EventTarget {
 
                 // Clear & reset timeout with new minimumUpdatePeriod
```
gesinger, line 765: It may be worth created a `updateMinimumUpdatePeriodTimeout_()` function to centralize the logic of the three uses (even can include the clearing) and add a comment around why we use the target duration when the minimumUpdatePeriod is 0.

alex-barstow: Yup makes sense. Done

src/dash-playlist-loader.js
```diff
@@ -646,6 +650,16 @@ export default class DashPlaylistLoader extends EventTarget {
     }
   }
 
+  updateMinimumUpdatePeriodTimeout_() {
+    const minimumUpdatePeriod = this.master && this.master.minimumUpdatePeriod;
+
+    if (minimumUpdatePeriod >= 0) {
+      this.minimumUpdatePeriodTimeout_ = window.setTimeout(() => {
+        this.trigger('minimumUpdatePeriod');
+      }, minimumUpdatePeriod || this.media().targetDuration * 1000);
```
gesinger, line 659: We should add a comment here just saying why the default is target duration.

src/dash-playlist-loader.js
```diff
@@ -646,6 +650,16 @@ export default class DashPlaylistLoader extends EventTarget {
     }
   }
 
+  updateMinimumUpdatePeriodTimeout_() {
```
gesinger, line 653: It would also make sense to add the clearing of the old one here too.

gkatsev (approved):

PR #948, opened 2020-09-09
fix: detect demuxed video underflow gaps

brandonocasey:
## Description
We are seeing a video underflow for demuxed fmp4 content. I think the main issue is timing related, but I have yet to narrow down the cause. While going through our code I noticed that we already detect video underflows for **muxed** content. I added code to detect **demuxed** video underflows so that we can fix those as well.

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -443,10 +443,15 @@ export default class PlaybackWatcher {
       return true;
     }
 
+    const sourceUpdater = this.tech_.vhs.masterPlaylistController_.sourceUpdater_;
     const buffered = this.tech_.buffered();
-    const nextRange = Ranges.findNextRange(buffered, currentTime);
+    const videoUnderflow = this.videoUnderflow_({
+      audioBuffered: sourceUpdater.audioBuffer && sourceUpdater.audioBuffered() || null,
+      videoBuffered: sourceUpdater.videoBuffer && sourceUpdater.videoBuffered() || null,
+      currentTime
+    });
 
-    if (this.videoUnderflow_(nextRange, buffered, currentTime)) {
+    if (videoUnderflow) {
       // Even though the video underflowed and was stuck in a gap, the audio overplayed
       // the gap, leading currentTime into a buffered range. Seeking to currentTime
       // allows the video to catch up to the audio position without losing any audio
@@ -459,6 +464,7 @@ export default class PlaybackWatcher {
       this.tech_.trigger({type: 'usage', name: 'hls-video-underflow'});
       return true;
     }
+    const nextRange = Ranges.findNextRange(buffered, currentTime);
 
     // check for gap
     if (nextRange.length > 0) {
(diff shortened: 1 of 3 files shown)


src/playback-watcher.js
```diff
@@ -512,18 +518,36 @@ export default class PlaybackWatcher {
...
-  videoUnderflow_(nextRange, buffered, currentTime) {
-    if (nextRange.length === 0) {
+  videoUnderflow_({videoBuffered, audioBuffered, currentTime}) {
+    if (!videoBuffered) {
+      return;
+    }
+    let gap;
+
+    if (videoBuffered.length && audioBuffered.length) {
+      const lastVideoRange = Ranges.findRange(videoBuffered, currentTime - 3);
```
gkatsev, line 528: why the `-3`?

src/playback-watcher.js
```diff
@@ -512,18 +518,36 @@ export default class PlaybackWatcher {
...
+      if (audioRange.length && !videoRange.length && lastVideoRange.length) {
+        gap = {start: lastVideoRange.end(0), end: audioRange.end(0)};
+      }
+
+    } else {
+      const nextRange = Ranges.findNextRange(audioBuffered, currentTime);
+
       // Even if there is no available next range, there is still a possibility we are
       // stuck in a gap due to video underflow.
-      const gap = this.gapFromVideoUnderflow_(buffered, currentTime);
```
gkatsev, line 519: and this generic buffered gap check isn't necessary because we still check for this gap above which may not be a video underflow? Or is this equivalent to checking against videoBuffered?

src/playback-watcher.js
brandonocasey: `gapFromVideoUnderflow` implies that audio will keep playing for ~3 seconds during a video underflow and that is what I have seen as well. I will add a comment.

src/playback-watcher.js
brandonocasey: The generic `buffered` only worked when the it was referring to muxed content because we will go off of the "video buffer", which is really the muxed content, in that case. Now that we have a case that checks for demuxed content we can just use videoBuffered for the original muxed check and both buffers for the demuxed check.

test/test-helpers.js
```diff
@@ -183,13 +183,24 @@ export const useFakeEnvironment = function(assert) {
...
-              return args.join(', ');
+              return args.reduce((acc, val) => {
+                if (acc) {
+                  acc += ', ';
+                }
+
+                acc += val;
+
+                if (val.stack) {
+                  acc += '\n' + val.stack;
```
brandonocasey, line 194: actually print out log error stack traces when we assert.

src/playback-watcher.js
```diff
@@ -512,18 +518,42 @@ export default class PlaybackWatcher {
...
+  videoUnderflow_({videoBuffered, audioBuffered, currentTime}) {
+    // audio only content will not have video underflor :)
+    if (!videoBuffered) {
+      return;
+    }
+    let gap;
+
+    // find a gap in demuxed content.
+    if (videoBuffered.length && audioBuffered.length) {
+      // in Chrome audio will continue to play for ~3 when we run out of video
```
gesinger, line 530:
```suggestion
      // in Chrome audio will continue to play for ~3s when we run out of video
```

src/playback-watcher.js
gesinger: Should we add a fudge to it?

test/playback-watcher.test.js
```diff
@@ -188,7 +239,17 @@ QUnit.test('skips over gap in Chrome due to video underflow', function(assert) {
   assert.equal(vhsVideoUnderflowEvents, 0, 'no video underflow event got triggered');
   assert.equal(hlsVideoUnderflowEvents, 0, 'no video underflow event got triggered');
 
-  this.player.currentTime(13);
+  const mpc = this.player.tech_.vhs.masterPlaylistController_;
+
+  mpc.sourceUpdater_.videoBuffered = () => {
+    return videojs.createTimeRanges([[0, 10], [10, 15]]);
```
gesinger, line 245: This (and below): should they be one range? Or meant to have a gap here?

src/playback-watcher.js
brandonocasey: I'm thinking that we shouldn't as 3 is already a "fudge" here.

test/playback-watcher.test.js
brandonocasey: I just copied from the other test, but I don't think we need a gap here. removing it.

gkatsev (approved):

PR #949, opened 2020-09-10
fix: dash manifest not refreshed if only some playlists are updated

alex-barstow:
## Description
This fixes an issue where an MPD update can be missed if some media playlists are updated while the last playlist remains unchanged (see [here](https://github.com/videojs/http-streaming/blob/master/src/dash-playlist-loader.js#L71-L80))
## Specific Changes proposed
In `updateMaster()`, update `noChanges` to default to `true`, and set it to `false` if any playlist is changed.

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -61,7 +61,7 @@ export const parseMasterXml = ({ masterXml, srcUrl, clientOffset, sidxMapping })
  *         playlists merged in
  */
 export const updateMaster = (oldMaster, newMaster) => {
-  let noChanges;
+  let noChanges = true;
   let update = mergeOptions(oldMaster, {
     // These are top level properties that can be updated
     duration: newMaster.duration,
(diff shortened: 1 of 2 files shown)


src/dash-playlist-loader.js
```diff
@@ -61,7 +61,7 @@ export const parseMasterXml = ({ masterXml, srcUrl, clientOffset, sidxMapping })
  *         playlists merged in
  */
 export const updateMaster = (oldMaster, newMaster) => {
```
gkatsev, line 63: one thing that's missing that we need to account for potentially is that since we are looping through the newMaster's playlists, if an item was removed we wouldn't notice it as a change.
Or are we waiting to make that change until we have a better understanding of when and how periods can change?

alex-barstow: Since MPD updates can only add/remove representations by adding/removing periods, I think we should wait to see how period changes will be handled, as it may not be necessary to compare the representations themselves in those cases.

gkatsev: Sounds good!

gkatsev (approved):
Do we want to merge this in now and then have a separate PR for the period stuff if necessary?

gkatsev:
Looks like with https://github.com/videojs/http-streaming/pull/942, either that or this PR may need to be updated slightly.

gesinger (approved):

PR #950, opened 2020-09-10
fix: report the correct buffered regardless of playlist change

brandonocasey:
## Description
OK so this bug was a doozy to try to understand. What happens with one of our demuxed videos is:

1. We fetch and append segment 0 for video and audio segment loader
2. We fetch and append segment 1 for video and audio segment loader
3. Rendition switch happens for video
4. We fetch and append segment 2 for audio
5. We fetch and append segment 0 again for video because:
    * `buffered_()` returns an empty buffered `TimeRange` as we don't have `startingMedia_` yet.
    * This causes us to grab content at `0` again as current time is within that range and buffered has no length. (see `SegmentLoader#checkBuffer_`)
6. We video underflow and the content freezes as the invalid append does not fail, but also never `updateend`
  

## Possible Solutions
1. The current one: store what we call `startingMedia` in a variable that is only ever updated, not deleted. 
2. Return buffered based on the `loaderType` and various sourceUpdater checks if `startingMedia` does not exist.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -632,12 +632,12 @@ export default class SegmentLoader extends videojs.EventTarget {
    *         TimeRange object representing the current buffered ranges
    */
   buffered_() {
-    if (!this.sourceUpdater_ || !this.startingMedia_) {
+    if (!this.sourceUpdater_ || !this.mediaConfig_) {
       return videojs.createTimeRanges();
     }
 
     if (this.loaderType_ === 'main') {
-      const { hasAudio, hasVideo, isMuxed } = this.startingMedia_;
+      const { hasAudio, hasVideo, isMuxed } = this.mediaConfig_;
 
       if (hasVideo && hasAudio && !this.audioDisabled_ && !isMuxed) {
         return this.sourceUpdater_.buffered();
@@ -1089,12 +1089,6 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
-    if (segmentInfo.mediaIndex === this.playlist_.segments.length - 1 &&
-        this.mediaSource_.readyState === 'ended' &&
-        !this.seeking_()) {
-      return;
-    }
-
     segmentInfo.timestampOffset = timestampOffsetForSegment({
       segmentTimeline: segmentInfo.timeline,
       currentTimeline: this.currentTimeline_,
@@ -1150,9 +1144,8 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @param {Object} syncPoint - a segment info object that describes the
    * @return {Object} a segment request object that describes the segment to load
    */
-  checkBuffer_(buffered, playlist, mediaIndex, hasPlayed, currentTime, syncPoint) {
+  checkBuffer_(buffered, playlist, currentMediaIndex, hasPlayed, currentTime, syncPoint) {
     let lastBufferedEnd = 0;
-    let startOfSegment;
 
     if (buffered.length) {
       lastBufferedEnd = buffered.end(buffered.length - 1);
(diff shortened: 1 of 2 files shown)


brandonocasey:
I think that this issue would have been a lot easier to find if we had more logging in `checkBuffer_` about **why** a segment is going to be fetched. Perhaps I will work on adding that in this pull request.

gkatsev:
Looks like nulling-out starting media was done as part of the codec switching work? I'm not sure this would explain why the stream stopped working in 1.x of VHS too.

However, if it works, this seems like a simple solution. (Also, can confirm that it plays through fine in this PR).

src/segment-loader.js
```diff
@@ -1089,12 +1089,6 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
-    if (segmentInfo.mediaIndex === this.playlist_.segments.length - 1 &&
```
brandonocasey, line 1092: This moved to checkBuffer_ so that we don't log about requesting segmentInfo when we wont be. It also seems like it should belong there.

src/segment-loader.js
```diff
@@ -1176,31 +1169,33 @@ export default class SegmentLoader extends videojs.EventTarget {
...
 
+    let segmentMediaIndex = null;
+    let startOfSegment;
+    let syncRequest = false;
+
     // When the syncPoint is null, there is no way of determining a good
     // conservative segment index to fetch from
     // The best thing to do here is to get the kind of sync-point data by
     // making a request
     if (syncPoint === null) {
```
brandonocasey, line 1180: Made this one big if/else so that we can log all of the data used to generate segment info at the end.

src/segment-loader.js
```diff
@@ -1176,31 +1169,33 @@ export default class SegmentLoader extends videojs.EventTarget {
       return null;
     }
 
+    let segmentMediaIndex = null;
+    let startOfSegment;
+    let syncRequest = false;
```
gesinger, line 1174: minor: isSyncRequest

src/segment-loader.js
```diff
@@ -1176,31 +1169,33 @@ export default class SegmentLoader extends videojs.EventTarget {
       return null;
     }
 
+    let segmentMediaIndex = null;
```
gesinger, line 1172: I think it may be helpful to distinguish between `mediaIndex` and `segmentMediaIndex`. Maybe `mediaIndex` can be `currentMediaIndex` and `segmentMediaIndex` can be `mediaIndex`, or `segmentMediaIndex` can be something like `nextMediaIndex`.

src/segment-loader.js
```diff
@@ -1220,11 +1215,35 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    const segmentInfo = this.generateSegmentInfo_(playlist, segmentMediaIndex, startOfSegment, syncRequest);
+
+    if (!segmentInfo) {
+      return;
+    }
+
+    // if this is the last segment in the playlist
+    // we are not seeking and end of stream has already been called
+    // do not re-request
+    if (this.mediaSource_ && this.playlist_ && segmentInfo.mediaIndex === this.playlist_.segments.length - 1 &&
```
gesinger, line 1231: Is it possible for `this.playlist_` to be falsey here?

src/segment-loader.js
```diff
@@ -1220,11 +1215,35 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        this.mediaSource_.readyState === 'ended' &&
+        !this.seeking_()) {
+      return;
+    }
+
+    this.logger_(`requesting ${segmentInfo.uri}`, {
```
gesinger, line 1237: Since it's possible the request will be delayed/won't happen, it might be more accurate to log something like `buffer check returning` or something like that.

src/segment-loader.js
```diff
@@ -1458,6 +1477,7 @@ export default class SegmentLoader extends videojs.EventTarget {
         video: true
       };
 
+      this.mediaConfig_ = trackInfo;
       this.startingMedia_ = trackInfo;
```
gesinger, line 1481: I think it may be better to have more descriptive names here, since technically `startingMedia` is the `currentMediaConfig` and `mediaConfig` is the `startingMediaConfig`.

gesinger (commented):
Would be good to add a test if possible.

src/segment-loader.js
brandonocasey: in the tests it happens quite often, not sure if it will happen in the wild.

src/segment-loader.js
brandonocasey: I'm hesitant to make such a change, at least in this pull request as that will require a lot of code changes in many different locations, perhaps we can make such a change in another pull request.

gesinger (approved):
If it isn't too hard to add a test for the case, that would be helpful, but otherwise looks good.

gkatsev (dismissed):

gkatsev (approved):

PR #952, opened 2020-09-17
fix: audio groups with the same uri as media do not count

brandonocasey:
## Description
When An audio group has the same uri as a main playlist, don't count it as an audio group. 

The code that I am using is from [this commit](https://github.com/videojs/http-streaming/commit/3cefde2c66ae7bfac8086b45d443502960bba844#diff-46e43d417013a9cfb72f6b3210979b63)

[I removed this code early on the the VHS 2 process](https://github.com/videojs/http-streaming/pull/469) as it was causing other issues, but after digging up some demuxed live streams I no longer see issues with live stream timeouts.

I created a source that replicates this issue and uploaded it next to some of our other test content. I also tested with a few of the demuxed live streams that we have and they all appear to work. It is likely that the live stream startup issues that I was seeing were fixed somewhere else.

### DIFF
--- a/src/media-groups.js
+++ b/src/media-groups.js
@@ -402,11 +402,8 @@ export const initialize = {
         return playlist.attributes[type] === groupId;
       });
 
-      // List of playlists that have an AUDIO attribute value matching the current
-      // group ID
       for (const variantLabel in mediaGroups[type][groupId]) {
         let properties = mediaGroups[type][groupId][variantLabel];
-        let playlistLoader;
 
         // List of playlists for the current group ID that have a matching uri with
         // this alternate audio variant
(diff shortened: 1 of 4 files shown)


src/media-groups.js
```diff
@@ -398,11 +398,31 @@ export const initialize = {
 
       // List of playlists that have an AUDIO attribute value matching the current
       // group ID
+      const groupPlaylists = playlists.filter(playlist => {
+        return playlist.attributes[type] === groupId;
+      });
 
+      // List of playlists that have an AUDIO attribute value matching the current
+      // group ID
```
gesinger, line 406: Looks like the comment got doubled

```suggestion
```

src/media-groups.js
```diff
@@ -398,11 +398,31 @@ export const initialize = {
...
       for (const variantLabel in mediaGroups[type][groupId]) {
         let properties = mediaGroups[type][groupId][variantLabel];
         let playlistLoader;
```
gesinger, line 409: Minor, but can move this down.

gesinger (approved):

gkatsev (approved):

PR #953, opened 2020-09-21
Chore: rename mediaConfig_ and startingMedia_

brandonocasey:
mediaConfig_ and startingMedia_ are fairly poor names for variables, we should do better and rename them to something that can be easily understood. This pull request aims to do that. Originally this work would have been done as a part of #950 but this change might have a bigger impact than intended.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -808,8 +808,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     if (this.mediaTypes_.AUDIO.activePlaylistLoader) {
       // if the audio playlist loader exists, then alternate audio is active
-      if (!this.mainSegmentLoader_.startingMedia_ ||
-          this.mainSegmentLoader_.startingMedia_.hasVideo) {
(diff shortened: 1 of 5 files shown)


src/master-playlist-controller.js
```diff
@@ -808,8 +808,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     if (this.mediaTypes_.AUDIO.activePlaylistLoader) {
       // if the audio playlist loader exists, then alternate audio is active
-      if (!this.mainSegmentLoader_.startingMedia_ ||
-          this.mainSegmentLoader_.startingMedia_.hasVideo) {
+      if (!this.mainSegmentLoader_.startingMediaInfo_ ||
```
brandonocasey, line 811: I think we want to using `startingMediaInfo` here as we want to call end of stream regardless of the state of `currentMediaInfo`. This is how it worked prior to the codec switching changes. It's possible that this isn't an issue though and we should just keep it as `currentMediaInfo`

gkatsev (approved):

PR #954, opened 2020-09-24
fix: MPD not refreshed if minimumUpdatePeriod is 0

alex-barstow:
## Description
This fixes a bug caused by an oversight in #942. If the `MPD@minimumUpdatePeriod` is 0 in the first dynamic MPD we load, we should use the media's target duration as the MPD refresh interval, however we need to wait until a playlist has been selected before we can know `DashPlaylistLoader.media().targetDuration`, otherwise we get a TypeError.

## Specific Changes proposed
If the `minimumUpdatePeriod` is 0 and we haven't loaded a playlist yet, wait for `'loadedplaylist'` to setup the `minimumUpdatePeriodTimeout_`

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -653,15 +653,31 @@ export default class DashPlaylistLoader extends EventTarget {
     // Clear existing timeout
     window.clearTimeout(this.minimumUpdatePeriodTimeout_);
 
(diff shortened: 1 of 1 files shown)


gkatsev:
Confirmed fixed with https://livesim.dashif.org/livesim/mup_0/testpic_2s/Manifest.mpd

gkatsev (approved):

PR #962, opened 2020-09-29
fix: filter out unsupported subtitles for dash

brandonocasey:
## Description
In the newest version of mpd-parser `0.13.0` we added support for exposing subtitle codecs so that we could filter out any that we do not support. For now I have it filtering out all subtitles that have codecs, but we may want to exclude only known codecs. This behavior can be seen with `Tears of Steal - Widevine (Unified Streaming)` as the subs there are `stpp.ttml.im1t`

Depends upon updates in:
* [x] [`mpd-parser`](https://github.com/videojs/mpd-parser/pull/106)
* [x] [`@videojs/vhs-utils`](https://github.com/videojs/vhs-utils/pull/19)


src/media-groups.js
```diff
@@ -522,12 +526,17 @@ export const initialize = {
 
         let properties = mediaGroups[type][groupId][variantLabel];
 
+        // TODO: we should filter for hls as well.
```
brandonocasey, line 529: We should probably do the same of HLS playlists, but we don't have a good way to do that right now.

src/media-groups.js
```diff
@@ -522,12 +526,17 @@ export const initialize = {
...
+        // Filter out unsupport playlists.
+        if (properties.playlists) {
+          properties.playlists = properties.playlists.filter((p) => !p.attributes || !p.attributes.CODECS);
```
gkatsev, line 532: I think we should explicitly filter the im1t codec rather than anything that has codecs set because it's possible that someone has the vtt codec set (unlikely, but possible)

src/media-groups.js
gkatsev: Looks like HLS does the same. The codec param for the subs playlist will indicate `stpp.ttml.im1t` From https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-07.html#section-4.4.6.2

      CODECS

      ...

      Note that if a Variant Stream specifies one or more Renditions
      that include IMSC subtitles, the CODECS attribute MUST indicate
      this with a format identifier such as "stpp.ttml.im1t".

      Every EXT-X-STREAM-INF tag SHOULD include a CODECS attribute.

brandonocasey:
I reworked this pull request to work for dash/hls using more standard functions in master-playlist-controller. This pull request now depends upon: https://github.com/videojs/vhs-utils/pull/19

gkatsev:
I think this needs to be rebased

gkatsev:
Looks like this needs an update to mpd-parser?

gkatsev (approved):

PR #963, opened 2020-09-30
use and expose the new max buffer low water line

gkatsev:


src/videojs-http-streaming.js
```diff
@@ -60,6 +60,7 @@ const Vhs = {
   'GOAL_BUFFER_LENGTH_RATE',
   'BUFFER_LOW_WATER_LINE',
   'MAX_BUFFER_LOW_WATER_LINE',
+  'MAX_BUFFER_LOW_WATER_LINE_NEW',
```
gesinger, line 63: It may be worth a comment somewhere explaining more what this is (or a different name instead of the _NEW suffix).

brandonocasey: It's in the other pr https://github.com/videojs/http-streaming/pull/886/files#diff-9d9a6cd82f41984872a66a3ab0d440c4R15

gkatsev: yeah, I think eventually we should probably rename it to like `EXPERIMENTAL_LOW_WATER_LINE` or something like that but that can be done later.

brandonocasey (approved):

PR #965, opened 2020-10-01
fix: abort all loaders on earlyabort

### LINKED ISSUE #964: experimentalBufferBasedABR (and maybe regular ABR too): fixup ABR when no progress events from XHR
When we get really bad network for a long time and download of gear5 fails, we end up running out of buffer because our downswitch didn't run because last check happened when we had plenty of buffer and we no longer earlyabort

brandonocasey:
During low bandwidth situations continuing to download any segments over the video/audio segment is ill-advised as we need to get the buffer filled again ASAP. So when we get an early abort we have to abort all segment loaders to make way for the new playlist that is about to load, so that it can have priority.

Fixes #964

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -12,7 +12,7 @@ export default {
   BUFFER_LOW_WATER_LINE: 0,
   MAX_BUFFER_LOW_WATER_LINE: 30,
 
-  // TODO: Remove this when useBufferWaterLines is removed
+  // TODO: Remove this when experimentalBufferBasedABR is removed
   EXPERIMENTAL_MAX_BUFFER_LOW_WATER_LINE: 16,
 
   BUFFER_LOW_WATER_LINE_RATE: 1,
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -643,13 +643,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
         return;
       }
 
-      this.mainSegmentLoader_.abort();
-      if (this.mediaTypes_.AUDIO.activePlaylistLoader) {
-        this.audioSegmentLoader_.abort();
-      }
-      if (this.mediaTypes_.SUBTITLES.activePlaylistLoader) {
-        this.subtitleSegmentLoader_.abort();
-      }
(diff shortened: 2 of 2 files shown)


src/config.js
```diff
@@ -12,7 +12,7 @@ export default {
   BUFFER_LOW_WATER_LINE: 0,
   MAX_BUFFER_LOW_WATER_LINE: 30,
 
-  // TODO: Remove this when useBufferWaterLines is removed
+  // TODO: Remove this when experimentalBufferBasedABR is removed
```
brandonocasey, line 15: so that we can remove this comments/code when we need to later on.

gkatsev (dismissed):

gkatsev:
Is this something that we think is still necessary? Or could just be closed because we're not doing any earlyabort in the new ABR stuff?

src/master-playlist-controller.js
```diff
@@ -617,11 +619,24 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+        abortSeconds = Config.EXPERIMENTAL_MAX_BUFFER_LOW_WATER_LINE;
       }
+
+      this.mainSegmentLoader_.abort();
+      if (this.mediaTypes_.AUDIO.activePlaylistLoader) {
+        this.audioSegmentLoader_.abort();
+      }
+      if (this.mediaTypes_.SUBTITLES.activePlaylistLoader) {
+        this.subtitleSegmentLoader_.abort();
+      }
```
gesinger, line 634: Should this use `delegateLoaders_`?

src/master-playlist-controller.js
```diff
@@ -599,6 +599,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('earlyabort', (event) => {
+      let abortSeconds = ABORT_EARLY_BLACKLIST_SECONDS;
```
gesinger, line 602: Minor, but this may make more sense as `blacklistSeconds`

gesinger (approved):

PR #966, opened 2020-10-01
fix: sustained bad network shouldn't pause

gkatsev:
When playing back bipbop with experimentalBufferBasedABR turned on and
we switch to Slow 3G in chrome, we may time out.
This is because
- earlyabort is turned off
- our last attempt at downswitching occurs when our forward buffer is
too full

To mitigate this, when we get an early abort set up a timer for 10
seconds to see if earlyabort conditions are met again. If a progress
event happens in the meantime, the earlyabort timer will be cancelled.

TODO:
- [ ] Make sure that the `setTimeout` is cancelled properly if the player is disposed (not sure if we have a `player.setTimeout()` like method
- [ ] figure out if there's a better way of doing this

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -556,6 +556,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('progress', () => {
+      if (this.mainSegmentLoader_.earlyabortTimer) {
+        window.clearTimeout(this.mainSegmentLoader_.earlyabortTimer);
+        this.mainSegmentLoader_.earlyabortTimer = null;
+      }
       if (this.experimentalBufferBasedABR) {
         const nextPlaylist = this.selectPlaylist();
 
@@ -615,6 +619,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
         // if we shouldn't switch to the next playlist, do nothing
         if (!this.shouldSwitchToMedia_(nextPlaylist)) {
           this.logger_(`earlyabort triggered, but we will not be switching from ${currentPlaylist.id} -> ${nextPlaylist.id}.`);
+          this.mainSegmentLoader_.earlyabortTimer = window.setTimeout(() => {
+            this.mainSegmentLoader_.trigger('earlyabort');
+          }, 10 * 1000);
           return;
         }
       }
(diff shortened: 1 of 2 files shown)


src/master-playlist-controller.js
```diff
@@ -556,6 +556,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('progress', () => {
+      if (this.mainSegmentLoader_.earlyabortTimer) {
+        window.clearTimeout(this.mainSegmentLoader_.earlyabortTimer);
```
brandonocasey, line 560: We have to clear this on dispose as well

src/master-playlist-controller.js
```diff
@@ -615,6 +619,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
         // if we shouldn't switch to the next playlist, do nothing
         if (!this.shouldSwitchToMedia_(nextPlaylist)) {
           this.logger_(`earlyabort triggered, but we will not be switching from ${currentPlaylist.id} -> ${nextPlaylist.id}.`);
+          this.mainSegmentLoader_.earlyabortTimer = window.setTimeout(() => {
+            this.mainSegmentLoader_.trigger('earlyabort');
+          }, 10 * 1000);
```
brandonocasey, line 624: 10s might be too long, what do you think about 5s?

src/master-playlist-controller.js
```diff
@@ -615,6 +619,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
         // if we shouldn't switch to the next playlist, do nothing
         if (!this.shouldSwitchToMedia_(nextPlaylist)) {
           this.logger_(`earlyabort triggered, but we will not be switching from ${currentPlaylist.id} -> ${nextPlaylist.id}.`);
+          this.mainSegmentLoader_.earlyabortTimer = window.setTimeout(() => {
```
brandonocasey, line 622: We should clear this timer every time we come into this function right?

src/master-playlist-controller.js
gkatsev: Probably should be target duration dependent or something, maybe even forward buffer? Made it 10s just for starters.

src/master-playlist-controller.js
gkatsev: Yeah, I have that as part of the TODOs. I wasn't sure whether we had a `player.setTimeout` method available here or whether I needed to make sure that the timer is handled fully as appropriate.
The variable should be renamed to include a `_` as well.

gkatsev:
I keep finding various edge cases here. I think we need to implement the proper timer based solution for rendition switches.

PR #967, opened 2020-10-01
fix: inline json version

brandonocasey:
## Description
Ideally we would be able to specify regex externals in videojs-generate-rollup-config, but we cannot currently. For now we will use [@rollup/plugin-replace](https://github.com/rollup/plugins/tree/master/packages/replace) to fix issues with downstream webpack builds caused by importing json directly.


gkatsev:
Aftering updating mux.js and vhs-utils tags to latest, I think this is working as expected now locally. Was a bit tricky to get it working, but I think it was my local setup.
I'll merge and release and then retest via https://github.com/videojs/video.js/pull/6867.

gkatsev (approved):

PR #968, opened 2020-10-02
Fix/check tech error

brandonocasey:
A poster error is causing playback to stall as it triggers an error on tech. Errors on tech and the video element were thought to be fatal, guess that assumption was wrong. :doh:


gkatsev (approved):

gesinger (approved):

gesinger:
Might be worth a comment, but code looks good

gkatsev:
Fixed via 0c7b2cb

PR #969, opened 2020-10-02
fix: check tech error before pause

brandonocasey:
See #968


gkatsev (approved):

gesinger (approved):
Would be worth a comment, but code looks good

gkatsev (approved):

PR #971, opened 2020-10-05
fix: appendsdone abort and handle multiple id3 sections.

brandonocasey:
`appendsdone` may cause an abort in segmentloaders which can then cause `pendingSegment` to be null **after** we just null checked it. Instead we should check if we have a `pendingSegment` then trigger appends done and see if an abort happens.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -2529,6 +2529,11 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @private
    */
   handleAppendsDone_() {
+    // appendsdone can cause an abort
+    if (this.pendingSegment_) {
+      this.trigger('appendsdone');
+    }
+
     if (!this.pendingSegment_) {
       this.state = 'READY';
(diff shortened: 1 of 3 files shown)


src/segment-loader.js
```diff
@@ -2529,6 +2529,11 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @private
    */
   handleAppendsDone_() {
+    // appendsdone can cause an abort
+    if (this.pendingSegment_) {
+      this.trigger('appendsdone');
```
brandonocasey, line 2534: this can cause `this.pendingSegment_` to become `null` via an abort.

gkatsev: So `pendingSegment_` be removed as part of `appendsdone`? Which will allow us to enter into the if statement below?

brandonocasey: Yes

gkatsev (approved):

Issue #975, opened 2020-10-21
Player enters tight infinite loop when live rendition manifests stop updating


skidder:
## Description
The VideoJS player enters a tight infinite loop requesting a single rendition manifest when the rendition manifests for a live stream stops being updated. The request-rate is only limited by the network connection speed. Rates exceeding 50 requests/sec have been observed. This has the potential to overwhelm the player device and video origin.

This can occur when the input for an HLS live stream is disconnected and the encoder keeps the live stream active during a reconnect window. During the reconnect window, the video origin might continue to serve rendition manifests that don't contain any new segments. Once the input is reconnected, the video origin can resume updating the rendition manifests and playback continues. If the reconnect window expires without the input being reconnected, then the rendition manifests can be marked as ended.

While the reconnect window is open, the player iterates over the set of rendition manifests looking for a manifest containing new segments. If all of the rendition manifests have been retrieved and no updates were found, then the player becomes stuck trying the same rendition manifest endlessly.

## Sources
HLS live streams that support input-stream reconnect-windows.

## Steps to reproduce
Demonstration: [https://youtu.be/bGCVSud2ZGk](https://youtu.be/bGCVSud2ZGk)

1. Start sending input stream to encoder configured with a reconnect window (e.g. 5 minutes)
2. Start playback in VideoJS
3. Disconnect input stream
4. Observe the player iterate over rendition manifests
5. Once all rendition manifests have been exhausted, the player enters a tight loop requesting the same rendition manifest endlessly
6. Reconnect the input stream
7. Observe the player resume playback

### Expected
The player should continually cycle through the set of rendition manifests with some delay. This is typically enforced through the `blacklistDuration`.

### Error output
`VIDEOJS: WARN: Problem encountered with the current HLS playlist. Playlist no longer updating. Switching to another playlist.`

### videojs-http-streaming version
videojs-http-streaming 2.2.2
possibly more

### videojs version
VideoJS 7.6.6
VideoJS 7.8.4
possibly more

### Browsers
* Chrome 86.0.4240.75 (64-bit) on Windows 10
* Edge 86.0.622.43 (64-bit) on Windows 10

### Platforms
* Consistently reproducible on Windows 10
* Occasionally reproducible on Mac OS X 10.15.6

### Other Plugins
No

### Other JavaScript
No

gkatsev:
From @gesinger in slack:

[...]

PR #976, opened 2020-10-21
fix: correct retry-behavior for rendition-manifest retrieval on HLS Live streams that stop updating.

skidder:
## Description
Fixes https://github.com/videojs/http-streaming/issues/975

This addresses a bug affecting live streams that support input-stream reconnect windows.

The player will iterate over all rendition manifests looking for new segments to download and play. When all rendition manifests have been evaluated and no updates are found, the player enters a tight infinite loop requesting a single rendition manifest until either a terminated rendition manifest is returned or the player is stopped/destroyed.

## Specific Changes proposed
This change updates the `this.media_` variable even when the rendition manifest contains no changes. This is necessary for the rendition manifest blacklisting to work in this scenario.

Demonstration of fix: https://youtu.be/hBsYg77_Ctw

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -261,10 +261,10 @@ export default class PlaylistLoader extends EventTarget {
     const update = updateMaster(this.master, playlist);
 
     this.targetDuration = playlist.targetDuration;
+    this.media_ = this.master.playlists[id];
 
     if (update) {
       this.master = update;
-      this.media_ = this.master.playlists[id];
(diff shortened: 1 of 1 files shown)


gkatsev:
As mentioned in https://github.com/videojs/http-streaming/issues/975, this isn't quite the right fix, while it will fix the issue, it can cause some other issues to pop up again.

PR #978, opened 2020-10-23
fix(experimentalBufferBasedABR): call selectPlaylist and change media on an interval

### LINKED ISSUE #964: experimentalBufferBasedABR (and maybe regular ABR too): fixup ABR when no progress events from XHR

gkatsev:
This is a followup to #886. There, I noticed that if we get sustained bad network we will time out because we will an opportunity for earlyabort. At first, I tried working around by delaying earlyabort in that case (#966) but as I was getting that ready I kept finding edge cases that we needed to account for. Specifically, the issue is that we only run our ABR algorithm on `progress` and `bandwidthchange` events from the segment loader. This means that if the download stalls, we stop run our algorithm and will eventually stall playback. Instead, we should remove earlyabort altogether (f897dde) and set up an interval (currently 250ms) to re-run our ABR algorithm. This means that if the network becomes bad for a sustained period, we will drop down once the buffer allows us but if the network recovers, we will switch up appropriately as well.

Also, the interval is stopped while we're paused.

Fixes #964.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -209,6 +209,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     const segmentLoaderSettings = {
       vhs: this.vhs_,
+      experimentalBufferBasedABR,
       mediaSource: this.mediaSource,
       currentTime: this.tech_.currentTime.bind(this.tech_),
       seekable: () => this.seekable(),
@@ -599,25 +600,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('earlyabort', (event) => {
-      if (this.experimentalBufferBasedABR) {
-        const currentPlaylist = this.masterPlaylistLoader_.media();
-
-        // temporarily exclude the current playlist so that we can
-        // determine the next playlist that would be selected
-        // if this playlist were to be excluded.
-        currentPlaylist.excludeUntil = Infinity;
-
-        const nextPlaylist = this.selectPlaylist();
-
-        // un-exclude the current playlist for now
-        currentPlaylist.excludeUntil = null;
-
-        // if we shouldn't switch to the next playlist, do nothing
-        if (!this.shouldSwitchToMedia_(nextPlaylist)) {
-          this.logger_(`earlyabort triggered, but we will not be switching from ${currentPlaylist.id} -> ${nextPlaylist.id}.`);
-          return;
-        }
-      }
       this.blacklistCurrentPlaylist({
         message: 'Aborted early because there isn\'t enough bandwidth to complete the ' +
           'request without rebuffering.'
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -209,7 +209,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     const segmentLoaderSettings = {
       vhs: this.vhs_,
-      experimentalBufferBasedABR,
       mediaSource: this.mediaSource,
       currentTime: this.tech_.currentTime.bind(this.tech_),
       seekable: () => this.seekable(),
@@ -634,6 +633,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('earlyabort', (event) => {
+      // never try to early abort with the new ABR algorithm
+      if (this.experimentalBufferBasedABR) {
+        return;
+      }
+
       this.blacklistCurrentPlaylist({
         message: 'Aborted early because there isn\'t enough bandwidth to complete the ' +
           'request without rebuffering.'
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -923,7 +923,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return;
     }
 
-    this.abrTimer_();
+    this.stopABRTimer_();
     this.sourceUpdater_.endOfStream();
   }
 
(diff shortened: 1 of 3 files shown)


src/master-playlist-controller.js
```diff
@@ -599,25 +642,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('earlyabort', (event) => {
-      if (this.experimentalBufferBasedABR) {
```
brandonocasey, line 602: Should we return here for `this.experimentalBufferBasedABR` so that we don't use earlyabort?

src/master-playlist-controller.js
```diff
@@ -258,6 +259,12 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.setupSegmentLoaderListeners_();
 
+    if (this.experimentalBufferBasedABR) {
+      this.startABRTimer_();
+      this.tech_.on('pause', () => this.stopABRTimer_());
```
brandonocasey, line 264: I wonder if we should handle seeks that cause a pause differently here. Should they be ignored so that we can switch playlists during a seek? Maybe that already happens?

src/master-playlist-controller.js
```diff
@@ -545,27 +586,29 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
         }
-      }
 
-      this.trigger('progress');
-    });
+        this.tech_.trigger('bandwidthupdate');
+      });
+
+      this.mainSegmentLoader_.on('progress', () => {
+        if (this.experimentalBufferBasedABR) {
```
brandonocasey, line 601: I think we can remove this if statement and everything inside of it now.

src/master-playlist-controller.js
```diff
@@ -1066,6 +1091,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
    */
   pauseLoading() {
     this.delegateLoaders_('all', ['abort', 'pause']);
+    this.stopABRTimer_();
```
brandonocasey, line 1094: Should we start the abr timer in `load()`?

src/master-playlist-controller.js
```diff
@@ -258,6 +259,12 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.setupSegmentLoaderListeners_();
 
+    if (this.experimentalBufferBasedABR) {
+      this.startABRTimer_();
```
brandonocasey, line 263: Should we be starting the timer before the first `play`?

brandonocasey (commented):
A few comments for now, still thinking about this. I think we might need to start the abr timer in a few other places just to be safe. Going to do some testing.

src/master-playlist-controller.js
gkatsev: I turned off triggering earlyabort in the segment loader, so, theoretically, we should never get here if `experimentalBufferBasedABR` is set.

src/master-playlist-controller.js
brandonocasey: Was thinking about this a bit more and perhaps we just want to call `this.checkABR_` on `seeking` so that we can change playlist during a seek rather than afterwards?

brandonocasey:
After some testing I think that we start/stop the timers in a fine places. Everything seems to work very smoothly now too. This solution seems way better than earlyabort, and it will keep all of the logic in masterPlaylistController where it belongs 👍

src/master-playlist-controller.js
gkatsev: yeah, this is unnecessary now.

src/master-playlist-controller.js
gkatsev: we probably need to update this based on the preload setting. If preload is none or metedata (i.e. when we shouldn't have downloaded any segments) there's no point to it but with preload auto, I think we should.

I'll add updating this as part of the story to fix preload=none

src/master-playlist-controller.js
gkatsev: If we're just seeking by clicking, the pause is so small as to not matter. If we're scrubbing, we'll get a whole bunch of seeking events, so, we'd want to debounce it or something.
Maybe `stopABRTimer_` should check whether scrubbing is set? So, we only stop it if we're paused for not scrubbing.

gkatsev: Updated the stop handler to check `tech.scrubbing()` if it exists, which is being added here https://github.com/videojs/video.js/pull/6920

test/master-playlist-controller.test.js
```diff
@@ -5534,53 +5534,6 @@ QUnit.module('MasterPlaylistController experimentalBufferBasedABR', {
   }
 });
 
-QUnit.test('Determines if playlist should be aborted on earlyabort', function(assert) {
```
gesinger, line 5537: Should this test be left in just when `experimentalBufferBasedABR` is false?

gkatsev: Maybe, though, this test was added as part of https://github.com/videojs/http-streaming/pull/886.

gkatsev: Actually, looks like we do have a similar test that already https://github.com/videojs/http-streaming/blob/ca7655e0f8ef12f708367e0bedbde3d38e900659/test/master-playlist-controller.test.js#L1847-L1887. This test is a variation of that that accounts for `shouldSwitchToMedia` variations with `experimentalBufferBasedABR`

PR #980, opened 2020-10-26
fix(playback-watcher): ignore subtitles

gkatsev:
It's possible that a subtitle track will have segments without any text
in them depending on how the subtitles and captions were created.
Therefor, we should ignore it in the playback-watcher where we're
checking for excessive downloads.

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -218,6 +218,10 @@ export default class PlaybackWatcher {
     this.resetSegmentDownloads_(type);
     this.tech_.trigger({type: 'usage', name: `vhs-${type}-download-exclusion`});
 
+    if (type === 'subtitle') {
+      return;
+    }
+
     // TODO: should we exclude audio tracks rather than main tracks
     // when type is audio?
     mpc.blacklistCurrentPlaylist({
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -11,7 +11,6 @@
 import window from 'global/window';
 import * as Ranges from './ranges';
 import logger from './util/logger';
-import videojs from 'video.js';
 
 // Set of events that reset the playback-watcher time check logic and clear the timeout
 const timerCancelEvents = [
@@ -218,18 +217,6 @@ export default class PlaybackWatcher {
     this.logger_(`${type} loader stalled download exclusion`);
     this.resetSegmentDownloads_(type);
     this.tech_.trigger({type: 'usage', name: `vhs-${type}-download-exclusion`});
-
-    if (type === 'subtitle') {
-      // TODO: Is there anything else that we can do here?
-      // removing the track and disabling could have accesiblity implications.
-      const track = loader.track();
-      const label = track.label || track.language || 'Unknown';
-
-      videojs.log.warn(`Text track "${label}" is not working correctly. It will be disabled and excluded.`);
-      track.mode = 'disabled';
-      this.tech_.textTracks().removeTrack(track);
-      return;
-    }
 
     // TODO: should we exclude audio tracks rather than main tracks
     // when type is audio?
--- a/test/playback-watcher.test.js
+++ b/test/playback-watcher.test.js
@@ -933,64 +933,64 @@ QUnit.module('PlaybackWatcher download detection', {
 });
 
 loaderTypes.forEach(function(type) {
-  QUnit.test(`detects ${type} appends without buffer changes and excludes`, function(assert) {
-    this.setup();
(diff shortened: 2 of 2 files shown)


src/playback-watcher.js
```diff
@@ -219,18 +218,6 @@ export default class PlaybackWatcher {
     this.resetSegmentDownloads_(type);
     this.tech_.trigger({type: 'usage', name: `vhs-${type}-download-exclusion`});
 
-    if (type === 'subtitle') {
```
gesinger, line 222: I think by removing this we will still try to blacklist subtitle tracks when we don't have subtitles in the segments. Should we just return early for subtitles tracks, or, in loaderTypes above, remove subtitles?

gkatsev: oh, we'll still try to exclude it but at least it won't be removed. Yeah, exiting early here when it's of `type` `subtitle` sounds good.

gkatsev: Updated to return here but not remove the track.

test/playback-watcher.test.js
```diff
@@ -933,64 +933,64 @@ QUnit.module('PlaybackWatcher download detection', {
 });
 
 loaderTypes.forEach(function(type) {
-  QUnit.test(`detects ${type} appends without buffer changes and excludes`, function(assert) {
-    this.setup();
-    const loader = this.mpc[`${type}SegmentLoader_`];
-    const track = {label: 'foobar', mode: 'showing'};
+  if (type !== 'subtitle') {
```
brandonocasey, line 936: perhaps we should just return in here if `type === 'subtitle'` would keep the changes down. LGTM otherwise.

gkatsev: I tried that but it didn't seem to work.

gesinger: That seems strange. It wouldn't execute for other loader types of returning early on subtitles?

test/playback-watcher.test.js
```diff
@@ -933,64 +933,64 @@ QUnit.module('PlaybackWatcher download detection', {
...
+    QUnit.test(`detects ${type} appends without buffer changes and excludes`, function(assert) {
+      this.setup();
+      const loader = this.mpc[`${type}SegmentLoader_`];
+      const track = {label: 'foobar', mode: 'showing'};
 
-    if (type === 'subtitle') {
-      loader.track = () => track;
-      sinon.stub(this.player.tech_.textTracks(), 'removeTrack');
-    }
+      if (type === 'subtitle') {
```
gesinger, line 942: Should this past be removed because we return early on subtitles?

test/playback-watcher.test.js
```diff
@@ -933,64 +933,64 @@ QUnit.module('PlaybackWatcher download detection', {
...
+      const expectedUsage = {};
 
-    expectedUsage[`vhs-${type}-download-exclusion`] = 1;
+      expectedUsage[`vhs-${type}-download-exclusion`] = 1;
 
-    if (type !== 'subtitle') {
-      expectedUsage['vhs-rendition-blacklisted'] = 1;
-      expectedUsage['hls-rendition-blacklisted'] = 1;
-    }
+      if (type !== 'subtitle') {
```
gesinger, line 962: This conditional probably can be removed.

test/playback-watcher.test.js
```diff
@@ -933,64 +933,64 @@ QUnit.module('PlaybackWatcher download detection', {
...
+        expectedUsage['vhs-rendition-blacklisted'] = 1;
+        expectedUsage['hls-rendition-blacklisted'] = 1;
+      }
 
-    assert.deepEqual(this.usageEvents, expectedUsage, 'usage as expected');
+      assert.deepEqual(this.usageEvents, expectedUsage, 'usage as expected');
 
-    if (type !== 'subtitle') {
-      const message = 'Playback cannot continue. No available working or supported playlists.';
+      if (type !== 'subtitle') {
```
gesinger, line 969: This conditional probably can be removed, and the else as well.

src/playback-watcher.js
```diff
@@ -220,14 +219,6 @@ export default class PlaybackWatcher {
     this.tech_.trigger({type: 'usage', name: `vhs-${type}-download-exclusion`});
 
     if (type === 'subtitle') {
```
gesinger, line 221: Might be worth having a comment here. Something like `// subtitles aren't guaranteed to be present for every bit of video, therefore, we can't determine whether downloads have stalled based on the subtitle's buffer`

gesinger (commented):
Sorry for the late review, just catching up.

Issue #977, opened 2020-10-23
Playback fails for stream with different video and audio offsets


evanfarina:
## Description
The [stream in question](https://livectorprodmedia11-usw22.licdn.com/cf8af838-31ed-48f0-89ba-ae9f7f4a4667/L565cc2c652e3029000-livemanifest.ism/manifest(format=m3u8-aapl)) has a video PTS of 25.385 and an audio PTS of 3.392. In Chrome ([test URL](https://videojs-http-streaming.netlify.app/?debug=false&autoplay=false&muted=false&minified=false&liveui=false&partial=false&url=https%3A%2F%2Flivectorprodmedia11-usw22.licdn.com%2Fcf8af838-31ed-48f0-89ba-ae9f7f4a4667%2FL565cc2c652e3029000-livemanifest.ism%2Fmanifest(format%3Dm3u8-aapl)&type=application%2Fx-mpegurl&keysystems=&buffer-water=false)), the video will spend three seconds trying to play before it ultimately stalls out.

## Sources
https://livectorprodmedia11-usw22.licdn.com/cf8af838-31ed-48f0-89ba-ae9f7f4a4667/L565cc2c652e3029000-livemanifest.ism/manifest(format=m3u8-aapl)

## Steps to reproduce

1. Visit the [http-streaming demo page](https://videojs-http-streaming.netlify.app/?debug=false&autoplay=false&muted=false&minified=false&liveui=false&partial=false&url=https%3A%2F%2Flivectorprodmedia11-usw22.licdn.com%2Fcf8af838-31ed-48f0-89ba-ae9f7f4a4667%2FL565cc2c652e3029000-livemanifest.ism%2Fmanifest(format%3Dm3u8-aapl)&type=application%2Fx-mpegurl&keysystems=&buffer-water=false)
2. Enter the above source into the box for the media's source
3. Click the "load" button and then attempt to play the stream
4. Allow the stream to play for ~3-4 seconds. Notice that the stream will stall and the loading spinner will appear. The stream will not recover from this point on

### Expected
I expect the stream to play without stalling out.

### Error output
No error output

## Additional Information
[...]

PR #981, opened 2020-10-28
fix: only prevent audio group creation if no other playlists are using it

brandonocasey:
Fixes an issue (not the only issue) with #977

caused by the fix in https://github.com/videojs/http-streaming/pull/952

### DIFF
--- a/src/media-groups.js
+++ b/src/media-groups.js
@@ -412,7 +412,7 @@ export const initialize = {
         });
 
         // If there are no playlists using this audio group other than ones
-        // that match it's uri. Then playlist is audio only. We delete the resolvedUri
+        // that match it's uri, then the playlist is audio only. We delete the resolvedUri
         // property here to prevent a playlist loader from being created so that we don't have
         // both the main and audio segment loaders loading the same audio segments
         // from the same playlist.
--- a/src/media-groups.js
+++ b/src/media-groups.js
@@ -405,18 +405,18 @@ export const initialize = {
       for (const variantLabel in mediaGroups[type][groupId]) {
         let properties = mediaGroups[type][groupId][variantLabel];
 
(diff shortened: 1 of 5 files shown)


gkatsev (approved):

src/media-groups.js
```diff
@@ -405,18 +405,18 @@ export const initialize = {
...
         });
 
-        if (matchingPlaylists.length) {
-          // If there is a playlist that has the same uri as this audio variant, assume
-          // that the playlist is audio only. We delete the resolvedUri property here
-          // to prevent a playlist loader from being created so that we don't have
-          // both the main and audio segment loaders loading the same audio segments
-          // from the same playlist.
+        // If there are no playlists using this audio group other than ones
+        // that match it's uri. Then playlist is audio only. We delete the resolvedUri
```
gesinger, line 415:
```suggestion
        // that match it's uri, then the playlist is audio only. We delete the resolvedUri
```

gesinger, line 415: One thing I'm wondering is: what if we have a manifest where there are two playlists using the audio group, one is audio only, and another uses it as its demuxed audio for video. Wouldn't we need the audio loader running for that case?

gesinger (commented):
Would probably be good to have some tests for this one, particularly since it might help document it.

src/media-groups.js
brandonocasey: We would but we don't support going from video/audio to just audio or vice versa, if I understand what you are saying.

evanfarina:
Thanks for such a fast turnaround @brandonocasey , @gesinger and @gkatsev. I'll give this a shot first thing Monday.

gkatsev:
This only fixes a portion of it. There's still the PTS weirdness.
Interestingly, with this fix, your stream doesn't start but it does work if I seek.

PR #984, opened 2020-11-01
[Fix #983] - Introduce an initialization option to limit reloads of blacklisted playlists

### LINKED ISSUE #983: Player is not respecting excludeUntil time, causing player to rapidly fetch playlists that it previously failed to fetch
## Description
In short, the player continues to re-fetch playlists if it encounters an error while fetching. The `excludeUntil` value is not respected and the user has no ability to pass in an option to throttle this behavior. This ultimately results in a 401 response being sent every second until the player is destroyed (user closes tab).

## Sources
The repro source that I'm sharing here is an encrypted stream which our backend has marked as expired. Because the stream has expired, our backend returns a 401 for the encryption key fetch. The 401 response causes the player to then try fetching a new playlist, and so on and so forth.

https://streamweu-livectorprodmedia17-euwe.licdn.com/2888a657-100e-4e8d-bb2a-e7c614e5099e/L4E5d5d28e840816000-livemanifest.ism/manifest(format=m3u8-aapl,encryption=cbc)

## Steps to reproduce

1. Create a player using the above source. No plugins or extra configuration is necessary.
2. Play the stream
3. Open Chrome's inspector and go to the network tab. Notice that the requests for the encryption key result in a 401. Also notice that the playlist being requested rotates.

### Expected
1. In general, I'm okay with the player cycling through playlists when it encounters a failure. However, we should not remove a playlist's blacklist after reaching the final rendition. If we keep this behavior, it at the very least should be behind an option (e.g; maxRetries=3 will let the player try fetching each playlist 3 times before giving up).
2. We should be respecting the excludeUntil time. It defaults to 30s, but because of the issue described in #1, it's not respected.
3. Perhaps we can consider not treating all errors in the same way. In the case of a 401, is it really worth trying a new playlist?

Overall, my expectations are that the player does not continue to fetch playlists that previously failed. And, if it does, it should do so behind an option so that users can limit how long the retry goes on for.

I tested with the 1.x release and the 2.x release. Same issue in both releases.

### videojs version
video.js 7.10.1

### Browsers
* Chrome

### Platforms
* Mac

### Other Plugins
* No other plugins

evanfarina:
This addresses #983 by introducing a new initialization option which allows clients to limit how many times the player will retry to load a previously-blacklisted playlist.

## Description
Fixes #983 .

## Specific Changes proposed

1. Introduce a new initialization option called `maxBlacklistedPlaylistReloads`
2. Add a new condition before removing a previously-blacklisted playlist from the blacklist
3. Update docs for `blacklistDuration`
4. Added a new test that demonstrates that a playlist is not re-tried after hitting the max # of retries set by `maxBlacklistedPlaylistReloads`

## Requirements Checklist
  - [ ] Docs/guides updated


src/master-playlist-controller.js
```diff
@@ -161,6 +162,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.useCueTags_ = useCueTags;
     this.blacklistDuration = blacklistDuration;
     this.enableLowInitialPlaylist = enableLowInitialPlaylist;
+    this.maxBlacklistedPlaylistReloads = maxBlacklistedPlaylistReloads;
+
+    this.blacklistedPlaylistReloads = 0;
```
gkatsev, line 167: I think this should be suffixed with a `_` and renamed similarly to how we end up renaming `maxBlacklistedPlaylistReloads`

src/master-playlist-controller.js
```diff
@@ -1029,14 +1034,16 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-
         // a playlist cannot be reincluded if it wasn't excluded to begin with.
+
         if (typeof excludeUntil !== 'undefined' && excludeUntil !== Infinity) {
           reincluded = true;
           delete playlist.excludeUntil;
         }
       });
 
+      this.blacklistedPlaylistReloads++;
```
gkatsev, line 1045: Looks like this value is only ever incremented here, I'd imagine that on a successful playback it should be reset back to 0? I'm not exactly sure where it should go, though.

gkatsev (commented):
Also, a rebase would be good so that it picks up the new github ci settings.

src/master-playlist-controller.js
evanfarina: Renamed to `finalRenditionRetries_`

src/master-playlist-controller.js
evanfarina: Updated so that we now reset the value on `appendsdone`. In case of an error, we `return` before appendsdone is fired. @gesinger can you confirm that `appendsdone` is an appropriate place to reset the number of final rendition retries?

wseymour15:
Can we close this review out? It looks like the issue has been resolved.

gkatsev:
Yup, since we have https://github.com/videojs/http-streaming#maxplaylistretries, I think this is more or less covered by that.

PR #989, opened 2020-11-04
chore: switch from travis to github actions for ci

brandonocasey:
This switches our code to be tested on github actions as travis ci has become unreasonably slow.

Depends upon https://github.com/videojs/videojs-generate-karma-config/pull/37

The tests for this pull request are run in a forked branch to show that they can work for a fork. To see local tests using browserstack see https://github.com/videojs/http-streaming/runs/1360756885?check_suite_focus=true


test/playback.test.js
```diff
@@ -38,7 +38,7 @@ if (videojs.browser.IE_VERSION || videojs.browser.IS_EDGE) {
 
 QUnit.module('Playback', {
   beforeEach(assert) {
-    assert.timeout(50000);
+    assert.timeout(60000);
```
brandonocasey, line 41: Timeouts for playback tests need to be a bit longer when we run on github actions for some reason.

test/playback.test.js
```diff
@@ -19,7 +19,7 @@ const playFor = function(player, time, cb) {
 
   const checkPlayerTime = function() {
     window.setTimeout(() => {
-      if (player.currentTime() <= targetTime) {
+      if (player.tech_ && player.tech_.el_ && player.currentTime() <= targetTime) {
```
brandonocasey, line 22: This prevents an error from being logged when this functoin runs after the player has disposed on a test timeout.

.github/workflows/npm-test.yml
```diff
@@ -0,0 +1,33 @@
+name: unit tests
+
+on: [push, pull_request]
+
+jobs:
+  run-tests:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest]
```
brandonocasey, line 10: I think that in the future it will be possible for us to test on windows and drop browserstack entirely. Not that I hate browserstack but it would allow us to unify testing in pull requests and pushes. Just have to drop ie 11 support...

.nvmrc
```diff
@@ -1 +1 @@
-lts/dubnium
```
brandonocasey, line 1: I think we should switch to using a number rather than a name. This will use the latest version of node 10, same as the name, but it is much easier to understand.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Run Mac test
+      run: npm run test
+      if: ${{ startsWith(matrix.os, 'macos') }}
+
+    # only run ie 11 on windows
+    - name: Run Windows test
+      run: npm run test
+      if: ${{ startsWith(matrix.os, 'windows') }}
+
+    # run chrome/firefox in linux
```
gkatsev, line 54: is it possible to run these in parallel?
Though, we can do that as a followup too.

.nvmrc
gkatsev: I agree that the codenames aren't great, though, this file is mostly for usage with nvm directly. Having the `lts/` in there sends a signal to people looking that we are on an lts release rather than just a "random" version number.
We previously used `lts/*` but that's a problem because sometimes LTS updates can cause an issue.

On a separate note we're 2 LTS majors behind, though, definitely shouldn't be updated as part of this release.

.github/workflows/ci.yml
gkatsev: Though, currently we're only running on linux for BS?
I'm OK with making these updates as a follow-up.

brandonocasey: all os steps run in parallel yeah, we are only running bs on linux

.nvmrc
brandonocasey: Yes we are 2 LTS versions behind and likely all of our tooling will drop support for v10 in April when it is deprecated, but things will likely still work a bit past April. I would recommend that we move to v16 at that point (or wait for it) so that we won't have to update again for awhile. That is, unless we run into problems with v16. but you are right we should hold off on all that for another discussion/time.

.github/workflows/ci.yml
brandonocasey: We are actually only running on linux right now period for everything, as windows/mac are not included in the os matrix list at all.

gkatsev: cool, yeah, wasn't sure whether the matrix items run in parallel or not.

.nvmrc
gkatsev: node 14 just went into LTS, so, it's what we should update it. 16 won't get to LTS until october 2021 https://nodejs.org/en/about/releases/

brandonocasey: ahh good point, didn't look up the timeline, but yeah 14 is fine too then.

gkatsev: fyi, I still would prefer to keep `lts/dubnium` :)

.nvmrc
```diff
@@ -1 +1 @@
...
+10
```
gkatsev, line 1: This was switched to a number because `actions/setup-node` requires a number and can't use `lts/dubnium`. While I still like the signal that we use `lts` here, having a single source of truth for the node version we're on is nice.

gkatsev (approved):

PR #992, opened 2020-11-10
refactor: Add a better distinction between master and child dash loaders

brandonocasey:
## Description
I am working on refactoring the parsing logic in dash-playlist-loader. This is difficult because the logic is almost impossible to follow in the current state of the code. I tried a grand over-arching refactor and ultimately wasn't happy with it. Instead I am going to opt for smaller changes to get us towards what we want.

## Steps
1. (this pull request) Add a property (isMaster_) to identify if the dash-playlist-loader is master. When masterPlaylistLoader is not passed into the constructor we know that the master playlist loader will be this one. We also shadow `this.masterPlaylistLoader` to `this` for master so that operations that update master always happen on `this.masterPlaylistLoader`. This simplifies a lot of the code and makes it much easier to understand. 
2. refactor parseMasterXml so that it keeps track of xml that has already been parsed, and does not parse again. Also make `filterChangedSidxMappings` use this logic rather than parsing on it's own.
3. (might be part of 2) refactor any requesting/parsing logic that we can into a central place so that we are not calling the same functions everywhere.

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -252,6 +252,11 @@ export default class DashPlaylistLoader extends EventTarget {
   constructor(srcUrlOrPlaylist, vhs, options = { }, masterPlaylistLoader) {
     super();
 
+    if (!masterPlaylistLoader) {
+      this.masterPlaylistLoader_ = this;
+      this.isMaster_ = true;
+    }
+
     const { withCredentials = false, handleManifestRedirects = false } = options;
 
     this.vhs_ = vhs;
@@ -277,20 +282,15 @@ export default class DashPlaylistLoader extends EventTarget {
 
     // initialize the loader state
     // The masterPlaylistLoader will be created with a string
-    if (typeof srcUrlOrPlaylist === 'string') {
+    if (this.isMaster_) {
       this.srcUrl = srcUrlOrPlaylist;
       // TODO: reset sidxMapping between period changes
       // once multi-period is refactored
       this.sidxMapping_ = {};
-      return;
+    } else {
+      this.masterPlaylistLoader_ = masterPlaylistLoader;
+      this.childPlaylist_ = srcUrlOrPlaylist;
     }
-
-    this.setupChildLoader(masterPlaylistLoader, srcUrlOrPlaylist);
-  }
-
-  setupChildLoader(masterPlaylistLoader, playlist) {
-    this.masterPlaylistLoader_ = masterPlaylistLoader;
-    this.childPlaylist_ = playlist;
   }
 
   dispose() {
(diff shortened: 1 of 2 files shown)


src/dash-playlist-loader.js
```diff
@@ -277,20 +282,15 @@ export default class DashPlaylistLoader extends EventTarget {
 
     // initialize the loader state
     // The masterPlaylistLoader will be created with a string
-    if (typeof srcUrlOrPlaylist === 'string') {
+    if (this.isMaster_) {
```
gkatsev, line 285: should we verify that `srcUrlOrPlaylist` is a string or is it basically guaranteed to be one if we're a master manifest?

gkatsev (commented):
Just one question, really

src/dash-playlist-loader.js
brandonocasey: For master it will always be a string, for others it can be a playlist or string. So yes it would seem so.

src/dash-playlist-loader.js
```diff
@@ -528,7 +518,7 @@ export default class DashPlaylistLoader extends EventTarget {
 
     // request the specified URL
     this.request = this.vhs_.xhr({
-      uri: this.srcUrl,
+      uri: this.masterPlaylistLoader_.srcUrl,
```
gesinger, line 521: Since we return early above for non master loaders, would it be clearer to use just `this`?

src/dash-playlist-loader.js
```diff
@@ -277,20 +282,15 @@ export default class DashPlaylistLoader extends EventTarget {
...
     // initialize the loader state
     // The masterPlaylistLoader will be created with a string
-    if (typeof srcUrlOrPlaylist === 'string') {
+    if (this.isMaster_) {
       this.srcUrl = srcUrlOrPlaylist;
       // TODO: reset sidxMapping between period changes
       // once multi-period is refactored
-      this.sidxMapping_ = {};
-      return;
+      this.masterPlaylistLoader_.sidxMapping_ = {};
```
gesinger, line 289: Since we have the conditional above for master loaders, would it be clearer to use just `this`?

test/dash-playlist-loader.test.js
```diff
@@ -667,6 +668,8 @@ QUnit.test('constructor sets srcUrl and other properties', function(assert) {
   assert.strictEqual(childLoader.state, 'HAVE_NOTHING', 'correct state');
   assert.deepEqual(childLoader.loadedPlaylists_, {}, 'correct loadedPlaylist state');
   assert.ok(childLoader.masterPlaylistLoader_, 'should be a masterPlaylistLoader');
+  assert.notEqual(childLoader.masterPlaylistLoader_, childLoader, 'should be a masterPlaylistLoader');
```
gesinger, line 671:
```suggestion
  assert.notEqual(childLoader.masterPlaylistLoader_, childLoader, 'should not be a masterPlaylistLoader');
```

src/dash-playlist-loader.js
```diff
@@ -631,12 +621,12 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    if (this.isMaster_) {
       this.updateMainManifest_(parseMasterXml({
-        masterXml: this.masterXml_,
-        srcUrl: this.srcUrl,
-        clientOffset: this.clientOffset_,
-        sidxMapping: this.sidxMapping_
+        masterXml: this.masterPlaylistLoader_.masterXml_,
+        srcUrl: this.masterPlaylistLoader_.srcUrl,
+        clientOffset: this.masterPlaylistLoader_.clientOffset_,
+        sidxMapping: this.masterPlaylistLoader_.sidxMapping_
```
gesinger, line 629: Same question re: above

src/dash-playlist-loader.js
brandonocasey: I think we want it to be like this because:
1. The distinction will help us migrate master playlist loader code into a new file after these refactors.
2. It's much easier to know that you are accessing the correct object if you always access it the same way.

gesinger: Makes sense. Thanks!

gesinger (approved):

PR #995, opened 2020-11-13
fix: Revert #574 to fix caption removal issue

alex-barstow:
## Description
Revert #574. That change is resulting in WebVTT cues with overlapping time-intervals (which the spec allows) being removed from the cue list.


gkatsev:
Unfortunately, testing with bipbop, most of the cues are no longer present.

PR #997, opened 2020-11-16
feat: Configurable live seekable window

dario-fiore:
## Description
During live streaming in low bandwidth situations `playback-watcher` try different times to seek to live edge without waiting for buffered contents. This behaviour causing waiting status "loop" because buffer will never reach minimum contents in order to perform playback.

## Specific Changes proposed
Configure method `beforeSeekableWindow_` in order to accept new configuration parameter `liveRangeSafeTimeDelta` to adjust safe time delta.

## Requirements Checklist
  - [X] Docs/guides updated
  - [X] Example created ([JSBin](https://jsbin.com/vuguqik/edit?html,output))


src/playback-watcher.js
```diff
@@ -502,7 +503,7 @@ export default class PlaybackWatcher {
     if (seekable.length &&
         // can't fall before 0 and 0 seekable start identifies VOD stream
         seekable.start(0) > 0 &&
-        currentTime < seekable.start(0) - Ranges.SAFE_TIME_DELTA) {
+        currentTime < seekable.start(0) - (this.liveRangeSafeTimeDelta || Ranges.SAFE_TIME_DELTA)) {
```
gkatsev, line 506: What about using `this.liveRangeSafeTimeDelta` here always and then in https://github.com/videojs/http-streaming/blob/c9b195b8d8451118960aba2977c1bbc95fda0120/src/videojs-http-streaming.js#L639-L643 defaulting it to `Ranges.SAFE_TIME_DELTA` if a value wasn't provided. Probably by adding an object before `this.options` there.
```js
videojs.mergeOptions({
  liveRangeSafeTimeDelta: Ranges.SAFE_TIME_DELTA
},
this.options_,
{
  seekable: () => this.seekable(),
  media: () => this.masterPlaylistController_.media(),
  masterPlaylistController: this.masterPlaylistController_
});
```

README.md
```diff
@@ -451,6 +452,11 @@ This option defaults to `false`.
 * Default: `false`
 * Use partial appends in the transmuxer and segment loader
 
+##### liveRangeSafeTimeDelta
+* Type: `boolean`,
+* Default: `false`
```
gkatsev, line 457: This is a number that defaults to the SAFE_TIME_DELTA

gkatsev (commented):
Thanks!

src/playback-watcher.js
dario-fiore: Done 👍🏻

README.md
dario-fiore: Done 👍🏻

dario-fiore:
@gkatsev It seems that tests fail for timeout. I tried locally on 2.3.0 and errors are the same. Any ideas?

gkatsev:
Not sure, I'll take a look.

src/playback-watcher.js
dario-fiore: @gkatsev Are the new changes ok?

gkatsev: yes, I think so. We'll re-review and take it for a quick spin but likely good to go.

gkatsev (approved):

PR #994, opened 2020-11-13
refactor: add sidx segments instead of reparsing

brandonocasey:
## Description
~~The second pull request in our dash-playlist-loader refactor, based on #992~~ now merged

~~This pull request depends upon https://github.com/videojs/mpd-parser/pull/109~~ now released, and using that version.
## Specific Changes proposed
1. Consider sidx in updateMaster so that playlists that had sidx added and thus haven't actually changed do not trigger a master update.
2. Only parse master xml once for a given xml string. Previously we parsed it 3 times for non-sidx and 4 times for sidx.
3. Unify some sidx and master playlist handling logic

PR #998, opened 2020-11-16
refactor: unify sidx/master/error request logic

brandonocasey:
## Description
Third and final dash-playlist-loader refactor.
Based on #994 

## Specific Changes proposed
* Merge all master playlist loader request logic into `requestMaster_`
* Merge all sidx logic into `addSidxSegments_`
* Merge all error logic into `requestErrored_`

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -252,9 +252,8 @@ export default class DashPlaylistLoader extends EventTarget {
     this.request = null;
 
     if (err) {
-      // use the provided error or create one
-      // see requestSidx_ for the container request
-      // that can cause this.
+      // use the provided error object or create one
+      // based on the request/response
       this.error = typeof err === 'object' && !(err instanceof Error) ? err : {
         status: request.status,
         message: 'DASH request error at URL: ' + request.uri,
@@ -627,7 +626,7 @@ export default class DashPlaylistLoader extends EventTarget {
       newMaster = updateMaster(oldMaster, newMaster, this.masterPlaylistLoader_.sidxMapping_);
     }
 
-    // only update master, if we have a new master.
+    // only update master if we have a new master
     this.masterPlaylistLoader_.master = newMaster ? newMaster : oldMaster;
     const location = this.masterPlaylistLoader_.master.locations && this.masterPlaylistLoader_.master.locations[0];
 
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -195,58 +195,6 @@ export const filterChangedSidxMappings = (master, oldSidxMapping) => {
   return mediaGroupSidx;
 };
 
-// exported for testing
-export const requestSidx_ = (loader, sidxRange, playlist, xhr, options, finishProcessingFn) => {
-  const sidxInfo = {
-    // resolve the segment URL relative to the playlist
-    uri: resolveManifestRedirect(options.handleManifestRedirects, sidxRange.resolvedUri),
-    // resolvedUri: sidxRange.resolvedUri,
-    byterange: sidxRange.byterange,
-    // the segment's playlist
-    playlist
-  };
-
-  const sidxRequestOptions = videojs.mergeOptions(sidxInfo, {
-    responseType: 'arraybuffer',
-    headers: segmentXhrHeaders(sidxInfo)
-  });
-
-  return containerRequest(sidxInfo.uri, xhr, (err, request, container, bytes) => {
-    if (err) {
-      return finishProcessingFn(err, request);
-    }
-
-    if (!container || container !== 'mp4') {
-      return finishProcessingFn({
-        status: request.status,
-        message: `Unsupported ${container || 'unknown'} container type for sidx segment at URL: ${sidxInfo.uri}`,
-        // response is just bytes in this case
-        // but we really don't want to return that.
-        response: '',
-        playlist,
-        internal: true,
-        blacklistDuration: Infinity,
-        // MEDIA_ERR_NETWORK
-        code: 2
-      }, request);
-    }
-
-    // if we already downloaded the sidx bytes in the container request, use them
-    const {offset, length} = sidxInfo.byterange;
-
-    if (bytes.length >= (length + offset)) {
-      return finishProcessingFn(err, {
-        response: bytes.subarray(offset, offset + length),
-        status: request.status,
-        uri: request.uri
-      });
-    }
-
-    // otherwise request sidx bytes
-    loader.request = xhr(sidxRequestOptions, finishProcessingFn);
-  });
-};
-
 export default class DashPlaylistLoader extends EventTarget {
   // DashPlaylistLoader must accept either a src url or a playlist because subsequent
   // playlist loader setups from media groups will expect to be able to pass a playlist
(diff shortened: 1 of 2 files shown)


src/dash-playlist-loader.js
```diff
@@ -295,6 +243,111 @@ export default class DashPlaylistLoader extends EventTarget {
     }
   }
 
+  requestErrored_(err, request, startingState) {
```
brandonocasey, line 246: triggers a request error if it finds an error and returns true if there was an error so request functions can exit early. We used this logic in many places in the original code and we use it in two places now.

src/dash-playlist-loader.js
```diff
@@ -295,6 +243,111 @@ export default class DashPlaylistLoader extends EventTarget {
...
+      this.trigger('error');
+      return true;
+    }
+  }
+
+  /**
+   * Verify that the container of the sidx segment can be parsed
+   * and if it can, get and parse that segment.
+   */
+  addSidxSegments_(playlist, startingState, cb) {
```
brandonocasey, line 279: All of the logic from `sidxRequestFinished` and `requestSidx` merged into one because they were really doing the same things anyway. I also added some extra logic here for things that we were doing in 3+ places outside of this function, but might as well have shared.

src/dash-playlist-loader.js
```diff
@@ -295,6 +243,111 @@ export default class DashPlaylistLoader extends EventTarget {
...
+      // keep this function async
+      this.mediaRequest_ = window.setTimeout(() => cb(false), 0);
+      return;
+    }
+
+    // resolve the segment URL relative to the playlist
+    const uri = resolveManifestRedirect(this.handleManifestRedirects, playlist.sidx.resolvedUri);
+    const sidxMapping = this.masterPlaylistLoader_.sidxMapping_;
+
+    sidxMapping[sidxKey] = {
```
brandonocasey, line 293: This is an addition, every function that requests sidx must also update the sidx mapping.

src/dash-playlist-loader.js
```diff
@@ -295,6 +243,111 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    };
+
+    const fin = (err, request) => {
+      if (this.requestErrored_(err, request, startingState)) {
+        return;
+      }
+
+      const sidx = parseSidx(toUint8(request.response).subarray(8));
+
+      sidxMapping[sidxKey].sidx = sidx;
```
brandonocasey, line 304: This line and the line below were previously done after the sidx request, but it just didn't make sense to do it everywhere we call this function. We will always want to add sidx segments to the playlist that we request sidx for, and we will always update our sidx mapping cache.

src/dash-playlist-loader.js
```diff
@@ -403,52 +420,10 @@ export default class DashPlaylistLoader extends EventTarget {
...
-        addSidxSegmentsToPlaylist(playlist, sidx, playlist.sidx.resolvedUri);
-
-        // everything is ready just continue to haveMetadata
-        this.haveMetadata({
-          startingState,
-          playlist
-        });
-      })
-    );
+    this.addSidxSegments_(playlist, startingState, (sidxChanged) => {
```
brandonocasey, line 423: Since this function now calls back for playlists that don't have sidx segments, we don't have to separate the logic.

src/dash-playlist-loader.js
```diff
@@ -700,110 +681,30 @@ export default class DashPlaylistLoader extends EventTarget {
...
-          this.updateMinimumUpdatePeriodTimeout_();
-
-          // TODO: do we need to reload the current playlist?
-          this.refreshMedia_(this.media().id);
-
-          return;
-        })
-      );
 
+      this.addSidxSegments_(this.media(), this.state, (sidxChanged) => {
```
brandonocasey, line 704: We don't have to check `this.media()` has sidx as `addSidxSegments_` will do that now, and the callback will be called.

test/dash-playlist-loader.test.js
```diff
@@ -1694,52 +1684,50 @@ QUnit.test('refreshXml_: updates media playlist reference if master changed', fu
...
-    stubDone.getCall(0).args[1].references,
-    'returned a parsed sidx box'
-  );
-  assert.strictEqual(
-    stubDone.getCall(0).args[1].references[0].referencedSize,
+  assert.ok(stubDone.getCall(0).args[0], 'sidx segments were added');
+  assert.ok(fakePlaylist.segments.length, 'added a parsed sidx segment to playlist');
+
+  assert.deepEqual(
+    sidxMapping['undefined-0-43'].sidx.references[0].referencedSize,
```
gesinger, line 1724: Should this have a proper uri?

src/dash-playlist-loader.js
```diff
@@ -285,16 +233,120 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    if (!this.request) {
+      return true;
+    }
+
+    // pending request is cleared
+    this.request = null;
+
+    if (err) {
+      // use the provided error or create one
+      // see requestSidx_ for the container request
```
gesinger, line 256: Now that this is reused, is this part of the comment relevant?

src/dash-playlist-loader.js
```diff
@@ -648,24 +620,26 @@ export default class DashPlaylistLoader extends EventTarget {
...
-        newMaster,
-        this.masterPlaylistLoader_.sidxMapping_
-      );
+    if (oldMaster) {
+      newMaster = updateMaster(oldMaster, newMaster, this.masterPlaylistLoader_.sidxMapping_);
     }
 
-    // only update master if we have a new master
-    this.masterPlaylistLoader_.master = newMaster ? newMaster : this.masterPlaylistLoader_.master;
+    // only update master, if we have a new master.
```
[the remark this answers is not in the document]
brandonocasey: must have gotten lost in one of the merges

src/dash-playlist-loader.js
```diff
@@ -285,16 +233,120 @@ export default class DashPlaylistLoader extends EventTarget {
...
+   * Verify that the container of the sidx segment can be parsed
+   * and if it can, get and parse that segment.
+   */
+  addSidxSegments_(playlist, startingState, cb) {
+    const sidxKey = playlist.sidx && generateSidxKey(playlist.sidx);
+
+    // playlist lacks sidx or sidx segments were added to this playlist already.
+    if (!playlist.sidx || !sidxKey || this.masterPlaylistLoader_.sidxMapping_[sidxKey]) {
+      // keep this function async
+      this.mediaRequest_ = window.setTimeout(() => cb(false), 0);
```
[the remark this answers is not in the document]
brandonocasey: I honestly wonder why we even need `mediaRequest_`. It seems to me like we only use it one place in the code to verify that a media change is pending. What if we just set `pentingMedia_` in `media()` and removed `mediaRequest`?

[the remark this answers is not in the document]
gesinger: I'm all for removing more things. Sounds good to me 👍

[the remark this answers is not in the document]
brandonocasey: After making the change it causes a lot of tests to fail. Seems like we should do one more refactor pull request for this change to keep the size of the pull request down.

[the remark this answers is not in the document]
gesinger: Was there any way to centralize the mediaRequest_ clearing (rather than removing it) without as many test failures?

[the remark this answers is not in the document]
brandonocasey: Not really as clearing it out in a central location and switching to `pendingMedia_` required similar changes, and had similar test failures.

gesinger (approved):

PR #1002, opened 2020-11-18
Wait for EME initialization before appending content

gesinger:
Requires https://github.com/videojs/videojs-contrib-eme/pull/124

## Description
Wait for EME initialization before appending content. This is particularly important for Chrome, where, if unencrypted content is appended before encrypted content and the key session has not been created, a MEDIA_ERR_DECODE will be thrown once the encrypted content is reached during playback.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/source-updater.js
+++ b/src/source-updater.js
@@ -79,7 +79,11 @@ const shiftQueue = (type, sourceUpdater) => {
   // Media source queue entries don't need to consider whether the source updater is
   // started (i.e., source buffers are created) as they don't need the source buffers, but
   // source buffer queue entries do.
-  if (!sourceUpdater.started_ || sourceUpdater.mediaSource.readyState === 'closed' || updating(type, sourceUpdater)) {
+  if (
+    !sourceUpdater.ready() ||
+    sourceUpdater.mediaSource.readyState === 'closed' ||
+    updating(type, sourceUpdater)
+  ) {
     return;
   }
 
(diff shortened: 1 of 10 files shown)


src/source-updater.js
```diff
@@ -331,24 +335,45 @@ export default class SourceUpdater extends videojs.EventTarget {
       // used for debugging
       this.audioError_ = e;
     };
-    this.started_ = false;
+    this.createdSourceBuffers_ = false;
+    this.initializedEme_ = false;
+  }
+
+  initializedEme() {
```
brandonocasey, line 342: maybe this isn't a big deal, but should the source updater be the one to worry about initialized eme? to me that seems like a no, but if its the only place we can do it in a shared way, I kinda understand.

src/videojs-http-streaming.js
```diff
@@ -288,15 +277,90 @@ const setupEmeOptions = ({
...
   // only place where it should not be deduped is for ms-prefixed APIs, but the early
   // return for IE11 above, and the existence of modern EME APIs in addition to
   // ms-prefixed APIs on Edge should prevent this from being a concern.
   // initializeMediaKeys also won't use the webkit-prefixed APIs.
   keySystemsOptionsArr.forEach((keySystemsOptions) => {
-    player.eme.initializeMediaKeys({
-      keySystems: keySystemsOptions
-    });
+    keySessionCreatedPromises.push(new Promise((resolve, reject) => {
+      player.tech_.one('keysessioncreated', () => {
```
brandonocasey, line 290: can just pass resolve, without the wrapper `() => {}`

brandonocasey (commented):
code mostly lgtm, I do have an issue though. Will this break ie 11 since it doesn't support promises at all? I think we do some polyfilling for ie 11 in eme, but not here.

src/source-updater.js
gesinger: I thought about that for a bit, and the reason I put it here was because it technically should worry about it, since Chrome's source buffers care about order of operations, source updater, being the closest to the source buffers, needs to worry as well.

brandonocasey (approved):
We will have to make sure to test drm content on IE 11 but the code looks good to me.

gkatsev (approved):
I tested this Chrome, Firefox, and Safari on macOS as well as Chrome, IE11, and Chromium Edge on Windows 10.
Works well everywhere. IE11 and Edge has the known issue where transitioning between clear content and DRMed content will not play the DRM content well.

PR #1003, opened 2020-11-18
test: fix tests on firefox 83

brandonocasey:
closing this in favor of a fork pr in #1003

PR #1004, opened 2020-11-18
Test/ff 83

brandonocasey:
See browserstack results in #1003

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,11 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: '${{ steps.nvm.outputs.NVMRC }}'
+    # turn off the default setup-node problem watchers...
+    - run: echo "::remove-matcher owner=eslint-compact::"
+    - run: echo "::remove-matcher owner=eslint-stylish::"
+    - run: echo "::remove-matcher owner=tsc::"
+
     - run: npm i --prefer-offline --no-audit
 
     # run safari on macos
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,10 @@ jobs:
       shell: bash
       id: nvm
 
+    - name: update apt cache on linux w/o browserstack
+      run: sudo apt-get update
+      if: ${{ startsWith(matrix.os, 'ubuntu') && !env.BROWSER_STACK_USERNAME }}
+
     - name: Install ffmpeg/pulseaudio for firefox on linux w/o browserstack
       run: sudo apt-get install ffmpeg pulseaudio
       if: ${{ startsWith(matrix.os, 'ubuntu') && !env.BROWSER_STACK_USERNAME }}
--- a/src/ad-cue-tags.js
+++ b/src/ad-cue-tags.js
@@ -5,6 +5,15 @@ import window from 'global/window';
 
 /**
  * Searches for an ad cue that overlaps with the given mediaTime
+ *
+ * @param {Object} track
+ *        the track to find the cue for
+ *
+ * @param {number} mediaTime
+ *        the time to find the cue at
+ *
+ * @return {Object|null}
+ *         the found cue or null
  */
 export const findAdCue = function(track, mediaTime) {
   const cues = track.cues;
--- a/src/bin-utils.js
+++ b/src/bin-utils.js
@@ -7,6 +7,7 @@
  *
  * @param {TimeRange} range the timerange to use for conversion
  * @param {number} i the iterator on the range to convert
+ * @return {string} the range in string format
  */
 const textRange = function(range, i) {
   return range.start(i) + '-' + range.end(i);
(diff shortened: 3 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -44,6 +44,11 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: '${{ steps.nvm.outputs.NVMRC }}'
+    # turn off the default setup-node problem watchers...
```
brandonocasey, line 47: Turns off "problem matchers" that add in linter warnings to unchanged files.

src/ad-cue-tags.js
```diff
@@ -5,6 +5,15 @@ import window from 'global/window';
...
+ *
+ * @param {Object} track
+ *        the track to find the cue for
+ *
+ * @param {number} mediaTime
+ *        the time to find the cue at
+ *
+ * @return {Object|null}
+ *         the found cue or null
  */
```
brandonocasey, line 17: We can remove jsdoc changes, but I already did them, so we might as well keep them.

src/source-updater.js
```diff
@@ -416,7 +416,7 @@ export default class SourceUpdater extends videojs.EventTarget {
     // IE reports that it supports removeSourceBuffer, but often throws
     // errors when attempting to use the function. So we report that it
     // does not support removeSourceBuffer.
-    return !videojs.browser.IE_VERSION && window.MediaSource &&
+    return !videojs.browser.IE_VERSION && !videojs.browser.IS_FIREFOX && window.MediaSource &&
```
brandonocasey, line 419: I think this is one of the fixes that we need. It seems like removeSourceBuffer is failing to work correctly in new versions of firefox.

test/playback.test.js
```diff
@@ -26,6 +17,19 @@ const playFor = function(player, time, cb) {
     }, 10);
   };
 
+  if (player.paused()) {
```
brandonocasey, line 20: Actually use the play promise. prevents timeouts for playback tests on firefox.

test/source-updater.test.js
```diff
@@ -44,13 +44,16 @@ const mp4AudioTotal = () => concatSegments(mp4AudioInit(), mp4Audio());
...
+    this.video = document.createElement('video');
+
+    this.fixture.appendChild(this.video);
 
     this.mediaSource = new window.MediaSource();
 
     // need to attach the real media source to a video element for the media source to
     // change to an open ready state
-    video.src = URL.createObjectURL(this.mediaSource);
+    this.video.src = URL.createObjectURL(this.mediaSource);
```
brandonocasey, line 56: Not sure if this helped or not, but we should be doing this cleanup.

test/playback.test.js
```diff
@@ -232,101 +236,101 @@ QUnit[testFn]('Big Buck Bunny audio only, groups & renditions same uri', functio
...
-    assert.equal(player.error(), null, 'no errors');
+    playFor(player, 2, function() {
+      assert.ok(true, 'played for at least two seconds');
+      assert.equal(player.error(), null, 'no errors');
 
-    player.one('ended', () => {
-      assert.ok(true, 'triggered ended event');
-      done();
-    });
+      player.one('ended', () => {
```
brandonocasey, line 291: All tests still exist here, some were just moved down to be a part of the block that skips tests on firefox.

src/bin-utils.js
```diff
@@ -17,6 +18,7 @@ const textRange = function(range, i) {
  *
  * @param {number} e The number
  * @param {number} i the iterator
+ * @return {string} the hex formated number as a string
```
gesinger, line 21:
```suggestion
 * @return {string} the hex formatted number as a string
```

test/playback.test.js
```diff
@@ -23,9 +14,22 @@ const playFor = function(player, time, cb) {
...
       cb();
-    }, 10);
+    }, 100);
   };
 
+  if (player.paused()) {
+    const playPromise = player.play();
+
+    // Catch/silence error when a pause interrupts a play request
+    // on browsers which return a promise
```
gesinger, line 24: Might be worth adding in the comment above here (and removing this, since we no longer silence the error)

test/playback.test.js
```diff
@@ -232,101 +236,101 @@ QUnit[testFn]('Big Buck Bunny audio only, groups & renditions same uri', functio
...
-    src: 'https://d2zihajmogu5jn.cloudfront.net/audio-only-dupe-groups/prog_index.m3u8',
-    type: 'application/x-mpegURL'
+    src: 'https://livesim.dashif.org/livesim/mup_30/testpic_2s/Manifest.mpd',
+    type: 'application/dash+xml'
   });
 });
 
-QUnit[testFn]('Big Buck Bunny Demuxed av, audio only rendition same as group', function(assert) {
-  const done = assert.async();
+if (!videojs.browser.IS_FIREFOX) {
```
gesinger, line 244: Would be worth adding a comment above here

PR #1006, opened 2020-11-19
chore: don't run tests on release

gkatsev:


gesinger (approved):
Tests run on CI and elsewhere. Local tests have not been as consistent and reliable. Hopefully we'll get to fixing that too, but CI and elsewhere are more reliable for versioning.

brandonocasey (approved):

PR #1015, opened 2020-12-01
fix: prevent double source buffer ready on ie 11

brandonocasey:
Fixes one broken aspect of https://github.com/videojs/http-streaming/pull/1002 that can cause a double sourceUpdater ready on ie 11.


gesinger (commented):
Do you think it would be worth having a setTimeout in the createdsourcebuffers for IE11 so that that code always runs asynchronously instead of knowing about the behavior here?

brandonocasey:
Updated the code slightly here so that we can trigger ready before `createdsourcebuffers` that way we won't double trigger `ready` if created source buffers also initializes eme syncronously.

gesinger (approved):

PR #1033, opened 2020-12-10
chore: better worker build

brandonocasey:
## Description
I wrote [rollup-plugin-worker-factory](https://github.com/brandonocasey/rollup-plugin-worker-factory) for Mux.js so that we could support nodejs workers and browser workers using the same plugin. It has side benefits of:
1. not needing a separate build target for workers that has to run before other builds
2. We can share a rollup cache between the worker and main builds
3. Worker code doesn't have to be wrapped in a strange function or export anything.
4. The new rollup plugin is much more readable than our current plugin.
5. We can support add a debug build for web workers that run in the same thread and not a WebWorker blob for debugging.

### DIFF
--- a/scripts/rollup.config.js
+++ b/scripts/rollup.config.js
@@ -4,7 +4,7 @@ const {terser} = require('rollup-plugin-terser');
 const createTestData = require('./create-test-data.js');
 const replace = require('@rollup/plugin-replace');
 
-let mockWorker;
+let syncWorker;
 // see https://github.com/videojs/videojs-generate-rollup-config
 // for options
 const options = {
@@ -63,7 +63,7 @@ const options = {
       defaults.babel
     ]});
 
-    mockWorker = worker({type: 'mock', plugins: [
+    defaults.syncWorker = syncWorker = worker({type: 'mock', plugins: [
       defaults.resolve,
       defaults.json,
       defaults.commonjs,
@@ -100,7 +100,7 @@ if (config.builds.browser) {
     }
   });
 
-  config.builds.syncWorkers.plugins[0] = mockWorker;
+  config.builds.syncWorkers.plugins[0] = syncWorker;
 }
 
 // Add additonal builds/customization here!
--- a/scripts/rollup.config.js
+++ b/scripts/rollup.config.js
@@ -1,5 +1,5 @@
 const generate = require('videojs-generate-rollup-config');
-const worker = require('@gkatsev/rollup-plugin-bundle-worker');
+const worker = require('rollup-plugin-worker-factory');
 const {terser} = require('rollup-plugin-terser');
 const createTestData = require('./create-test-data.js');
 const replace = require('@rollup/plugin-replace');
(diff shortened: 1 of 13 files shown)


src/decrypter-worker.js
```diff
@@ -6,43 +6,36 @@ import { createTransferableMessage } from './bin-utils';
...
-      data.key.bytes,
-      data.key.byteOffset,
-      data.key.byteLength / 4
-    );
-    const iv = new Uint32Array(
-      data.iv.bytes,
-      data.iv.byteOffset,
-      data.iv.byteLength / 4
-    );
+self.onmessage = function(event) {
```
brandonocasey, line 10: We no longer wrap in a function here, instead the code looks exactly as it would running on a web worker, inside of the main script body.

src/transmuxer-worker.js
```diff
@@ -411,23 +411,19 @@ class MessageHandlers {
  *
  * @param {Object} self the scope for the web worker
  */
-const TransmuxerWorker = function(self) {
-  self.onmessage = function(event) {
-    if (event.data.action === 'init' && event.data.options) {
-      this.messageHandlers = new MessageHandlers(self, event.data.options);
-      return;
-    }
+self.onmessage = function(event) {
```
brandonocasey, line 414: again no longer wrapped in a function.

test/loader-common.js
```diff
@@ -13,7 +13,7 @@ import { MasterPlaylistController } from '../src/master-playlist-controller';
 import SourceUpdater from '../src/source-updater';
 import SyncController from '../src/sync-controller';
 import TimelineChangeController from '../src/timeline-change-controller';
-import Decrypter from 'worker!../src/decrypter-worker.worker.js';
+import Decrypter from 'worker!../src/decrypter-worker.js';
```
brandonocasey, line 16: slight change since we no longer pre-bundle the worker.

scripts/rollup.config.js
```diff
@@ -20,14 +20,15 @@ const options = {
...
       ])
     });
   },
   plugins(defaults) {
-    defaults.module.splice(2, 0, 'worker');
-    defaults.browser.splice(2, 0, 'worker');
-    defaults.test.splice(3, 0, 'worker');
+    defaults.module.splice(0, 0, 'worker');
+    defaults.browser.splice(0, 0, 'worker');
+    defaults.test.splice(1, 0, 'worker');
```
gesinger, line 31: Just curious if there's a reason this has to be spliced in at 1 (maybe a comment might help)

src/transmuxer-worker.js
```diff
@@ -1,6 +1,6 @@
 /* global self */
 /**
- * @file transmuxer-worker.js
+ * @file transmuxer.js
```
gesinger, line 3: I think this is still transmuxer-worker.js

scripts/rollup.config.js
```diff
@@ -26,11 +27,11 @@ const options = {
...
   plugins(defaults) {
-    defaults.module.splice(2, 0, 'worker');
-    defaults.browser.splice(2, 0, 'worker');
-    defaults.test.splice(3, 0, 'worker');
-
-    defaults.test.splice(0, 0, 'createTestData');
+    // add worker and createTestData to the front of plugin lists
+    defaults.module.unshift('worker');
+    defaults.browser.unshift('worker');
+    defaults.test.unshift('worker');
```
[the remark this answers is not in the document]
brandonocasey: sure I'll do that.

gesinger (approved):

PR #1038, opened 2020-12-18
fix: if a playlist has been retried less than half target duration, delay retry

### LINKED ISSUE #975: Player enters tight infinite loop when live rendition manifests stop updating

gkatsev:
We have a case where if a playlist stops refreshing, we will exclude the playlist and try switching to another playlist. During regular playback, we want to immediately download this new playlist because we want to reduce rebuffering. However, if all playlists are in this case, we will re-include the other playlists and retry them. However, if it will fail, so, we will once again, exclude it, re-include the other one and try again. With 2 playlists, it will ping pong between the two incredibly fast.
This is fine if there's only one playlist because we have a special case to delay during a final rendition, and we don't have any other playlists to re-include.

This PR introduces a `lastRefresh` property on playlist objects, which gets updated each time it is downloaded. Then, when we need to download a new playlist as part of its exclusion, we check to see whether the last time it was refreshed was less than half target duration, with a default of 5 seconds. If it has been less time, we will delay by half target duration, otherwise, we will make the request immediately.

Fixes #975.


src/playlist-loader.js
```diff
@@ -312,11 +314,11 @@ export default class PlaylistLoader extends EventTarget {
     *
     * @param {Object=} playlist the parsed media playlist
     * object to switch to
-    * @param {boolean=} is this the last available playlist
+    * @param {boolean=} shouldDelay whether we should delay the request by half target duration
     *
     * @return {Playlist} the current loaded media
     */
-  media(playlist, isFinalRendition) {
+  media(playlist, shouldDelay) {
```
brandonocasey, line 321: way better name! 👍

src/playlist-loader.js
```diff
@@ -251,6 +251,8 @@ export default class PlaylistLoader extends EventTarget {
       customTagMappers: this.customTagMappers
     });
 
+    playlist.lastRefresh = Date.now();
```
gesinger, line 254: It looks like this is updated both in this function and within a different XHR request below. It may be worth centralizing it by creating a `requestManifest` wrapping function that takes in xhr options and a callback. It may also make testing easier.

gesinger (commented):
Probably worth adding a test or two.

src/playlist-loader.js
```diff
@@ -414,6 +416,8 @@ export default class PlaylistLoader extends EventTarget {
         return;
       }
 
+      playlist.lastRefresh = Date.now();
```
brandonocasey, line 419: Should we be unsetting lastRefresh somewhere? Maybe before the request in case if fails?

brandonocasey (commented):
Would be good to have a test for this.

src/playlist-loader.js
gkatsev: No need, we want to know when the last time we tried to reach this manifest. Maybe `lastRefresh` should be `lastRetry` instead?

src/playlist-loader.js
gkatsev: I think this one is a bit less necessary than the one below, I even considered removing it, but there's one codepath that leads here that isn't from the XHR below.
If it's useful for testing I'll add it. Otherwise, since it's a single line and only in two places, I think it's fine to leave as is. If we'd need it in another place then definitely worth refactoring it out.

test/manifests/double.m3u8
```diff
@@ -0,0 +1,6 @@
+# A simple main playlist with 2 media playlists
+#EXTM3U
+#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=240000,RESOLUTION=396x224
```
gesinger, line 3: We can probably remove `PROGRAM-ID` since it's no longer in the spec (same for other occurrences)

test/master-playlist-controller.test.js
```diff
@@ -5718,3 +5718,163 @@ QUnit.test('false without nextPlaylist', function(assert) {
 
   this.env.log.warn.callCount = 0;
 });
+
+QUnit.module('MasterPlaylistController blacklistCurrentPlaylist', sharedHooks);
+
+QUnit.test('dont exclude only playlist unless it was excluded forever', function(assert) {
```
gesinger, line 5724:
```suggestion
QUnit.test('don\'t exclude only playlist unless it was excluded forever', function(assert) {
```

test/master-playlist-controller.test.js
```diff
@@ -5718,3 +5718,163 @@ QUnit.test('false without nextPlaylist', function(assert) {
...
+  let mpc = this.masterPlaylistController;
+  let mpl = mpc.masterPlaylistLoader_;
+  let playlist = mpl.master.playlists[0];
+  let shouldDelay = false;
+
+  mpl.load = (delay) => (shouldDelay = delay);
+
+  mpc.blacklistCurrentPlaylist();
+
+  assert.notOk('excludeUntil' in playlist, 'playlist was not excluded since excludeDuration was finite');
```
gesinger, line 5751: Technically this is only saying `playlist not was not excluded`. I'm not sure if there's any check around the `blacklistDuration` here

test/master-playlist-controller.test.js
```diff
@@ -5718,3 +5718,163 @@ QUnit.test('false without nextPlaylist', function(assert) {
...
+  assert.ok(shouldDelay, 'we delay retry since it is a final rendition');
```
gesinger, line 5752:
```suggestion
  assert.ok(shouldDelay, 'we delay retry since it is the final rendition');
```

test/master-playlist-controller.test.js
```diff
@@ -5718,3 +5718,163 @@ QUnit.test('false without nextPlaylist', function(assert) {
...
+  mpl.load = (delay) => (shouldDelay = delay);
+  mpc.on('error', () => {
+    assert.ok(true, 'we triggered a playback error');
+  });
+
+  // exclude forever
+  mpc.blacklistCurrentPlaylist({}, Infinity);
+
+  assert.ok('excludeUntil' in playlist, 'playlist was excluded');
+  assert.equal(this.env.log.error.callCount, 1, 'logged an error');
```
gesinger, line 5785: Should we have a check down here to ensure that `load` wasn't called?

test/master-playlist-controller.test.js
```diff
@@ -5718,3 +5718,163 @@ QUnit.test('false without nextPlaylist', function(assert) {
...
+  const playlist = mpl.master.playlists[0];
+  let playlist2 = mpl.master.playlists[1];
+  let shouldDelay = false;
+
+  mpl.load = (delay) => (shouldDelay = delay);
+
+  mpc.blacklistCurrentPlaylist();
+
+  assert.ok('excludeUntil' in playlist, 'playlist was excluded since there is another playlist');
+  assert.notOk(shouldDelay, 'we do not delay retry since it is not a final rendition');
```
gesinger, line 5820:
```suggestion
  assert.notOk(shouldDelay, 'we do not delay retry since it is not the final rendition');
```

test/master-playlist-controller.test.js
```diff
@@ -5718,3 +5718,163 @@ QUnit.test('false without nextPlaylist', function(assert) {
...
+  // media1
+  this.standardXHRResponse(this.requests.shift());
+  playlist2 = mpl.master.playlists[1];
+
+  mpc.blacklistCurrentPlaylist();
+
+  assert.ok('excludeUntil' in playlist2, 'playlist2 was excluded');
+  assert.notOk('excludeUntil' in playlist, 'playlist was re-included');
+  assert.equal(this.env.log.warn.callCount, 3, 'logged another warning');
+  assert.ok(this.env.log.warn.calledWithMatch('exclusion'), 'we logged a warning that we reincluded playlists');
```
gesinger, line 5834: Very minor, but it is a bit strange to see that we check for reinclusion based on presence of exclusion. Maybe keep `Removing other playlists from the exclusion list`?

test/master-playlist-controller.test.js
```diff
@@ -5718,3 +5718,163 @@ QUnit.test('false without nextPlaylist', function(assert) {
...
+
+  this.env.log.warn.callCount = 0;
+});
+
+QUnit.test('should delay loading of new playlist if lastRefresh was less than half target duration', function(assert) {
+  // set time to be not 0
+  // this.clock.tick(100000);
```
gesinger, line 5841: I think this was left in

gesinger (commented):
Might be better to name the manifests as one-rendition and two-renditions or something like that for clarity.

test/master-playlist-controller.test.js
gkatsev: I guess I can verify that excludeUntil was set to a finite number.

test/master-playlist-controller.test.js
gkatsev: Yeah, I just didn't feel like copying the entire sentence. I can do it if you think it'll be better.

test/master-playlist-controller.test.js
gkatsev: Oh, misread.
The blacklistDuration is either passed in on the error, the second argument, or is on the MPC. The default value is `300`. I think given that we expect the default value to be used, this should be good enough.

test/master-playlist-controller.test.js
gkatsev: yeah, let me change the load override to be an assert instead.

test/master-playlist-controller.test.js
gesinger: I think it might be worth it, if only because of clarity, and if we ever add another log with `exclusion` it could silently allow this test to pass if it's inadvertently broken.

test/master-playlist-controller.test.js
```diff
@@ -5718,3 +5718,164 @@ QUnit.test('false without nextPlaylist', function(assert) {
...
+  mpl = mpc.masterPlaylistLoader_;
+  playlist = mpl.master.playlists[0];
+  shouldDelay = false;
+
+  mpl.load = (delay) => {
+    shouldDelay = delay;
+    assert.ok(false, 'load should not be called in this case');
+  };
+  mpc.on('error', () => {
+    assert.ok(true, 'we triggered a playback error');
```
gesinger, line 5781: We may want an `assert.expect` at the top, since this could not run and the test would still pass.

gesinger (commented):
It may still be worth renaming double.m3u8 to something like `two-renditions.m3u8` and single.m3u8 to something like `one-rendition.m3u8`

test/master-playlist-controller.test.js
gkatsev: this test has an expect at the top now, and it accounts for this one.

test/master-playlist-controller.test.js
```diff
@@ -5839,9 +5839,7 @@ QUnit.test('switch playlists if current playlist gets excluded and re-include if
   assert.notOk('excludeUntil' in playlist, 'playlist was re-included');
   assert.equal(this.env.log.warn.callCount, 3, 'logged another warning');
   assert.ok(
-    this.env.log.warn.calledWith(
-      'Removing other playlists from the exclusion list because the last rendition is about to be excluded.'
-    ),
+    this.env.log.warn.calledWith('Removing other playlists from the exclusion list because the last rendition is about to be excluded.'),
```
gkatsev, line 5842: the linter did this.

gesinger (approved):

PR #1039, opened 2020-12-22
fix: continue retrying playlist if even if it fails when live

gkatsev:
For Live HLS, even if a playlist errors out during playback, we want to continue retrying it because it's possible it'll recover.


gkatsev:
This needs to be done another way as it causes a couple of other minor issues.

PR #1045, opened 2021-01-13
fix: store queue and current transmux on transmuxer instead of global

brandonocasey:
## Problem
So this is somewhat of a bigger issue then I originally realized when I found an issue in tests for #1043 . The segmentTransmuxer stores a global cache of the currentTransmux and transmux queues. This causes a few issues:

1. Separate audio/video transmuxes will bottleneck on another
2. When running two or more videos on a page each video has a chance to bottleneck all the others with it's transmuxQueue and currentTransmux
3. Tests/dispose is flakey because mediaSegmentRequest sets currentTransmux/transmuxQueue but only segmentLoader disposes it.

## Changes
Store the transmuxQueue and currentTransmux on the transmuxer that is being used rather than keeping a global list.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -528,14 +528,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       time: 0
     };
 
-    this.transmuxer_ = segmentTransmuxer.createTransmuxer({
-      remux: false,
-      alignGopsAtEnd: this.safeAppend_,
-      keepOriginalTimestamps: true,
-      handlePartialData: this.handlePartialData_,
-      parse708captions: this.parse708captions_
-    });
-
+    this.transmuxer_ = this.createTransmuxer_();
     this.triggerSyncInfoUpdate_ = () => this.trigger('syncinfoupdate');
     this.syncController_.on('syncinfoupdate', this.triggerSyncInfoUpdate_);
 
@@ -593,6 +586,16 @@ export default class SegmentLoader extends videojs.EventTarget {
         }
       });
     }
+  }
+
+  createTransmuxer_() {
+    return segmentTransmuxer.createTransmuxer({
+      remux: false,
+      alignGopsAtEnd: this.safeAppend_,
+      keepOriginalTimestamps: true,
+      handlePartialData: this.handlePartialData_,
+      parse708captions: this.parse708captions_
+    });
   }
 
   /**
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -7,7 +7,6 @@ import Config from './config';
 import window from 'global/window';
 import { initSegmentId, segmentKeyId } from './bin-utils';
 import { mediaSegmentRequest, REQUEST_ERRORS } from './media-segment-request';
-import TransmuxWorker from 'worker!./transmuxer-worker.js';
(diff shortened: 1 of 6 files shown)


src/segment-loader.js
```diff
@@ -590,23 +595,6 @@ export default class SegmentLoader extends videojs.EventTarget {
     }
   }
 
-  createTransmuxer_() {
```
brandonocasey, line 593: This moved to `segmentTransmuxer.createTransmuxer` for a few reasons: 
1. We use it in tests in exactly the same manner
2. We need to add some special properties to the transmuxer when we create it in segmentTransmuxer to prevent state from being kept in the segmentTransmuxer itself
3. We need to clear those same properties on terminate, which we do by wrapping terminate in segmentTransmuxer
4. segmentTransmuxer already handles all the other transmuxer message logic, so having logic here for it is weird.

src/segment-loader.js
```diff
@@ -632,9 +620,6 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.abort_();
     if (this.transmuxer_) {
       this.transmuxer_.terminate();
-      // Although it isn't an instance of a class, the segment transmuxer must still be
-      // cleaned up.
-      segmentTransmuxer.dispose();
```
brandonocasey, line 637: dispose is gone, segmentTransmuxer stores no state.

src/segment-transmuxer.js
```diff
@@ -1,5 +1,4 @@
-const transmuxQueue = [];
```
brandonocasey, line 1: now stored on transmuxer, not global.

src/segment-transmuxer.js
```diff
@@ -66,30 +65,31 @@ export const handleGopInfo_ = (event, transmuxedData) => {
...
+    onDone
+  } = options;
   const transmuxedData = {
     isPartial,
     buffer: []
   };
 
   const handleMessage = (event) => {
-    if (!currentTransmux) {
+    if (transmuxer.currentTransmux !== options) {
```
brandonocasey, line 92: We need access to options above so that we can verify that we haven't been disposed.

src/segment-transmuxer.js
```diff
@@ -187,30 +187,30 @@ export const processTransmux = ({
...
 
-export const dequeue = () => {
-  currentTransmux = null;
-  if (transmuxQueue.length) {
-    currentTransmux = transmuxQueue.shift();
-    if (typeof currentTransmux === 'function') {
-      currentTransmux();
+export const dequeue = (transmuxer) => {
+  transmuxer.currentTransmux = null;
+  if (transmuxer.transmuxQueue && transmuxer.transmuxQueue.length) {
```
gesinger, line 192: Does this need to check for `transmuxer.transmuxQueue`?

brandonocasey: nope, I think I missed this one when I made `createTransmuxer` something that we use in tests and segment-transmuxer. Originally I had null checks everywhere but if we always rely on `createTransmuxer` from this file then it will always exist.i

gesinger (approved):

gesinger (approved):

PR #1051, opened 2021-01-25
Consider object-fit when selecting playlist by player size 

tf:
## Description

So far, when `limitRenditionByPlayerDimensions` is `true`,
`simpleSelector` tried to either find a rendition with a resolution
that matches the size of the player exactly or, if that does not
exist, a rendition with the smallest resolution that has either
greater width or greater height than the player. This makes sense
since by default the video will be scaled to fit inside the media
element. So every resolution that exceeds player size in at least one
dimension will be scaled down.

Most browsers [support](https://caniuse.com/?search=object-fit) customizing this scaling behavior via the
[`object-fit` CSS property](https://developer.mozilla.org/de/docs/Web/CSS/object-fit). If it set to `cover`, the video will
instead be scaled up if video and player aspect ratio do not match.

The previous behavior caused renditions with low resolution to be
selected for players with small width (e.g. portrait phone aspect
ratio) even when videos were then scaled up to cover the whole player.

We therefore detect if `object-fit` is set to `cover` and instead
select the smallest rendition with a resolution that exceeds player
dimensions in both width and height.

## Specific Changes proposed

The first two commits are preparatory refactorings. The third commit contains the change itself.

## Requirements Checklist
- [x] Feature implemented
  - [x] Docs/guides updated
  - [x] Example created (see comment below)


tf:
Both `CI_TEST_TYPE=playback npm run test` and `CI_TEST_TYPE=unit npm run test` pass locally. Not sure if failure of playback test in CI related.

tf:
Regarding docs: A note about the feature could be added to the [docs of `limitRenditionByPlayerDimensions` in the readme](https://github.com/videojs/http-streaming#limitrenditionbyplayerdimensions). The [bitrate switching guide](https://github.com/videojs/http-streaming/blob/main/docs/bitrate-switching.md) also seems relevant, feels a bit outdated already, though.

tf:
Here's an example show casing the new behavior using a build that includes the changes from the PR branch.

https://jsbin.com/xeqirapani/edit?html,output

tf:
Could somebody take a look at this PR?

src/playlist-selectors.js
```diff
@@ -133,27 +133,35 @@ export const comparePlaylistResolution = function(left, right) {
 /**
  * Chooses the appropriate media playlist based on bandwidth and player size
  *
- * @param {Object} master
+ * @param {Object} settings
```
gkatsev, line 136: thanks for updating the jsdoc!

src/playlist-selectors.js
```diff
@@ -133,27 +133,35 @@ export const comparePlaylistResolution = function(left, right) {
...
  */
-export const simpleSelector = function(
-  master,
-  playerBandwidth,
-  playerWidth,
-  playerHeight,
-  limitRenditionByPlayerDimensions
-) {
+export const simpleSelector = function(settings) {
+  const {
```
gkatsev, line 157: At first, I wasn't sure about this PR because there were a lot of changes that seemed unrelated to the requested feature, however, upon closer inspection, I see that it's only changing a lot of arguments into an options object. Given that simpleSelector is only exported for tests, that seems fine to do.

test/playlist-selectors.test.js
```diff
@@ -203,11 +203,28 @@ test('simpleSelector limits using resolution information when it exists', functi
     master,
     bandwidth: 4194304,
     playerWidth: 444,
-    playerHeight: 790,
+    playerHeight: 500,
```
gkatsev, line 206: why change the player height here?

src/playlist-selectors.js
```diff
@@ -315,13 +334,14 @@ export const simpleSelector = function(
...
-    parseInt(safeGetComputedStyle(this.tech_.el(), 'width'), 10) * pixelRatio,
-    parseInt(safeGetComputedStyle(this.tech_.el(), 'height'), 10) * pixelRatio,
-    this.limitRenditionByPlayerDimensions
-  );
+  return simpleSelector({
+    master: this.playlists.master,
+    bandwidth: this.systemBandwidth,
+    playerWidth: parseInt(safeGetComputedStyle(this.tech_.el(), 'width'), 10) * pixelRatio,
+    playerHeight: parseInt(safeGetComputedStyle(this.tech_.el(), 'height'), 10) * pixelRatio,
+    playerObjectFit: safeGetComputedStyle(this.tech_.el(), 'objectFit'),
```
gkatsev, line 342: I think `playerObjectFit` should be a config option like `limitRenditionByPlayerDimensions` or `useDevicePixelRatio` for people to opt into. Changes in the ABR algorithm like this could potentially affect people's bandwidth usage and we try to have it be opt-in first and then eventually can be turned on by default (or made non-configurable).

tf: I've pushed a commit to add a `usePlayerObject` option.

test/playlist-selectors.test.js
tf: For the object-fit case, I wanted to test that it selects the smallest rendition that has both at least player width and player height. With the old player height value there would have been no rendition with greater height than the player height. So I needed a smaller player height to make sure it does not select a too big rendition.

I wanted the both test cases to only differ regarding "object-fit", to make it clear that that's what's causing the different result. Lowering the value here does not change anything since the rendition in chosen based on player width in this case.

src/playlist-selectors.js
```diff
@@ -315,13 +338,15 @@ export const simpleSelector = function(
...
```
brandonocasey, line 346: I think that we might be better to pass an empty string for `playerObjectFit` if `usePlayerObjectFit` is `false` and pass the actual value if `usePlayerObjectFit` is `true`. That keeps `simpleSelector` from having to worry about it at all.

test/playlist-selectors.test.js
brandonocasey: perhaps it would be better to change behavior in a new test and keep this test as (except for the change to an options object for simpleSelector)

brandonocasey (commented):
I also see the tests failing due to a linter error here, not sure why it would fail in `scripts/index-demo-page.js` as you didn't make any changes there. Perhaps a rebase would fix that?

src/playlist-selectors.js
tf: This would mean duplicating the logic in `movingAverageBandwidthSelector` and `lastBandwidthSelector`, though, where it not covered by a test - like the pixel ratio logic. Happy to change it if you feel strongly about it.

tf:
Changed the test case to use the original value again. Rebased onto main for linting fix.

adrums86 (approved):
Seems like a reasonable addition. I think we should get this into the main branch.

tf:
Rebased onto main. Updated jsbin showcasing the new behavior using the latest version of Video.js:

https://jsbin.com/vipikadeni/edit?html,output

dzianis-dashkevich (approved):

dzianis-dashkevich:
@tf thanks! < 5 years :)

PR #1055, opened 2021-01-27
feat: use EXT-X-PART for LL-HLS

brandonocasey:
## Description
Implements support for EXT-X-PART and EXT-X-PRELOAD-HINT
Requires https://github.com/videojs/m3u8-parser/pull/137

## Future pull requests
* ~~Use `#EXT-X-PART-INF`, query parameters for delta updates, `#EXT-X-SERVER-CONTROL`, and `#EXT-X-SKIP`~~ #1079 
* ~~Use `#EXT-X-PRELOAD-HINT` aka `preloadSegment` on the manifest~~ #1078 

## Investigation
* How would we use `#EXT-X-RENDITION-REPORT` when switching renditions? Is it only relevant in the context of delta updates?

### DIFF
--- a/scripts/index-demo-page.js
+++ b/scripts/index-demo-page.js
@@ -386,7 +386,7 @@
               overrideNative: getInputValue(stateEls['override-native']),
               handlePartialData: getInputValue(stateEls.partial),
               experimentalBufferBasedABR: getInputValue(stateEls['buffer-water']),
-              llhls: getInputValue(stateEls.llhls)
+              experimentalLLHLS: getInputValue(stateEls.llhls)
             }
           }
         });
--- a/scripts/index-demo-page.js
+++ b/scripts/index-demo-page.js
@@ -241,6 +241,7 @@
     'sync-workers',
     'liveui',
     'partial',
+    'llhls',
     'url',
     'type',
     'keysystems',
@@ -298,6 +299,13 @@
     stateEls.debug.addEventListener('change', function(event) {
       saveState();
       window.videojs.log.level(event.target.checked ? 'debug' : 'info');
+    });
+
+    stateEls.llhls.addEventListener('change', function(event) {
+      saveState();
+
+      // reload the player and scripts
+      stateEls.minified.dispatchEvent(newEvent('change'));
     });
 
     stateEls.partial.addEventListener('change', function(event) {
@@ -377,7 +385,8 @@
             vhs: {
               overrideNative: getInputValue(stateEls['override-native']),
               handlePartialData: getInputValue(stateEls.partial),
-              experimentalBufferBasedABR: getInputValue(stateEls['buffer-water'])
+              experimentalBufferBasedABR: getInputValue(stateEls['buffer-water']),
+              llhls: getInputValue(stateEls.llhls)
             }
           }
         });
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -26,7 +26,8 @@ export const parseManifest = ({
   oninfo,
   manifestString,
   customTagParsers = [],
-  customTagMappers = []
+  customTagMappers = [],
+  llhls
 }) => {
   const parser = new M3u8Parser();
 
@@ -43,7 +44,35 @@ export const parseManifest = ({
   parser.push(manifestString);
   parser.end();
 
-  return parser.manifest;
+  const manifest = parser.manifest;
+
+  // remove llhls features from the parsed manifest
+  if (!llhls) {
+    [
+      'preloadSegment',
+      'skip',
+      'serverControl',
+      'renditionReports',
+      'partInf',
+      'partTargetDuration'
+    ].forEach(function(k) {
+      if (manifest.hasOwnProperty(k)) {
+        delete manifest[k];
+      }
+    });
+
+    if (manifest.segments) {
+      manifest.segments.forEach(function(segment) {
+        ['parts', 'preloadHints'].forEach(function(k) {
+          if (segment.hasOwnProperty(k)) {
+            delete segment[k];
+          }
+        });
+      });
+    }
+  }
+
+  return manifest;
 };
 
 /**
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -12,12 +12,18 @@ export const createPlaylistID = (index, uri) => {
 /**
  * Parses a given m3u8 playlist
  *
+ * @param {Function} [onwarn]
+ *        a function to call when the parser triggers a warning event.
+ * @param {Function} [oninfo]
+ *        a function to call when the parser triggers an info event.
  * @param {string} manifestString
  *        The downloaded manifest string
  * @param {Object[]} [customTagParsers]
  *        An array of custom tag parsers for the m3u8-parser instance
  * @param {Object[]} [customTagMappers]
- *         An array of custom tag mappers for the m3u8-parser instance
+ *        An array of custom tag mappers for the m3u8-parser instance
+ * @param {boolean} [llhls=false]
+ *        Wether to keep ll-hls features in the manifest after parsing.
  * @return {Object}
  *         The manifest object
  */
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -22,7 +22,7 @@ export const createPlaylistID = (index, uri) => {
  *        An array of custom tag parsers for the m3u8-parser instance
  * @param {Object[]} [customTagMappers]
  *        An array of custom tag mappers for the m3u8-parser instance
- * @param {boolean} [llhls=false]
+ * @param {boolean} [experimentalLLHLS=false]
  *        Wether to keep ll-hls features in the manifest after parsing.
  * @return {Object}
  *         The manifest object
@@ -33,7 +33,7 @@ export const parseManifest = ({
   manifestString,
   customTagParsers = [],
   customTagMappers = [],
-  llhls
+  experimentalLLHLS
 }) => {
   const parser = new M3u8Parser();
 
@@ -53,7 +53,8 @@ export const parseManifest = ({
   const manifest = parser.manifest;
 
   // remove llhls features from the parsed manifest
-  if (!llhls) {
+  // if we don't want llhls support.
+  if (!experimentalLLHLS) {
     [
       'preloadSegment',
       'skip',
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -23,7 +23,7 @@ export const createPlaylistID = (index, uri) => {
  * @param {Object[]} [customTagMappers]
  *        An array of custom tag mappers for the m3u8-parser instance
  * @param {boolean} [experimentalLLHLS=false]
- *        Wether to keep ll-hls features in the manifest after parsing.
+ *        Whether to keep ll-hls features in the manifest after parsing.
  * @return {Object}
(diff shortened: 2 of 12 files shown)


src/media-segment-request.js
```diff
@@ -971,7 +971,7 @@ export const mediaSegmentRequest = ({
   }
 
   const segmentRequestOptions = videojs.mergeOptions(xhrOptions, {
-    uri: segment.resolvedUri,
+    uri: segment.part && segment.part.resolvedUri || segment.resolvedUri,
```
brandonocasey, line 974: we don't actually use uri anywhere :shrug: but for completeness sake I changed it.

src/segment-loader.js
```diff
@@ -975,6 +976,22 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        this.partIndex = null;
+      } else {
+        const segment = this.playlist_.segments[this.mediaIndex];
+
+        // partIndex should remain the same for the same segment
+        // unless parts fell off of the playlist for this segment.
+        // In that case we need to reset partIndex and resync
+        if (this.partIndex && (!segment.parts || !segment.parts.length || !segment.parts[this.partIndex])) {
+          this.logger_(`part fell off on part ${this.partIndex}`);
+          this.resetLoader();
```
brandonocasey, line 992: Not sure of a better way to do this. If we appended part 4 of 8 but parts fell off the manifest, they will no longer be request-able, so we kind of have to reset partIndex and the buffer. Perhaps the solution is to never use the first sequence of parts in a manifest?

test/test-helpers.js
```diff
@@ -470,15 +475,23 @@ export const playlistWithDuration = function(time, conf) {
       segment.discontinuity = true;
     }
 
+    // add parts for the the last 3 segments in llhls playlists
+    if (conf && conf.llhls && (count - i) <= 3) {
```
brandonocasey, line 479: add parts if llhls is set.

test/test-helpers.js
```diff
@@ -433,10 +433,14 @@ export const playlistWithDuration = function(time, conf) {
...
 
+  if (conf && conf.llhls) {
+    result.partTargetDuration = conf.llhls.partTargetDuration || (targetDuration / 5);
+  }
+
   result.id = result.uri;
 
-  const count = Math.floor(time / targetDuration);
   const remainder = time % targetDuration;
+  const count = Math.floor(time / targetDuration) + (remainder ? 1 : 0);
```
brandonocasey, line 443: Don't append remainder after the loop, just do it in the for loop.

test/test-helpers.js
```diff
@@ -470,15 +475,26 @@ export const playlistWithDuration = function(time, conf) {
       segment.discontinuity = true;
     }
 
+    // add parts for the the last 3 segments in llhls playlists
+    if (conf && conf.llhls && (count - i) <= 3) {
+      segment.parts = [];
+      const partRemainder = segment.duration % result.partTargetDuration;
+      const partCount = Math.floor(segment.duration / result.targetDuration) + (partRemainder ? 1 : 0);
```
gesinger, line 482:
```suggestion
      const partCount = Math.floor(segment.duration / result.partTargetDuration) + (partRemainder ? 1 : 0);

```

test/loader-common.js
```diff
@@ -802,6 +802,151 @@ export const LoaderCommonFactory = ({
       });
     });
 
+    // only main/fmp4 segment loaders use parts/partIndex
+    if (usesAsyncAppends) {
+      QUnit.test('mediaIndex and partIndex are used', function(assert) {
+        const appendPart = (segmentIndex, partIndex) => {
```
gesinger, line 808: Would we be able to reuse this function for the other tests?

test/loader-common.js
```diff
@@ -802,6 +802,151 @@ export const LoaderCommonFactory = ({
...
+          loader.load();
+          loader.mediaIndex = 4;
+          return Promise.resolve();
+        }).then(() => appendPart(4, 0))
+          .then(() => appendPart(4, 1))
+          .then(() => appendPart(4, 2))
+          .then(() => {
+
+            // Update the playlist shifting the mediaSequence by 2 which will result
+            // in a decrement of the mediaIndex by 2 to 1
```
gesinger, line 931:
```suggestion
	    // Update the playlist shifting the mediaSequence by 4 which will result
            // in a decrement of the mediaIndex by 4 to 0
```

src/manifest.js
```diff
@@ -43,7 +44,35 @@ export const parseManifest = ({
   parser.push(manifestString);
   parser.end();
 
-  return parser.manifest;
+  const manifest = parser.manifest;
+
+  // remove llhls features from the parsed manifest
```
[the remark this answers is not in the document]
brandonocasey: This is so that we can turn off llhls by default, so that we can merge this and continue to play llhls back how we do now with just segments. Then we can remove this option when llhls is on by default.

src/playlist-loader.js
```diff
@@ -55,6 +72,23 @@ export const resolveSegmentUris = (segment, baseUri) => {
   if (segment.map && !segment.map.resolvedUri) {
     segment.map.resolvedUri = resolveUrl(baseUri, segment.map.uri);
   }
+  if (segment.parts && segment.parts.length) {
+    segment.parts.forEach((p) => {
+      if (p.resolvedUri) {
+        return;
+      }
+      p.resolvedUri = resolveUrl(baseUri, p.URI);
```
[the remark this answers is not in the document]
brandonocasey: I don't think we do, and I will be adding another pr for m3u8 parser to change to lowercase most of these.

scripts/index-demo-page.js
```diff
@@ -369,7 +377,8 @@
             vhs: {
               overrideNative: getInputValue(stateEls['override-native']),
               handlePartialData: getInputValue(stateEls.partial),
-              experimentalBufferBasedABR: getInputValue(stateEls['buffer-water'])
+              experimentalBufferBasedABR: getInputValue(stateEls['buffer-water']),
+              llhls: getInputValue(stateEls.llhls)
```
gkatsev, line 381: let's call it `experimentalLLHLS` for now.

src/segment-loader.js
```diff
@@ -1317,7 +1348,12 @@ export default class SegmentLoader extends videojs.EventTarget {
         startOfSegment = lastBufferedEnd;
       }
 
-      nextMediaIndex = currentMediaIndex + 1;
+      if (!segment || !segment.parts || !segment.parts.length || !segment.parts[nextPartIndex]) {
```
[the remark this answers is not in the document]
brandonocasey: It shouldn't I will remove the first check here.

src/segment-loader.js
```diff
@@ -975,6 +976,28 @@ export default class SegmentLoader extends videojs.EventTarget {
     // equal to the last appended mediaIndex
     if (this.mediaIndex !== null) {
       this.mediaIndex -= mediaSequenceDiff;
+
+      // this can happen if we are going to load the first segment, but get a playlist
+      // update during that. mediaIndex would go from 0 to -1 if mediaSequence in the
+      // new playlist was incremented by 1.
+      if (this.mediaIndex < 0) {
```
gesinger, line 983: I'm wondering if this will get us into trouble for short live playlists. Since we should start at the beginning of the playlist, the case may often happen that we end up with a mediaIndex of -1 while still processing the segment. In that case, the next segment we should load is segment 0 from the new playlist, which, if we do mediaIndex++, we'll be OK, unless we are at -2 or below. That may be a case for us to do a resync or seek. 

Do other areas of the code have trouble with a mediaIndex of -1?

src/playlist-loader.js
```diff
@@ -19,34 +19,73 @@ import {
...
+ * @return {Object} the merged segment
+ */
+export const updateSegment = (a, b) => {
+  if (!a) {
+    return b;
+  }
+
+  const result = mergeOptions(a, b);
+
+  // only the old segment has parts
```
gesinger, line 37:
```suggestion
  // if only the old segment has parts
```

src/playlist-loader.js
```diff
@@ -19,34 +19,73 @@ import {
...
+    }
+  }
+
+  return result;
+};
+
+/**
+ * Returns a new array of segments that is the result of merging
+ * properties from an older list of segments onto an updated
+ * list. No properties on the updated playlist will be overridden.
```
gesinger, line 58:
```suggestion
 * list. No properties on the updated playlist will be overwritten.
```

src/segment-loader.js
```diff
@@ -975,6 +976,28 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        // partIndex should remain the same for the same segment
+        // unless parts fell off of the playlist for this segment.
+        // In that case we need to reset partIndex and resync
+        if (this.partIndex && (!segment.parts || !segment.parts.length || !segment.parts[this.partIndex])) {
+          const mediaIndex = this.mediaIndex;
+
+          this.logger_(`currently processing part (index ${this.partIndex}) no longer exists.`);
+          this.resetLoader();
+
+          this.mediaIndex = mediaIndex;
```
gesinger, line 998: Is there a reason that we retain the media index after a reset?

brandonocasey: We want to throw away the partIndex and the data associated with it, but the mediaIndex is still valid. 
Example:
We are on mediaIndex 3 and partIndex 2, then a live refresh happens and we are on mediaIndex 2 and it no longer has parts.

src/segment-loader.js
brandonocasey: Don't we start at the end for live playlists?

brandonocasey: The reason that I added this code was: 
1. I have often seen issues with background tabs and live playlist refreshes happing way too slowly, so the segment we are on falls off of the playlist. 
2. having a negative media index doesn't really make sense, as we are dealing with zero based arrays. If the segment we are on falls off of the current playlist then the current media index would no longer be a valid media index for referring to the segment that we have appended.

I think ultimately the difference will be that we will never have a negative mediaIndex. So if we were two segments behind mediaIndex would not become -2 and leave us in a bad state. It would become negative and we would get the currentMediaIndex for time. 

It's possible that when mediaIndex and pendingSegment are both null, that we should probably try to jump closer to the live point in the stream. Right now we will just keep trying to request a segment at mediaIndex -1

gesinger: Looking again, I think we may be OK. My concern was mainly for the case of 3 segment live playlists. We start at the beginning of the first segment in the playlist, and it will probably be pretty often that we'll update our playlist during a segment request. If that happens, we clear out mediaIndex here (since the mediaIndex was 0 and we'd go to -1), so when we get to checkBuffer for the next segment,  we'd skip past mediaIndex++, but it looks like we'd still fetch at the end of buffer, so the guess should be OK: https://github.com/videojs/http-streaming/blob/82d83a70ea668a833ad1affef69a996d62383823/src/segment-loader.js#L1368-L1378

src/segment-loader.js
gesinger: 👍 would be worth a comment there

gesinger (approved):

PR #1056, opened 2021-01-28
Update README.md

troy-lamerton:
## Description

Fix code snippet
`vhs` did nothing. I tested with `hls` and its overriding on android correctly now.

Correct snippet taken from https://github.com/videojs/video.js/issues/6087#issuecomment-508797919

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,7 +342,7 @@ consistent experience.
 // via the constructor
 var player = videojs('playerId', {
   html5: {
-    vhs: {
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks, though, `vhs` is the correct option now, `hls` will continue to work for the time being for backwards compatibility. My guess is that you're still using an older version of VHS, perhaps with Video.js 7.9 or older?

PR #1059, opened 2021-02-03
feat: Allow xhr override globally

alex-barstow:
## Description
It is not currently possible to override or add any custom XHR logic at initialization, since the `xhr()` method is not exposed on the tech until the source is set ([see](https://github.com/videojs/http-streaming/blob/main/src/videojs-http-streaming.js#L1153)) and uses a reference to the [original `videojs.xhr()`](https://github.com/videojs/http-streaming/blob/main/src/xhr.js#L77).

Although it is unadvisable in most cases (as mentioned in the docs), it is still useful to have the ability to do it, and the existing `beforeRequest()` hook is insufficient in many cases.

## Specific Changes proposed
- Substitute a custom `videojs.Vhs.xhr` function for `videojsXHR`, if it is provided
- This PR does not add any documentation, as I wasn't sure if it is something we want to document

### DIFF
--- a/src/xhr.js
+++ b/src/xhr.js
@@ -74,7 +74,9 @@ const xhrFactory = function() {
       }
     }
 
-    const request = videojsXHR(options, function(error, response) {
+    const xhrMethod = videojs.Vhs.xhr.name !== 'XhrFunction' ? videojs.Vhs.xhr : videojsXHR;
+
+    const request = xhrMethod(options, function(error, response) {
       return callbackWrapper(request, error, response, callback);
     });
     const originalAbort = request.abort;
--- a/src/xhr.js
+++ b/src/xhr.js
@@ -75,6 +75,7 @@ const xhrFactory = function() {
     }
 
     // Use the standard videojs.xhr() method unless `videojs.Vhs.xhr` has been overriden
+    // TODO: switch back to videojs.Vhs.xhr.name === 'XhrFunction' when we drop IE11
     const xhrMethod = videojs.Vhs.xhr.original === true ? videojsXHR : videojs.Vhs.xhr;
 
     const request = xhrMethod(options, function(error, response) {
--- a/src/xhr.js
+++ b/src/xhr.js
@@ -74,7 +74,8 @@ const xhrFactory = function() {
       }
     }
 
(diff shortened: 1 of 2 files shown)


misteroneill (dismissed):

alex-barstow:
I realized that `Function.name` is unsupported in IE11, and may not be a reliable source of truth if the code is minified. Might be better to set a flag when overriding: `videojs.Vhs.xhr.override = true`. which @gkatsev had mentioned. Thoughts?

gkatsev:
Oh, didn't realize that `Function.name` wasn't available on IE11 because why would it be available there...
I think it's better to add a flag to the XhrFunction ourselves and if it's present we know it hasn't been overridden.

gkatsev:
Also, we can do the flag *just* for IE11 so that we can easily delete it once we drop IE11

src/xhr.js
```diff
@@ -74,7 +74,8 @@ const xhrFactory = function() {
       }
     }
 
-    const xhrMethod = videojs.Vhs.xhr.name !== 'XhrFunction' ? videojs.Vhs.xhr : videojsXHR;
+    // Use the standard videojs.xhr() method unless `videojs.Vhs.xhr` has been overriden
```
gkatsev, line 77: can we add a comment here that we can go back to use `Function.name` when we drop IE11?
```suggestion
    // Use the standard videojs.xhr() method unless `videojs.Vhs.xhr` has been overriden
    // TODO: switch back to videojs.Vhs.xhr.name === 'XhrFunction' when we drop IE11
```

brandonocasey (approved):

gkatsev:
One thing I tried to do while testing is override it but then call the old xhr method, which caused a whole host of issues. Though, I don't think we need to fix it because this should only ever be used in extreme circumstances and comes with no warranty.
Only thing I can think of is whether we want to maybe log a "I sure hope you know what you're doing" the first time we encounter an overriden xhr method.

The code was as follows:
```js
          const old = videojs.Vhs.xhr;

          videojs.Vhs.xhr = (...args) => {
            console.log('overriden!', args[0].uri);
            return old(...args);
          };
```

Issue #1041, opened 2021-01-11
BufferBasedABR + log.debug spams console


phloxic:
## Description

When enabling `experimentalBufferBasedABR` and `videojs.log.level('debug')` the console is spammed with debug messages.
https://videojs-http-streaming.netlify.app/?debug=true&autoplay=false&muted=false&minified=false&liveui=false&partial=false&url=https%3A%2F%2Fe3e2p5m2.ssl.hwcdn.net%2Fhls%2Fmain.m3u8&type=application%2Fx-mpegurl&keysystems=&buffer-water=true&override-native=true

## Sources
Is a certain source or a certain segment affected?

No.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.

1. Go to https://videojs-http-streaming.netlify.app/?debug=true&autoplay=false&muted=false&minified=false&liveui=false&partial=false&url=https%3A%2F%2Fe3e2p5m2.ssl.hwcdn.net%2Fhls%2Fmain.m3u8&type=application%2Fx-mpegurl&keysystems=&buffer-water=true&override-native=true
2. open console, tons of repeated message (especially at the outset, while buffering?) like the following:

```
VIDEOJS: DEBUG: VHS: MPC > allowing switch 2-/hls/360p/playlist.m3u8 -> 2-/hls/360p/playlist.m3u8 as current playlist is live video.core.js:84:49
VIDEOJS: DEBUG: VHS: PlaylistSelector > choosing {"id":"2-/hls/360p/playlist.m3u8","bandwidth":912767,"width":640,"height":360,"codecs":"avc1.4D0028,mp4a.40.2"} using resolutionPlusOneRep with options 
Object { bandwidth: 4194304, width: 640, height: 264, limitRenditionByPlayerDimensions: true }
video.core.js:84:49
VIDEOJS: DEBUG: VHS: MPC > allowing switch 2-/hls/360p/playlist.m3u8 -> 2-/hls/360p/playlist.m3u8 as current playlist is live video.core.js:84:49
VIDEOJS: DEBUG: VHS: PlaylistSelector > choosing {"id":"2-/hls/360p/playlist.m3u8","bandwidth":912767,"width":640,"height":360,"codecs":"avc1.4D0028,mp4a.40.2"} using resolutionPlusOneRep with options 
Object { bandwidth: 4194304, width: 640, height: 264, limitRenditionByPlayerDimensions: true }
video.core.js:84:49
```

### Expected

[...]

PR #1060, opened 2021-02-03
fix: do not request manifests until play when preload is none

brandonocasey:
## Description
We don't do a very good job respecting `preload=none` on the video element because even with preload=none we still download the manifests and potentially waste bandwidth with videos that never get played. We should prevent that when preload is set to `none`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,6 +32,9 @@
     input[type=url], select {
       min-width: 600px;
     }
+    #preload {
+      min-width: auto;
+    }
     h3 {
       margin-bottom: 5px;
     }
@@ -85,6 +88,13 @@
       <input id=override-native type="checkbox" checked>
       Override Native (reloads player)
     </label>
+    <label>
+      Preload (reloads player)
+      <select id=preload>
+        <option selected>auto</option>
+        <option>none</option>
+        <option>metadata</option>
+      </select>
   </div>
 
   <h3>Load a URL</h3>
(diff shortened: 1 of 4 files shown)


test/videojs-http-streaming.test.js
```diff
@@ -3718,7 +3718,7 @@ QUnit.test('switching playlists with an outstanding key request aborts request a
   );
 });
 
-QUnit.test('does not download segments if preload option set to none', function(assert) {
+QUnit.test('does not download anything until play if preload option set to none', function(assert) {
```
brandonocasey, line 3721: This test was failing, and didn't really make sense anymore. I modified for the new behavior.

blacktrash:
Great. I haven't tested it yet. But this probably also prevents a scenario with experimentalBufferBasedABR where the buffer check is constantly repeated at a short interval while the player is not 'loaded', see #1041

index.html
```diff
@@ -85,6 +88,13 @@ <h3>Options</h3>
       <input id=override-native type="checkbox" checked>
       Override Native (reloads player)
     </label>
+    <label>
```
gkatsev, line 91: awesome

src/master-playlist-controller.js
```diff
@@ -279,7 +279,16 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.logger_ = logger('MPC');
 
     this.triggeredFmp4Usage = false;
-    this.masterPlaylistLoader_.load();
+    if (this.tech_.preload() === 'none') {
```
gkatsev, line 282: glad it's surprisingly simple.

gkatsev (approved):
Works nicely!

PR #1065, opened 2021-02-09
fix: live dash segment changes should be considered a playlist update

brandonocasey:
## Description
Currently dash content goes off of the `number` attribute for segments to determine the mediaSequence. This is problematic as the first segment is almost always listed as `1`. So in essence when we request the playlist again the media sequence will be the same, and likely the amount of segments will be the same, so we will determine that this playlist update didn't update anything at all.

You can currently see this issue with the `Unified Streaming Live DASH` source. Eventually we will get into a state where the playlist isn't seen to be refreshing, even though the underlying segment uri's are changing every request.

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -62,8 +62,8 @@ export const resolveSegmentUris = (segment, baseUri) => {
 // and this playlist hasn't become the end of the playlist
 export const isPlaylistUnchanged = (a, b) => a === b ||
   (a.segments && b.segments && a.segments.length === b.segments.length &&
-  a.endList === b.endList &&
-    a.mediaSequence === b.mediaSequence);
+   a.endList === b.endList &&
+   a.mediaSequence === b.mediaSequence);
 
 /**
   * Returns a new master playlist that is the result of merging an
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -57,6 +57,14 @@ export const resolveSegmentUris = (segment, baseUri) => {
   }
 };
 
+// consider the playlist unchanged if the playlist object is the same or
+// the number of segments is equal, the media sequence number is unchanged,
+// and this playlist hasn't become the end of the playlist
+export const isPlaylistUnchanged = (a, b) => a === b ||
+  (a.segments && b.segments && a.segments.length === b.segments.length &&
+  a.endList === b.endList &&
+    a.mediaSequence === b.mediaSequence);
+
 /**
   * Returns a new master playlist that is the result of merging an
   * updated media playlist into the original version. If the
@@ -69,7 +77,7 @@ export const resolveSegmentUris = (segment, baseUri) => {
   * master playlist with the updated media playlist merged in, or
   * null if the merge produced no change.
   */
-export const updateMaster = (master, media, dash = false) => {
+export const updateMaster = (master, media, unchangedCheck = isPlaylistUnchanged) => {
   const result = mergeOptions(master, {});
   const playlist = result.playlists[media.id];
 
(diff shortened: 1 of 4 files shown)


src/playlist-loader.js
```diff
@@ -79,11 +79,59 @@ export const updateMaster = (master, media) => {
...
+  // for dash mediaSequence and segment lengths will often match as
+  // mediaSequence is almost always 1 and the number of segments generated for a
+  // given time is often the same. So we need to make sure that the underlying segments are
+  // different.
+  if (dash && unchanged) {
+    const oldSidx = playlist.sidx;
+    const newSidx = media.sidx;
+
+    // if sidx changed then the playlists are different.
+    if (oldSidx && newSidx && (oldSidx.offset !== newSidx.offset || oldSidx.length !== newSidx.length)) {
```
brandonocasey, line 97: Now dash playlists can also be listed as changing in cases where:
1. sidx does not match
2. playlist segment uris don't match up
3. playlist segment byterange uris don't match up.

test/manifests/dashByterange.mpd
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
brandonocasey, line 1: We didn't have a dash byterange source

src/playlist-loader.js
```diff
@@ -69,7 +69,7 @@ export const resolveSegmentUris = (segment, baseUri) => {
   * master playlist with the updated media playlist merged in, or
   * null if the merge produced no change.
   */
-export const updateMaster = (master, media) => {
+export const updateMaster = (master, media, dash = false) => {
```
gesinger, line 72: It might help separate things a bit if we move the DASH specific check into dash-playlist-loader. Maybe have a method updateAndCheckForChanges that wraps both updatePlaylist and didSegmentsChange or something like that.

Might also make it easier to test the function separately.

src/playlist-loader.js
```diff
@@ -79,11 +79,59 @@ export const updateMaster = (master, media) => {
...
+        const oldSegment = playlist.segments[i];
+        const newSegment = media.segments[i];
+
+        // if uris are different between segments there was a change
+        if (oldSegment.uri !== newSegment.uri) {
+          unchanged = false;
+          continue;
+        }
+
+        // neither segment has a byternage, there will be no byterange change.
```
gesinger, line 112:
```suggestion
        // neither segment has a byterange, there will be no byterange change.
```

gkatsev:
Testing with the US Live DASH stream, it's playing, though, after a bit it will bip bop and then buffer and then bip bop and then buffer, where-as before it would just eventually stall and buffer forever.

src/playlist-loader.js
brandonocasey: Added a separate function `isPlaylistUnchanged` that is passed in as `unchangedCheck` to this function. Dash passes in it's own version that does all the dash specific things and also calls `isPlaylistUnchanged` at the top.

gkatsev:
Here's a quick screen recoding of how it looks for me: 

https://user-images.githubusercontent.com/535884/107569543-6bb25000-6bb6-11eb-893d-fa5c2c7e3435.mov

gkatsev:
It does seem to just stop working properly some times too.

gkatsev:
Seeking seems to fix it getting stuck, but then it gets stuck again. Really weird.

brandonocasey:
Yeah there might be another issue here for that source. I think the behavior is better than it was though.

gkatsev:
Do we want to get this merged in and then fix other live related stuff in subsequent PRs?

brandonocasey:
I think we do, this fix is ready to go and fixes some streams right now

brandonocasey:
I don't have a dash byterange source, I was going off of an example I found where byterange was being used.

gesinger (approved):

ziddey:
Wouldn't it make sense to just fake the media sequence? We take the uri of the first segment in the new playlist, and find the matching segment in the old playlist. Use its index and the old playlist's media sequence to derive the new media sequence. If the segment isn't found, increase sequence by the number of segments to simulate falling behind the live window. Then `isPlaylistUnchanged` / `updateSegments` function properly.

As a test, I added to the start of `dashPlaylistUnchanged()`:

```
  const seqDelta = a.segments.findIndex(s => s.uri === b.segments[0].uri)
  b.mediaSequence = a.mediaSequence + (seqDelta === -1 ? a.segments.length : seqDelta)
```

Now the US Live DASH stream and other $Time$-based live manifests work.

We'd need to determine the conditions for applying this delta. Perhaps at least `a.segments && b.segments && b.mediaSequence === 1`? Or if all segments have `number === 1`? Basically to not interfere if the manifest does have a `startNumber`.

PR #1070, opened 2021-02-16
fix: allow buffer removes when there's no current media info in loader

gesinger:
Previously, if there wasn't any current media info in the loader, then a
remove of buffered contents would not be processed. This would
intuitively make sense, as if there wasn't current media info, then in
theory there shouldn't be any buffered contents. However, it is possible
for there to be buffered contents without current media info in the event
of a non destructive rendition switch. This was noticed when switching
renditions, expecting a fast quality change to clear the contents and
continue playback at a new rendition. Sometimes the remove wouldn't
happen, so the content would continue to play back the old buffered
content from a different rendition. This fix instead allows the removes
so long as there is starting media info (which is available after first
segment download).

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -750,6 +750,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
    */
   fastQualityChange_(media = this.selectPlaylist()) {
     if (media === this.masterPlaylistLoader_.media()) {
+      this.logger_('skipping fastQualityChange because new media is same as old');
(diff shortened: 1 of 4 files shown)


gkatsev (approved):
This seems too simple.

brandonocasey (approved):
Seems good to me, although it does make me wonder if we should even worry about startingMediaInfo, and just queue the remove no matter what.

PR #1071, opened 2021-02-16
chore: mirror player.src on the demo page demo page

brandonocasey:
## Description
Uses sourceset to mirror the source that was set via `player.src` in the ui so that it will be saved on page reload!

### DIFF
--- a/index.html
+++ b/index.html
@@ -90,7 +90,7 @@
     </label>
     <label>
       <input id=mirror-source type="checkbox" checked>
-      Mirror sources from player.src (reloads player)
+      Mirror sources from player.src (reloads player, uses EXPERIMENTAL sourceset option)
     </label>
     <label>
       Preload (reloads player)
--- a/index.html
+++ b/index.html
@@ -89,6 +89,10 @@
       Override Native (reloads player)
     </label>
     <label>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -88,6 +88,10 @@ <h3>Options</h3>
       <input id=override-native type="checkbox" checked>
       Override Native (reloads player)
     </label>
+    <label>
+      <input id=mirror-source type="checkbox" checked>
+      Mirror sources from player.src (reloads player)
```
gkatsev, line 93: Should we just call this `sourceset`? This the target of these pages are us, we can be as technical as we want.

gkatsev (approved):
Naming aside, LGTM

index.html
brandonocasey: At first I wanted to, but that isn't really the feature that we care about here. We care about mirroring sources from `player.src`, `sourceset` just happens to make that very easy to do.

gkatsev: Maybe adding `uses sourceset` in the parens too.
My thought process here is that we always want mirroring but that it's possible that sourceset may cause issues so we'd want to disable it.

brandonocasey: I will do that and merge

gkatsev (approved):

PR #1073, opened 2021-02-17
fix: sometimes subtitlesTrack_.cues is null

ysm0622:
fix: sometimes subtitlesTrack_.cues is null

## Description
image
image

sometimes subtitlesTrack_.cues is null

## Specific Changes proposed

I added null check code.

## Requirements Checklist
- [v] Feature implemented / Bug fixed


brandonocasey (approved):
I don't see any harm in adding another null check

gesinger (approved):

PR #1074, opened 2021-02-17
feat: deprecate smoothQualityChange

gesinger:
## Description
smoothQualityChange was used during certain user interactions: selecting a new rendition via the representations API (when set to true), and when going fullscreen. The method served to keep the content in the buffer but overlay new content on top, allowing playback to continue at the current rendition without the interruption of a loading screen. However, in these user interactions, it is generally expected behavior that the quality change should be immediate, and a short loading period is not an unexpected interruption.

As such, smoothQualityChange is deprecated as an option and will call fastQualityChange until it is removed in the next major version.

## Requirements Checklist
  - [X] Docs/guides updated


src/master-playlist-controller.js
```diff
@@ -728,16 +728,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
   smoothQualityChange_(media = this.selectPlaylist()) {
-    if (media === this.masterPlaylistLoader_.media()) {
-      return;
-    }
-
-    this.masterPlaylistLoader_.media(media);
-
-    this.mainSegmentLoader_.resetLoader();
-    // don't need to reset audio as it is reset when media changes
+    this.fastQualityChange_(media);
```
brandonocasey, line 734: Should we log something the first time this is called?

gesinger: I figured since we warn on use of the option and it's an internal method, we're probably safe with just the option use log.

brandonocasey (dismissed):

gkatsev (approved):

gkatsev:
@benesva4 actual smooth quality change isn't deprecated. We still change smoothly during normal playback. We've only deprecated the feature for when a user changes quality levels because it has caused some bugs for us but also because users almost always want fast switching when they select a different rendition.

PR #1078, opened 2021-02-22
feat: use serverControl and preloadSegment llhls features behind a flag

brandonocasey:
## Description
llhls features are already behind a flag, from the previous pull request #1055.

This implements usage of:
* `preloadSegment` when it has parts. Think of this as the segment that is about to appear in the manifest, but we don't have a uri for yet.
* `severControl` but only `PART-HOLD-BACK` and `HOLD-BACK` which specify how far behind a live playlist we should start
* `partTargetDuration` in any way that we current use targetDuration, such as how much of a delay we should have on a playlist refresh

### DIFF
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -26,7 +26,8 @@ export const parseManifest = ({
   oninfo,
   manifestString,
   customTagParsers = [],
-  customTagMappers = []
+  customTagMappers = [],
+  llhls
 }) => {
   const parser = new M3u8Parser();
 
@@ -43,7 +44,35 @@ export const parseManifest = ({
   parser.push(manifestString);
   parser.end();
 
-  return parser.manifest;
+  const manifest = parser.manifest;
+
+  // remove llhls features from the parsed manifest
+  if (!llhls) {
+    [
+      'preloadSegment',
+      'skip',
+      'serverControl',
+      'renditionReports',
+      'partInf',
+      'partTargetDuration'
+    ].forEach(function(k) {
+      if (manifest.hasOwnProperty(k)) {
+        delete manifest[k];
+      }
+    });
+
+    if (manifest.segments) {
+      manifest.segments.forEach(function(segment) {
+        ['parts', 'preloadHints'].forEach(function(k) {
+          if (segment.hasOwnProperty(k)) {
+            delete segment[k];
+          }
+        });
+      });
+    }
+  }
+
+  return manifest;
 };
 
 /**
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -12,12 +12,18 @@ export const createPlaylistID = (index, uri) => {
 /**
  * Parses a given m3u8 playlist
  *
+ * @param {Function} [onwarn]
+ *        a function to call when the parser triggers a warning event.
+ * @param {Function} [oninfo]
+ *        a function to call when the parser triggers an info event.
  * @param {string} manifestString
  *        The downloaded manifest string
  * @param {Object[]} [customTagParsers]
  *        An array of custom tag parsers for the m3u8-parser instance
  * @param {Object[]} [customTagMappers]
- *         An array of custom tag mappers for the m3u8-parser instance
+ *        An array of custom tag mappers for the m3u8-parser instance
+ * @param {boolean} [llhls=false]
+ *        Wether to keep ll-hls features in the manifest after parsing.
  * @return {Object}
  *         The manifest object
  */
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -22,7 +22,7 @@ export const createPlaylistID = (index, uri) => {
  *        An array of custom tag parsers for the m3u8-parser instance
  * @param {Object[]} [customTagMappers]
  *        An array of custom tag mappers for the m3u8-parser instance
- * @param {boolean} [llhls=false]
+ * @param {boolean} [experimentalLLHLS=false]
  *        Wether to keep ll-hls features in the manifest after parsing.
  * @return {Object}
  *         The manifest object
@@ -33,7 +33,7 @@ export const parseManifest = ({
   manifestString,
   customTagParsers = [],
   customTagMappers = [],
-  llhls
+  experimentalLLHLS
 }) => {
   const parser = new M3u8Parser();
 
@@ -53,7 +53,8 @@ export const parseManifest = ({
   const manifest = parser.manifest;
 
   // remove llhls features from the parsed manifest
-  if (!llhls) {
+  // if we don't want llhls support.
+  if (!experimentalLLHLS) {
     [
       'preloadSegment',
       'skip',
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -23,7 +23,7 @@ export const createPlaylistID = (index, uri) => {
  * @param {Object[]} [customTagMappers]
  *        An array of custom tag mappers for the m3u8-parser instance
  * @param {boolean} [experimentalLLHLS=false]
- *        Wether to keep ll-hls features in the manifest after parsing.
+ *        Whether to keep ll-hls features in the manifest after parsing.
  * @return {Object}
  *         The manifest object
  */
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -78,14 +78,18 @@ export const parseManifest = ({
       });
     }
   }
-  if (manifest.segments && manifest.segments.length && !manifest.targetDuration) {
-    const lastSegment = manifest.segments[manifest.segments.length - 1];
-    const lastDuration = lastSegment && Math.round(lastSegment) || 10;
+  if (!manifest.targetDuration) {
+    let targetDuration = 10;
+
+    if (manifest.segments && manifest.segments.length) {
+      targetDuration = manifest
+        .segments.reduce((acc, s) => Math.max(acc, s.duration), 0);
+    }
 
     if (onwarn) {
-      onwarn(`manifest has no targetDuration defaulting to ${lastDuration}`);
+      onwarn(`manifest has no targetDuration defaulting to ${targetDuration}`);
     }
-    manifest.targetDuration = lastDuration;
+    manifest.targetDuration = targetDuration;
   }
 
   return manifest;
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -78,6 +78,13 @@ export const parseManifest = ({
       });
     }
   }
+  if (manifest.segments && manifest.segments.length && !manifest.targetDuration) {
+    const lastSegment = manifest.segments[manifest.segments.length - 1];
+    const lastDuration = lastSegment && Math.round(lastSegment) || 10;
+
+    onwarn(`manifest has no targetDuration defaulting to ${lastDuration}`);
+    manifest.targetDuration = lastDuration;
+  }
 
   return manifest;
 };
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -82,7 +82,9 @@ export const parseManifest = ({
     const lastSegment = manifest.segments[manifest.segments.length - 1];
     const lastDuration = lastSegment && Math.round(lastSegment) || 10;
 
-    onwarn(`manifest has no targetDuration defaulting to ${lastDuration}`);
+    if (onwarn) {
+      onwarn(`manifest has no targetDuration defaulting to ${lastDuration}`);
+    }
     manifest.targetDuration = lastDuration;
   }
 
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -301,8 +301,20 @@ export class MasterPlaylistController extends videojs.EventTarget {
     const nextPlaylist = this.selectPlaylist();
 
     if (this.shouldSwitchToMedia_(nextPlaylist)) {
-      this.masterPlaylistLoader_.media(nextPlaylist);
+      this.switchMedia_(nextPlaylist, 'abr');
     }
+  }
+
+  switchMedia_(playlist, cause, delay) {
+    const oldMedia = this.media();
+    const oldId = oldMedia && (oldMedia.id || oldMedia.uri);
+    const newId = playlist.id || playlist.uri;
+
+    if (oldId && oldId !== newId) {
+      this.logger_(`switch media ${oldId} -> ${newId} from ${cause}`);
+      this.tech_.trigger({type: 'usage', name: `vhs-rendition-change-${cause}`});
+    }
+    this.masterPlaylistLoader_.media(playlist, delay);
   }
 
   /**
@@ -416,7 +428,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
         this.initialMedia_ = selectedMedia;
 
-        this.masterPlaylistLoader_.media(this.initialMedia_);
+        this.switchMedia_(this.initialMedia_, 'initial');
 
         // Under the standard case where a source URL is provided, loadedplaylist will
         // fire again since the playlist will be requested. In the case of vhs-json
@@ -617,7 +629,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
         const nextPlaylist = this.selectPlaylist();
 
         if (this.shouldSwitchToMedia_(nextPlaylist)) {
-          this.masterPlaylistLoader_.media(nextPlaylist);
+          this.switchMedia_(nextPlaylist, 'bandwidthupdate');
         }
 
         this.tech_.trigger('bandwidthupdate');
@@ -743,7 +755,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return;
     }
 
-    this.masterPlaylistLoader_.media(media);
+    this.switchMedia_(media, 'smooth-quality');
 
     this.mainSegmentLoader_.resetLoader();
     // don't need to reset audio as it is reset when media changes
@@ -763,7 +775,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return;
     }
 
-    this.masterPlaylistLoader_.media(media);
+    this.switchMedia_(media, 'fast-quality');
 
     // Delete all buffered data to allow an immediate quality switch, then seek to give
     // the browser a kick to remove any cached frames from the previous rendtion (.04 seconds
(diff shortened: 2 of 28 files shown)


src/master-playlist-controller.js
```diff
@@ -1287,8 +1287,21 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return;
     }
 
-    const suggestedPresentationDelay = this.masterPlaylistLoader_.master.suggestedPresentationDelay;
-    const mainSeekable = Vhs.Playlist.seekable(media, expired, suggestedPresentationDelay);
+    let delay;
+
+    if (this.masterPlaylistLoader_.master.hasOwnProperty('suggestedPresentationDelay')) {
+      delay = this.masterPlaylistLoader_.master.suggestedPresentationDelay;
+    } else if (media.serverControl && media.serverControl['PART-HOLD-BACK'] && media.partTargetDuration) {
```
brandonocasey, line 1294: ll-hls has it's own version of when a playlist should be started called `HOLD-BACK` and `PART-HOLD-BACK`

src/master-playlist-controller.js
```diff
@@ -1287,8 +1287,21 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      delay = media.serverControl['PART-HOLD-BACK'] * media.partTargetDuration;
+    } else if (media.serverControl && media.serverControl['HOLD-BACK'] && media.targetDuration) {
+      delay = media.serverControl['HOLD-BACK'] * media.targetDuration;
+    } else if (media.partTargetDuration) {
+      delay = media.partTargetDuration * 3;
+    } else if (media.targetDuration) {
+      delay = media.targetDuration * 3;
```
brandonocasey, line 1301: These are the default values for `HOLD-BACK` as per the spec, so even if a playlist doesn't have `HOLD-BACK` we can still set the default `HOLD-BACK`

src/playlist-loader.js
```diff
@@ -166,16 +175,16 @@ export const updateMaster = (master, media, unchangedCheck = isPlaylistUnchanged
...
   const lastSegment = media.segments[media.segments.length - 1];
-  let delay;
-
-  if (update && lastSegment && lastSegment.duration) {
-    delay = lastSegment.duration * 1000;
-  } else {
-    // if the playlist is unchanged since the last reload or last segment duration
-    // cannot be determined, try again after half the target duration
-    delay = (media.targetDuration || 10) * 500;
+  const lastPart = lastSegment && lastSegment.parts && lastSegment.parts[lastSegment.parts - 1];
```
brandonocasey, line 178: use the last part duration if it exists or the last segment duration

src/playlist-loader.js
```diff
@@ -166,16 +175,16 @@ export const updateMaster = (master, media, unchangedCheck = isPlaylistUnchanged
...
+  const lastDuration = lastPart && lastPart.DURATION || lastSegment && lastSegment.duration;
+
+  if (update && lastDuration) {
+    return lastDuration * 1000;
   }
-  return delay;
+
+  // if the playlist is unchanged since the last reload or last segment duration
+  // cannot be determined, try again after half the target duration
+  return (media.partTargetDuration || media.targetDuration || 10) * 500;
```
brandonocasey, line 187: use partTargetDuration or targetDuration

src/playlist-loader.js
```diff
@@ -304,7 +313,7 @@ export default class PlaylistLoader extends EventTarget {
     // merge this playlist into the master
     const update = updateMaster(this.master, playlist);
 
-    this.targetDuration = playlist.targetDuration;
+    this.targetDuration = playlist.partTargetDuration || playlist.targetDuration;
```
brandonocasey, line 316: We use a playlists targetDuration in a few places outside of these files. In the case where we have partTargetDuration that is always what we will use as out "true" targetDuration, I understand that this is ambiguous though and I think we might want to rename `this.targetDuration` to something else, but that might be a breaking change. Should we deprecate this name?

src/playlist.js
```diff
@@ -347,44 +348,58 @@ export const getMediaInfoForTime = function(
   startIndex,
   startTime
 ) {
-  let i;
-  let segment;
-  const numSegments = playlist.segments.length;
+
+  const partSegments = playlist.segments.reduce((acc, segment, si) => {
```
brandonocasey, line 352: We need this function to know about parts now. So we break put all of our segments and or parts into one list.

src/segment-loader.js
```diff
@@ -94,20 +94,28 @@ const segmentInfoString = (segmentInfo) => {
...
       segments = []
     },
     mediaIndex: index,
+    partIndex,
     timeline
   } = segmentInfo;
 
+  const name = segmentInfo.segment.uri ? 'segment' : 'pre-segment';
+
   return [
```
brandonocasey, line 112: Better logging in general, but especially needed now that we have parts.

src/segment-loader.js
```diff
@@ -1358,28 +1367,22 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-
-      nextMediaIndex = mediaSourceInfo.mediaIndex;
-      startOfSegment = mediaSourceInfo.startTime;
     } else {
-      // Find the segment containing currentTime
+      // Find the segment containing the end of the buffer or current time.
       const mediaSourceInfo = Playlist.getMediaInfoForTime(
         playlist,
-        currentTime,
+        this.fetchAtBuffer_ ? lastBufferedEnd : currentTime,
```
brandonocasey, line 1374: It was silly to have 10 lines duplicated for a single argument change.

src/segment-loader.js
```diff
@@ -1358,28 +1367,22 @@ export default class SegmentLoader extends videojs.EventTarget {
...
         syncPoint.segmentIndex,
         syncPoint.time
       );
 
       nextMediaIndex = mediaSourceInfo.mediaIndex;
       startOfSegment = mediaSourceInfo.startTime;
+      nextPartIndex = mediaSourceInfo.partIndex;
+    }
+
+    if (typeof nextPartIndex !== 'number' && playlist.segments[nextMediaIndex] && playlist.segments[nextMediaIndex].parts) {
```
brandonocasey, line 1384: if we don't have a nextPartIndex, check if we should.

src/segment-loader.js
```diff
@@ -1352,41 +1360,36 @@ export default class SegmentLoader extends videojs.EventTarget {
     } else if (currentMediaIndex !== null) {
     // Under normal playback conditions fetching is a simple walk forward
       const segment = playlist.segments[currentMediaIndex];
+      const partIndex = typeof currentPartIndex === 'number' ? currentPartIndex : -1;
```
gesinger, line 1363: Is it possible for it to be a valid case where the `currentMediaIndex` is the segment we're supposed to be fetching will the `partIndex` is 0? Are the hinted segments available here to see?

src/segment-loader.js
```diff
@@ -1406,11 +1409,8 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     this.logger_(`checkBuffer_ returning ${segmentInfo.uri}`, {
       segmentInfo,
-      playlist,
       currentMediaIndex,
-      nextMediaIndex,
-      startOfSegment,
-      isSyncRequest
```
gesinger, line 1413: These can actually be valuable to contain, in particular `startOfSegment`, which has been a source of various bugs we've encountered.

test/playlist.test.js
```diff
@@ -574,7 +574,7 @@ QUnit.test('safeLiveIndex accounts for liveEdgePadding in simple case', function
   );
 
   assert.equal(
-    Playlist.safeLiveIndex(playlist, 0), 6,
+    Playlist.safeLiveIndex(playlist, 0), 5,
     'returns 6 when liveEdgePadding is 0 and duration is 6'
```
gesinger, line 578: Was there a reason for this change? If so, we may have to update the message.

test/playlist.test.js
```diff
@@ -650,11 +650,100 @@ QUnit.test('safeLiveIndex accounts for liveEdgePadding in non-simple case', func
   );
 
   assert.equal(
-    Playlist.safeLiveIndex(playlist, 0), 6,
+    Playlist.safeLiveIndex(playlist, 0), 5,
     'returns 6 when liveEdgePadding is 0'
```
gesinger, line 654: Same re: above

test/playlist.test.js
```diff
@@ -650,11 +650,100 @@ QUnit.test('safeLiveIndex accounts for liveEdgePadding in non-simple case', func
...
+    'uses part target duration * 3'
+  );
+
+  media.partTargetDuration = null;
+  assert.equal(
+    Playlist.getHoldBack(master, media),
+    2,
+    'uses last three part durations'
+  );
+
```
gesinger, line 716: May want another test here with fewer than 3 parts available.

src/playlist-loader.js
```diff
@@ -147,6 +147,15 @@ export const updateMaster = (master, media, unchangedCheck = isPlaylistUnchanged
 
   const mergedPlaylist = mergeOptions(playlist, media);
 
+  media.segments = media.segments || [];
```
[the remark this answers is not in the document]
brandonocasey: Was probably added for tests that don't pass in a segments array, let me check that.

src/segment-loader.js
brandonocasey: currentPartIndex will come in as null, in which case we need to check if the current media index has parts at index 0. it could also come in as 0 and in that case we know we appended part 0 and need to look for part 1, or increment media index if we don't have it.

src/segment-loader.js
brandonocasey: These are all on segmentInfo, which is my reasoning behind removing them. But we could keep them as they keep things a bit more relevant to the checkBuffer log.

test/playlist.test.js
brandonocasey: so this was actually a bug. We return `playlists.segments.length` which isn't a valid index in the array, it was 1 too high.

test/playlist.test.js
brandonocasey: in a real scenario the last 3 segments will always have >= 2 parts. as having only one part is the same as not using parts at all. I will update the test for that.

src/segment-loader.js
gesinger: I don't think these get logged except for here. Maybe worth adding to `segmentInfoString`?

src/playlist.js
```diff
@@ -30,34 +30,39 @@ const getPartSegments = (playlist) => (playlist.segments || []).reduce((acc, seg
...
+    typeof partsAndSegments[partsAndSegments.length - 1].partIndex === 'number';
 
+  // get the last three part/segment durations
   let lastThreeDurations = 0;
 
-  if (partSegments.length >= 3) {
-    for (let i = 0; i < 3; i++) {
-      lastThreeDurations += partSegments[partSegments.length - 1 - i].duration;
+  if (partsAndSegments.length >= 3) {
+    for (let i = partsAndSegments.length - 1; i > partsAndSegments.length - 4; i--) {
```
gesinger: Worth adding a reference to the spec about using three.

test/playlist.test.js
```diff
@@ -681,66 +685,80 @@ QUnit.test('getHoldBack works as expected', function(assert) {
...
 
+  media.segments = media.segments.map((s) => {
+    s.duration = null;
+    return s;
+  });
+
+  assert.equal(
+    Playlist.liveEdgeDelay(master, media),
+    0,
+    'no segment durations live delay can be calculated'
```
gesinger, line 753:
```suggestion
    'no segment durations, live delay can\'t be calculated'
```

test/playlist.test.js
```diff
@@ -681,66 +685,80 @@ QUnit.test('getHoldBack works as expected', function(assert) {
...
+  );
+
   media.segments.length = 0;
 
   assert.equal(
-    Playlist.getHoldBack(master, media),
+    Playlist.liveEdgeDelay(master, media),
     0,
-    'no possible holdback can be calculated'
+    'no segments live delay can be calculated'
```
gesinger, line 761:
```suggestion
    'no segments, live delay can\'t be calculated'
```

src/playlist.js
```diff
@@ -72,10 +77,13 @@ export const getHoldBack = (master, media) => {
...
+    const lastPartSegment = partsAndSegments[partsAndSegments.length - 1];
     const lastPartDuration = lastPartSegment && lastPartSegment.duration || media.targetDuration;
 
     return lastPartDuration + media.targetDuration * 2;
+  // we shouldn't ever end up using lastThreeDurations as targetDuration
+  // is usually required, but if we somehow get here, with a missing
+  // targetDuration we should handle it
   } else if (lastThreeDurations) {
     return lastThreeDurations;
   }
```
gesinger: Since we shouldn't end up here, it's probably worth pulling all of that logic down into an `else` here and return either 0 or whatever we calculate as the last three. Might also be worth warning (if we don't warn elsewhere for missing target duration), or logging anyway to help us determinelogic in the event of a missing target duration.

src/segment-loader.js
brandonocasey: I have the next pull request in the series (where I refactor our segment selection logic) that adds selection to `segmentInfoString`. I think we should wait until that pull request. I will revert this logging change here for now.

src/playlist.js
brandonocasey: We use lastThreeDurations in the case where we have parts, but no partTargetDuration so that we get a more accurate hold back for ll-hls. We warn about target duration via the manifest parser warnings that now go to the debug log.

src/playlist.js
```diff
@@ -45,20 +63,10 @@ export const liveEdgeDelay = (master, media) => {
...
-      if (!partsAndSegments[i].duration) {
-        lastThreeDurations = 0;
-        break;
-      }
-      lastThreeDurations += partsAndSegments[i].duration;
-    }
-  }
+  // by default we use the last three durations of segments if
+  // part target duration or target duration isn't found.
+  // see: https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-08#section-4.4.3.8
```
gesinger, line 68: I may be missing something, but I don't think that section of the spec, if referring to `HOLD-BACK`, matches the comment here, since that is saying to use part target duration or target duration when the parameter is not available, and doesn't refer to the use of segment durations. Do we just have this in case the required target durations are missing?

brandonocasey: Yes we use three segment durations in case target durations are not available for whatever reason.

gesinger: I'm wondering if it's worth it to add the logic for this here if those are required attributes. It adds logic to the code that we shouldn't really encounter, and if we were to, then it's something that should be resolved by the creator of the stream, since target duration is always required (https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-08#section-4.4.3.1) and part target duration is required when parts are used (https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-08#section-4.4.3.7).

I think if we're trying to account for missing target durations we should do that at a higher level, when we first parse the manifest. It would help to simplify the code.

brandonocasey: ok I will remove the last three durations logic.

src/playlist.js
```diff
@@ -9,6 +9,68 @@ import {TIME_FUDGE_FACTOR} from './ranges.js';
...
+ *
+ * @param {Playlist} playlist the playlist to get the list for.
+ *
+ * @return {Array} The part/segment list.
+ */
+const getPartsAndSegments = (playlist) => (playlist.segments || []).reduce((acc, segment, si) => {
+  if (segment.parts) {
+    segment.parts.forEach(function(part, pi) {
+      acc.push({duration: part.duration, segmentIndex: si, partIndex: pi});
+
```
gkatsev, line 24:
```suggestion
```

src/playlist.js
gkatsev: What about just default to 10s target duration in case somehow it isn't set or whatever? While I agree that bad streams are bad, we should still try to play as much as possible and having a default, if bad value, is worth it. Plus, it should make things simpler than trying to figure out segment duration and using that for target duation.

gkatsev (commented):
While I don't see any major issues, I think there are a couple of small changes that should be made in playlist.js based on Garrett's comments. Would make that function slightly simpler.

gkatsev:
Also, this PR is now out of date with main.

src/manifest.js
```diff
@@ -78,6 +78,15 @@ export const parseManifest = ({
       });
     }
   }
+  if (manifest.segments && manifest.segments.length && !manifest.targetDuration) {
+    const lastSegment = manifest.segments[manifest.segments.length - 1];
+    const lastDuration = lastSegment && Math.round(lastSegment) || 10;
+
+    if (onwarn) {
+      onwarn(`manifest has no targetDuration defaulting to ${lastDuration}`);
```
gesinger, line 86: I think we may be better off using the largest segment duration found (and 10 if no segments included), since the TARGETDURATION is supposed to be the highest duration that will be encountered in a stream (and we will run into issues if the target duration is shorter than some segment durations).

Would also be worth having a test for it.

gesinger (dismissed):

test/manifest.test.js
```diff
@@ -4,389 +4,425 @@ import {
...
-      uri: 'uri-1'
-    }]
-  };
-
-  setupMediaPlaylists(master);
-
-  assert.equal(master.playlists[0].resolvedUri, urlTo('uri-0'), 'resolves URI');
-  assert.equal(master.playlists[1].resolvedUri, urlTo('uri-1'), 'resolves URI');
-});
+  QUnit.test('sets target duration to largest 10 without segments', function(assert) {
```
gesinger, line 36:
```suggestion
  QUnit.test('sets target duration to 10 without segments', function(assert) {
```

src/playlist.js
```diff
@@ -59,11 +59,7 @@ export const liveEdgeDelay = (master, media) => {
   } else if (media.serverControl && media.serverControl.holdBack) {
     return media.serverControl.holdBack;
   } else if (media.targetDuration) {
-    // TODO: this should probably be targetDuration * 3
-    // but we use this for backwards compatability.
-    const lastPartDuration = lastSegment && lastSegment.duration || media.targetDuration;
-
-    return lastPartDuration + media.targetDuration * 2;
+    return media.targetDuration * 3;
```
brandonocasey, line 62: This was changed in main via #1105

gesinger (approved):

PR #1079, opened 2021-02-22
feat: Use ll-hls query directives and support skipping segments

brandonocasey:
## Description
Implements "query directives" when supported by the server so that we can get delta playlist updates to offset the cost of requesting a manifests so frequently.


src/playlist-loader.js
```diff
@@ -632,11 +715,9 @@ export default class PlaylistLoader extends EventTarget {
 
       this.src = resolveManifestRedirect(this.handleManifestRedirects, this.src, req);
 
-      const manifest = parseManifest({
+      const manifest = this.parseManifest_({
```
brandonocasey, line 718: move parseManifest to `this.parseManifest` as we weren't previously passing all the options in both calls.

src/playlist-loader.js
```diff
@@ -146,28 +177,36 @@ export const isPlaylistUnchanged = (a, b) => a === b ||
   * master playlist with the updated media playlist merged in, or
   * null if the merge produced no change.
   */
-export const updateMaster = (master, media, unchangedCheck = isPlaylistUnchanged) => {
+export const updateMaster = (master, newMedia, unchangedCheck = isPlaylistUnchanged) => {
```
brandonocasey, line 180: I continuously got confused by `media` and `playlist` here and realized that they are just terrible names. Most of the changes in this function, other than those noted,  are for this rename.

src/playlist-loader.js
```diff
@@ -146,28 +177,36 @@ export const isPlaylistUnchanged = (a, b) => a === b ||
...
     return null;
   }
 
-  if (unchangedCheck(playlist, media)) {
+  if (unchangedCheck(oldMedia, newMedia)) {
     return null;
   }
 
-  const mergedPlaylist = mergeOptions(playlist, media);
+  newMedia.segments = getAllSegments(newMedia);
```
brandonocasey, line 192: preserve `preloadSegment` from `newMedia` during the merge that will happen below.

src/playlist-loader.js
```diff
@@ -146,28 +177,36 @@ export const isPlaylistUnchanged = (a, b) => a === b ||
...
 
-  media.segments = getAllSegments(media);
+  const mergedPlaylist = mergeOptions(oldMedia, newMedia);
 
   // if the update could overlap existing segment information, merge the two segment lists
-  if (playlist.segments) {
+  if (oldMedia.segments) {
+    if (newMedia.skip) {
```
brandonocasey, line 198: if we have skipped segments, add them back in so that we can merge information from the old media segments into them.

src/playlist-loader.js
```diff
@@ -255,11 +294,48 @@ export default class PlaylistLoader extends EventTarget {
         // only refresh the media playlist if no other activity is going on
         return;
       }
+      const media = this.media();
+
+      let uri = resolveUrl(this.master.uri, media.uri);
+
+      if (this.experimentalLLHLS) {
```
brandonocasey, line 301: This is all new, basically use query directives and server side blocking if available.

src/manifest.js
```diff
@@ -92,6 +93,21 @@ export const parseManifest = ({
     manifest.targetDuration = targetDuration;
   }
 
+  const parts = getLastParts(manifest);
```
brandonocasey, line 96: set a default part target duration for manifests with parts that do not have one.

src/playlist-loader.js
```diff
@@ -69,15 +75,30 @@ export const updateSegment = (a, b) => {
...
-  const result = update.slice();
+  const newSegments = update.slice();
 
   offset = offset || 0;
-  const length = Math.min(original.length, update.length + offset);
+  const result = [];
+
+  let currentMap;
+
+  for (let newIndex = 0; newIndex < newSegments.length; newIndex++) {
```
brandonocasey, line 85: This function was completely changed. Previously we used to create a new array from `newSegments` and then loop over old segments, merging in old timing information and properties. 

Now create an empty array for the result. Loop over every new segment and merge the accompanying old segment if we have one. Otherwise we just push the new segment into the list.

src/playlist-loader.js
```diff
@@ -69,15 +75,30 @@ export const updateSegment = (a, b) => {
...
+
+    if (oldSegment) {
+      currentMap = oldSegment.map || currentMap;
 
-  for (let i = offset; i < length; i++) {
-    const newIndex = i - offset;
+      result.push(updateSegment(oldSegment, newSegment));
+    } else {
+      // carry over map to new segment if it is missing
+      if (currentMap && !newSegment.map) {
```
brandonocasey, line 95: The big push for the refactor of this function was:

using `HLS_skip=YES` also skips init segment maps that have already been seen, so we need to make sure that the map that was in use for the last old segment is merged into new segments.

src/playlist-loader.js
gkatsev: thought (non-blocking): yeah, I'm sure have a lot of places where the terminology is confusing. We probably want to write up some guidelines around naming, particularly around words like `media`, `playlist`, etc.

src/playlist-loader.js
gkatsev: question: do we want to skip these two server-control checks if we have an end list tag? The spec says that the server MUST ignore these, so, we may as well not do the calculations in those cases, if we don't already.

gkatsev:
Also, was trying out https://conventionalcomments.org/. LMK what you think.

src/playlist-loader.js
```diff
@@ -34,6 +35,12 @@ export const updateSegment = (a, b) => {
 
   const result = mergeOptions(a, b);
 
+  // if only the old segment has preload hints
+  // and the new one does not, remove preload hints.
+  if (a.preloadHints && !b.preloadHints) {
+    delete result.preloadHints;
+  }
```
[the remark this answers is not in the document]
brandonocasey: Yes usually because the new segment added all the `preloadHints` to the `parts` list.

src/playlist-loader.js
brandonocasey: sure I will make this an `&& !endList`

src/manifest.js
```diff
@@ -92,6 +93,21 @@ export const parseManifest = ({
...
+
+  if (parts.length && !manifest.partTargetDuration) {
+    let partTargetDuration = (manifest.targetDuration / parts.length);
+
+    if (manifest.segments && manifest.segments.length) {
+      partTargetDuration = parts.reduce((acc, p) => Math.max(acc, p.duration), 0);
+    }
```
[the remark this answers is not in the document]
brandonocasey: I think this logic was a little half baked here. Going to switch it so that `partTargetDuration` is just the largest part duration that we have at the time of parsing, as we will only add a partTargetDuration if we also have parts.

src/playlist-loader.js
```diff
@@ -115,12 +149,27 @@ export const resolveSegmentUris = (segment, baseUri) => {
...
-  if (media.preloadSegment && media.preloadSegment.parts) {
-    segments.push(media.preloadSegment);
+  if (preloadSegment && preloadSegment.parts && preloadSegment.parts.length) {
+    // if preloadHints has a MAP that means that the
+    // init segment is going to change. We cannot use any of the parts
+    // from this preload segment.
+    if (preloadSegment.preloadHints) {
+      for (let i = 0; i < preloadSegment.preloadHints.length; i++) {
+        if (preloadSegment.preloadHints[i].type === 'MAP') {
+          return segments;
```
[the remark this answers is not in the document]
brandonocasey: So this is a special case where the preloadHint indicates that that the init segment is going to be changing. I thought it would be better to not try and use a preload segment that might have the map change as we need to have the changed map before we can append any parts that use it.

src/playlist-loader.js
```diff
@@ -181,13 +243,13 @@ export const updateMaster = (master, media, unchangedCheck = isPlaylistUnchanged
...
     }
   }
-  result.playlists[media.id] = mergedPlaylist;
+  result.playlists[newMedia.id] = mergedPlaylist;
   // URI reference added for backwards compatibility
-  result.playlists[media.uri] = mergedPlaylist;
+  result.playlists[newMedia.uri] = mergedPlaylist;
 
   // update media group playlist references.
   forEachMediaGroup(master, (properties, mediaType, groupKey, labelKey) => {
```
gkatsev, line 255: chore: in this fold, on linees 260 and 261 in the loops, we still refer to `media` but presumably, it should refer to `newMedia`? I can't comment on the direct line because of github. this is also why the build is failing on this PR.

gkatsev (approved):

src/playlist-loader.js
```diff
@@ -116,12 +150,27 @@ export const resolveSegmentUris = (segment, baseUri) => {
...
+    // from this preload segment.
+    if (preloadSegment.preloadHints) {
+      for (let i = 0; i < preloadSegment.preloadHints.length; i++) {
+        if (preloadSegment.preloadHints[i].type === 'MAP') {
+          return segments;
+        }
+      }
+    }
+    // set the duration for our preload segment to target duration.
+    preloadSegment.duration = media.targetDuration;
```
gesinger, line 170: Is there a reason we need a duration on preload segments?

src/playlist-loader.js
```diff
@@ -147,28 +196,41 @@ export const isPlaylistUnchanged = (a, b) => a === b ||
...
     return null;
   }
 
-  const mergedPlaylist = mergeOptions(playlist, media);
+  newMedia.segments = getAllSegments(newMedia);
 
-  media.segments = getAllSegments(media);
+  const mergedPlaylist = mergeOptions(oldMedia, newMedia);
+
+  // always use the new medias preload segment.
```
gesinger, line 215:
```suggestion
  // always use the new media's preload segment
```

src/playlist-loader.js
```diff
@@ -147,28 +196,41 @@ export const isPlaylistUnchanged = (a, b) => a === b ||
...
+    delete mergedPlaylist.preloadSegment;
+  }
 
   // if the update could overlap existing segment information, merge the two segment lists
-  if (playlist.segments) {
+  if (oldMedia.segments) {
+    if (newMedia.skip) {
+      newMedia.segments = newMedia.segments || [];
+      // add back in objects for skipped segments, so that we merge
+      // old properties into this new segment
```
gesinger, line 225:
```suggestion
      // old properties into the new segments
```

src/playlist-loader.js
```diff
@@ -147,28 +196,41 @@ export const isPlaylistUnchanged = (a, b) => a === b ||
...
+      for (let i = 0; i < newMedia.skip.skippedSegments; i++) {
+        newMedia.segments.unshift({skipped: true});
```
gesinger, line 227: Is the `skipped` property for debugging?

src/playlist-loader.js
```diff
@@ -268,11 +330,60 @@ export default class PlaylistLoader extends EventTarget {
         // only refresh the media playlist if no other activity is going on
         return;
       }
+      const media = this.media();
+
+      let uri = resolveUrl(this.master.uri, media.uri);
+
+      if (this.experimentalLLHLS && !media.endList) {
+        const query = [];
```
gesinger, line 338: The constructor is getting a bit long. May be worth breaking this into its own function. Can even be outside  of the class and return an object with necessary details to make testing easier.

src/playlist-loader.js
```diff
@@ -268,11 +330,60 @@ export default class PlaylistLoader extends EventTarget {
...
+            // if we are requesting a nextPart or preload segment was added
+            // to our segment list. We are requesting a part of the preload segment
+            // or the full preload segment. Either way we need to go down by 1
+            // in nextMSN
+            if (nextPart > -1 || parts.length) {
+              nextMSN--;
+            }
+          }
+
+          // add _HLS_msn= infront of any _HLS_part query
```
gesinger, line 367:
```suggestion
          // add _HLS_msn= in front of any _HLS_part query
```

src/playlist-loader.js
```diff
@@ -268,11 +330,60 @@ export default class PlaylistLoader extends EventTarget {
...
+          query.unshift(`_HLS_msn=${nextMSN}`);
+        }
+
+        if (media.serverControl && media.serverControl.canSkipUntil) {
+          // add _HLS_skip= infront of all other queries.
+          query.unshift('_HLS_skip=' + (media.serverControl.canSkipDateranges ? 'v2' : 'YES'));
```
gesinger, line 373: In order to separate concerns, maybe we could have a separate function which can add LL-HLS params to a URI.

src/playlist-loader.js
```diff
@@ -268,11 +330,60 @@ export default class PlaylistLoader extends EventTarget {
...
+            // and we need to add the _HLS_part= query
+            if (nextPart > -1 && nextPart !== (parts.length - 1)) {
+              // add existing parts to our preload hints
+              query.push(`_HLS_part=${nextPart}`);
+            }
+
+            // if we are requesting a nextPart or preload segment was added
+            // to our segment list. We are requesting a part of the preload segment
+            // or the full preload segment. Either way we need to go down by 1
+            // in nextMSN
```
gesinger, line 361: Is this because the `nextMSN` would be (and normally is, outside of this case) reflecting the predicted next media sequence number, rather than one that's reported in the playlist (the preload segment)?

test/playlist-loader.test.js
```diff
@@ -469,1593 +423,2109 @@ QUnit.test('updateMaster changes old values', function(assert) {
...
 
-QUnit.test(
-  'can switch playlists immediately after the master is downloaded',
-  function(assert) {
-    const loader = new PlaylistLoader('master.m3u8', this.fakeVhs);
+  QUnit.test('#EXT-X-SKIP does not add initial empty segments', function(assert) {
+    this.fakeVhs.options_ = {experimentalLLHLS: true};
+    const loader = new PlaylistLoader('media.m3u8', this.fakeVhs);
 
     loader.load();
```
gesinger, line 2068: Might be worth pulling these all into a `beforeEach` since they're done in each test.

test/playlist-loader.test.js
```diff
@@ -469,1593 +423,2109 @@ QUnit.test('updateMaster changes old values', function(assert) {
...
 
-    loader.on('loadedplaylist', function() {
-      loader.media('1-high.m3u8');
-    });
-    this.requests.pop().respond(
+    // no newline
```
gesinger, line 2070: Don't these all have newlines? Would it be worth using ``` instead?

test/playlist-loader.test.js
```diff
@@ -469,1593 +423,2109 @@ QUnit.test('updateMaster changes old values', function(assert) {
...
-                              'high-0.ts\n'
-  );
-  assert.strictEqual(loader.state, 'HAVE_METADATA', 'switched active media');
-  assert.strictEqual(
-    loader.media(),
-    loader.master.playlists[1],
-    'updated the active media'
-  );
-});
+    // no newline
```
gesinger, line 2096: What's this newline for?

test/playlist-loader.test.js
```diff
@@ -469,1593 +423,2109 @@ QUnit.test('updateMaster changes old values', function(assert) {
...
+    assert.deepEqual(
+      media.preloadSegment,
+      media.segments[media.segments.length - 1],
+      'last segment is preloadSegment'
     );
-    loader.media('1-high.m3u8');
-    this.requests.pop().respond(
+  });
+
+  QUnit.test('#EXT-X-PRELOAD without parts not to added to segment list', function(assert) {
```
gesinger, line 2169:
```suggestion
  QUnit.test('#EXT-X-PRELOAD without parts not added to segment list', function(assert) {
```

test/playlist-loader.test.js
```diff
@@ -469,1593 +423,2109 @@ QUnit.test('updateMaster changes old values', function(assert) {
...
-      'switched to loaded playlist'
-    );
-  }
-);
+    assert.equal(segments.length, 4, '4 segments');
+    assert.notOk(segments[0].parts, 'no parts for first segment');
+    assert.equal(segments[1].parts.length, 2, 'parts for second segment');
+    assert.equal(segments[2].parts.length, 2, 'parts for third segment');
+    assert.equal(segments[3].parts.length, 2, 'parts for forth segment');
+  });
```
gesinger, line 2257: This looks like the same test as the prior one.

```suggestion
```

test/playlist-loader.test.js
```diff
@@ -469,1593 +423,2109 @@ QUnit.test('updateMaster changes old values', function(assert) {
...
+      '#EXTINF:2\n' +
+      'segment5.ts\n' +
+      '#EXT-X-PART:URI="segment6-part1.ts",DURATION=1\n' +
+      '#EXT-X-PART:URI="segment6-part2.ts",DURATION=1\n' +
+      'segment6.ts\n' +
+      '#EXT-X-PART:URI="segment7-part1.ts",DURATION=1\n' +
+      '#EXT-X-PART:URI="segment7-part2.ts",DURATION=1\n' +
+      'segment7.ts\n' +
+      '#EXT-X-PART:URI="segment8-part1.ts",DURATION=1\n' +
+      '#EXT-X-PRELOAD-HINT:TYPE="PART",URI="segment8-part2.ts"\n'
```
gesinger, line 2524: Non blocking, but since these tests use a pretty standard form, it might be worth creating a function to generate the manifest strings (num segments, num parts, preload hints, etc.).

src/manifest.js
```diff
@@ -92,6 +93,17 @@ export const parseManifest = ({
     manifest.targetDuration = targetDuration;
   }
 
+  const parts = getLastParts(manifest);
+
+  if (parts.length && !manifest.partTargetDuration) {
+    const partTargetDuration = parts.reduce((acc, p) => Math.max(acc, p.duration), 0);
```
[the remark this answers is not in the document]
brandonocasey: In an ideal world I think we would, but we already guess at target duration above this so I think it makes sense to guess at partTargetDuration as well. Might be ok to log an error here though?

src/playlist-loader.js
brandonocasey: Yeah for duration calculations that they will be a part of in playlist.js and elsewhere.

src/playlist-loader.js
brandonocasey: Kind of, but I think it would be helpful to determine if a skipped segment ends up in our segment list with no data other than `skipped: true`. I don't see this happening but it would be nice to know that this segment was skipped and shouldn't have been because we are currently set to the mediaIndex of it.

test/playlist-loader.test.js
brandonocasey: I think this comment was from another test and got moved over to each test when writing them

src/playlist-loader.js
brandonocasey: Yes its based on sequence number and has to take into account weather we our requesting something that was a preload segment, a part of a current segment, or the next segment after what we currently have.

test/playlist-loader.test.js
brandonocasey: It might be nice to do that, but I think that ultimately it will just be a json representation of what you want that has to be built into a string, instead of just building the string directly.

src/playlist-loader.js
brandonocasey: I moved mediaupdatetimeout's handler to a new function and that made it much better.

src/playlist-loader.js
brandonocasey: Done

src/manifest.js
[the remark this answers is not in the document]
gesinger: I think an error would be good in that case. Maybe we can reference the spec too.

src/playlist-loader.js
gesinger: Make sense 👍

src/playlist-loader.js
gesinger: Maybe we can add those notes as a comment?

src/manifest.js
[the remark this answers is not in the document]
gkatsev: We do log `onwarn` below. Is that sufficient?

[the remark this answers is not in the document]
gesinger: I don't think a warning log is sufficient, as this goes against a requirement of the spec.

src/manifest.js
```diff
@@ -100,6 +100,7 @@ export const parseManifest = ({
 
     if (onwarn) {
       onwarn(`manifest has no partTargetDuration defaulting to ${partTargetDuration}`);
+      log.error('llhls manifest has parts but lacks required #EXT-X-PART-INF:PART-TARGET value. Playback is not guaranteed.');
```
gesinger, line 103:
```suggestion
      log.error('LL-HLS manifest has parts but lacks required #EXT-X-PART-INF:PART-TARGET value. See https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-09#section-4.4.3.7. Playback is not guaranteed.');
```

src/playlist-loader.js
```diff
@@ -44,10 +47,15 @@ const addLLHLSQueryDirectives = (uri, media) => {
...
 
-      // if we are requesting a nextPart or preload segment was added
-      // to our segment list. We are requesting a part of the preload segment
-      // or the full preload segment. Either way we need to go down by 1
-      // in nextMSN
+      // this if statement makes sure that we request the msn
+      // of the preload segment if:
+      // 1. the preload segment had parts (and was not yet a full segment)
+      //    but was added to our segments array
+      // 2. the preload segment had preload hints for parts that our not in
```
gesinger, line 54:
```suggestion
      // 2. the preload segment had preload hints for parts that are not in
```

gesinger (approved):

PR #1082, opened 2021-02-24
fix: Add exclude reason and skip duplicate playlist-unchanged

brandonocasey:
Add a reason for exclusions and skip `playlist-unchanged` if we have already seen it.

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -727,15 +727,12 @@ export default class DashPlaylistLoader extends EventTarget {
       mpl.createMupOnMedia_ = null;
     }
 
-    // clear any pending timeouts, master was just updated
-    // so if minimumUpdatePeriodTimeout_ is still valid
-    window.clearTimeout(mpl.minimumUpdatePeriodTimeout_);
-    mpl.minimumUpdatePeriodTimeout_ = null;
-    mpl.createMUPTimeout_();
-  }
+    // clear any pending timeouts
+    if (mpl.minimumUpdatePeriodTimeout_) {
+      window.clearTimeout(mpl.minimumUpdatePeriodTimeout_);
+      mpl.minimumUpdatePeriodTimeout_ = null;
+    }
 
-  createMUPTimeout_() {
-    const mpl = this.masterPlaylistLoader_;
     let mup = mpl.master && mpl.master.minimumUpdatePeriod;
 
     // If the minimumUpdatePeriod has a value of 0, that indicates that the current
@@ -751,21 +748,23 @@ export default class DashPlaylistLoader extends EventTarget {
       }
     }
 
-    // if minimumUpdatePeriod is invalid or <= zero skip creating a timeout
+    // if minimumUpdatePeriod is invalid or <= zero, which
+    // can happen when a live video becomes VOD. skip timeout
+    // creation.
     if (typeof mup !== 'number' || mup <= 0) {
       return;
     }
 
+    this.createMUPTimeout_(mup);
+  }
+
+  createMUPTimeout_(mup) {
+    const mpl = this.masterPlaylistLoader_;
+
     mpl.minimumUpdatePeriodTimeout_ = window.setTimeout(() => {
       mpl.minimumUpdatePeriodTimeout_ = null;
-      mpl.createMUPTimeout_();
-
-      // if a minimumUpdatePeriodTimeout_ was created again
-      // then the minimumUpdatePeriod was valid
-      // trigger minimumUpdatePeriod
-      if (mpl.minimumUpdatePeriodTimeout_) {
-        mpl.trigger('minimumUpdatePeriod');
-      }
+      mpl.trigger('minimumUpdatePeriod');
+      mpl.createMUPTimeout_(mup);
     }, mup);
   }
 
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -19,6 +19,7 @@ import {
 } from './manifest';
 import containerRequest from './util/container-request.js';
 import {toUint8} from '@videojs/vhs-utils/es/byte-helpers';
+import logger from './util/logger';
 
 const { EventTarget, mergeOptions } = videojs;
 
@@ -291,6 +292,7 @@ export default class DashPlaylistLoader extends EventTarget {
 
     this.state = 'HAVE_NOTHING';
     this.loadedPlaylists_ = {};
+    this.logger_ = logger('DashPlaylistLoader');
 
     // initialize the loader state
     // The masterPlaylistLoader will be created with a string
@@ -752,6 +754,9 @@ export default class DashPlaylistLoader extends EventTarget {
     // can happen when a live video becomes VOD. skip timeout
     // creation.
     if (typeof mup !== 'number' || mup <= 0) {
+      if (mup < 0) {
+        this.logger_('minimumUpdatePeriod is less then 0 and invalid');
+      }
       return;
     }
 
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -419,6 +419,11 @@ export default class DashPlaylistLoader extends EventTarget {
     this.mediaRequest_ = null;
     this.minimumUpdatePeriodTimeout_ = null;
 
+    if (this.masterPlaylistLoader_.createMupOnMedia_) {
+      this.off('loadedmetadata', this.masterPlaylistLoader_.createMupOnMedia_);
+      this.masterPlaylistLoader_.createMupOnMedia_ = null;
+    }
+
     this.off();
   }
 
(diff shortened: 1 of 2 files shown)


src/master-playlist-controller.js
```diff
@@ -1066,6 +1075,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // Blacklist this playlist
     currentPlaylist.excludeUntil = Date.now() + (blacklistDuration * 1000);
+    if (error.reason) {
+      currentPlaylist.lastExcludeReason = error.reason;
```
gesinger, line 1079: For the short term, do you think we should add this as `lastExcludeReason_` to prevent others from using it? Or maybe we should decide if it would make sense to have an array of reasons/what we want a public API on playlists should be.

gesinger (commented):
Would be worth a test.

src/dash-playlist-loader.js
```diff
@@ -705,6 +705,9 @@ export default class DashPlaylistLoader extends EventTarget {
   }
 
   updateMinimumUpdatePeriodTimeout_() {
+    if (!this.isMaster_) {
```
brandonocasey, line 708: only add a minimum update period timer on the master playlist object.

src/dash-playlist-loader.js
```diff
@@ -728,7 +731,7 @@ export default class DashPlaylistLoader extends EventTarget {
       // If we haven't yet selected a playlist, wait until then so we know the
       // target duration
       if (!this.media()) {
-        this.one('loadedplaylist', () => {
+        this.one('loadedmetadata', () => {
```
brandonocasey, line 734: When media loads on the master playlist for the first time it's called `loadedmetadata`. `loadedplaylist` signifies the load of the master playlist.

src/dash-playlist-loader.js
```diff
@@ -791,9 +794,10 @@ export default class DashPlaylistLoader extends EventTarget {
       this.trigger('playlistunchanged');
     }
 
-    if (!this.media().endList) {
+    if (!this.media().endList && !this.mediaUpdateTimeout) {
```
brandonocasey, line 797: only have one `mediaupdatetimeout` running at a time for an object. Mutex's ftw

src/master-playlist-controller.js
brandonocasey: I think making it an underscore works as I think our whole exclusion mechanism needs a rework and I don't want anyone to rely on it.

gkatsev: yeah, let's keep it private for now. We should think about this more formally for the longer term.

src/dash-playlist-loader.js
```diff
@@ -791,9 +794,10 @@ export default class DashPlaylistLoader extends EventTarget {
...
       this.mediaUpdateTimeout = window.setTimeout(() => {
         this.trigger('mediaupdatetimeout');
+        this.mediaUpdateTimeout = null;
```
gesinger, line 800: Will this clear out `this.mediaUpdateTimeout` only after the call to `refreshMedia_`? Would the check in the conditional above then block the new timeout?

brandonocasey: You are right. I just realized this with some testing. Using the (netlify deploy to test some drm that requires https with these changes)

gkatsev (dismissed):
LGTM, but I'll defer to @gesinger

src/dash-playlist-loader.js
```diff
@@ -696,45 +710,63 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    // cancel any pending creation of mup on media
+    // a new one will be added if needed.
+    if (mpl.createMupOnMedia_) {
+      mpl.off('loadedmetadata', mpl.createMupOnMedia_);
+      mpl.createMupOnMedia_ = null;
+    }
 
-    const minimumUpdatePeriod = this.masterPlaylistLoader_.master && this.masterPlaylistLoader_.master.minimumUpdatePeriod;
+    // clear any pending timeouts, master was just updated
+    // so if minimumUpdatePeriodTimeout_ is still valid
```
gesinger, line 731: I'm not sure if I understand this comment. Is it describing the scenario that, since the master playlist was just updated, this loader needs to check to see if the master playlist changed its minimum update period, and therefore should update its minimum update period timeout?

src/dash-playlist-loader.js
```diff
@@ -696,45 +710,63 @@ export default class DashPlaylistLoader extends EventTarget {
...
+      if (mpl.media()) {
+        mup = mpl.media().targetDuration * 1000;
       } else {
-        createMUPTimeout(this.media().targetDuration * 1000);
+        mpl.createMupOnMedia_ = mpl.updateMinimumUpdatePeriodTimeout_;
+        mpl.one('loadedmetadata', mpl.createMupOnMedia_);
       }
     }
+
+    // if minimumUpdatePeriod is invalid or <= zero skip creating a timeout
```
gesinger, line 754: In what cases can this happen? Removed to become a VOD?

It might be worth some log messages here.

src/dash-playlist-loader.js
```diff
@@ -696,45 +710,63 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    }
+
+    mpl.minimumUpdatePeriodTimeout_ = window.setTimeout(() => {
+      mpl.minimumUpdatePeriodTimeout_ = null;
+      mpl.createMUPTimeout_();
+
+      // if a minimumUpdatePeriodTimeout_ was created again
+      // then the minimumUpdatePeriod was valid
+      // trigger minimumUpdatePeriod
+      if (mpl.minimumUpdatePeriodTimeout_) {
```
gesinger, line 766: It looks like before this change, `minimumUpdatePeriod` was triggered regardless of whether a new one was created. Any specific reason for this change?

src/master-playlist-controller.js
```diff
@@ -472,6 +472,14 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.masterPlaylistLoader_.on('playlistunchanged', () => {
       const updatedPlaylist = this.masterPlaylistLoader_.media();
+
+      // ignore unchanged playlists that have already been
+      // excluded for not-changing. We likely just have a really slowly updating
+      // playlist.
+      if (updatedPlaylist.lastExcludeReason_ === 'playlist-unchanged') {
```
gesinger, line 479: Is it also worth checking the last `excludeUntil` date to see if we're passed the point where the exclusion should be considered?

src/dash-playlist-loader.js
brandonocasey: i'm honestly not sure, that's just what the original code assumed. I would think that it is removed as the source becomes a VOD, and I will clarify that in the comment.

src/master-playlist-controller.js
brandonocasey: 🤷🏻 It might be but I think that `playlist-unchanged` excludes are not likely to stop happening no matter how much time has passed.

src/dash-playlist-loader.js
brandonocasey: Will clarify this comment better.

src/dash-playlist-loader.js
```diff
@@ -751,21 +748,23 @@ export default class DashPlaylistLoader extends EventTarget {
       }
     }
 
-    // if minimumUpdatePeriod is invalid or <= zero skip creating a timeout
+    // if minimumUpdatePeriod is invalid or <= zero, which
+    // can happen when a live video becomes VOD. skip timeout
+    // creation.
```
gesinger, line 753: I think we can use the DASH Interop as a reference:

https://dashif.org/docs/DASH-IF-IOP-v4.2-clean.htm

```
4.6.4.             Transition Phase between Live and On-Demand

In the scenario for which the same MPD URL is used for live and On-Demand content, once the URL and publish time of the last Segment is known for the live service, and the duration of the service is known as well, the service provider acts as defined in clause 4.4.3.1, i.e.,

-          adds the attribute MPD@mediaPresentationDuration
-          removes the attribute MPD@minimumUpdatePeriod
```

If the value is `< 0` we should probably issue a warning, as I don't think that should be valid.

src/dash-playlist-loader.js
```diff
@@ -727,15 +727,12 @@ export default class DashPlaylistLoader extends EventTarget {
...
+    if (mpl.minimumUpdatePeriodTimeout_) {
+      window.clearTimeout(mpl.minimumUpdatePeriodTimeout_);
+      mpl.minimumUpdatePeriodTimeout_ = null;
+    }
 
-  createMUPTimeout_() {
-    const mpl = this.masterPlaylistLoader_;
     let mup = mpl.master && mpl.master.minimumUpdatePeriod;
 
     // If the minimumUpdatePeriod has a value of 0, that indicates that the current
```
gesinger, line 738: It looks like a 0 value is valid for instantaneous updates
> The MPD@minimumUpdatePeriod is present but is recommended to be set to 0 in order to announce instantaneous segment updates.

Should we immediately request instead of waiting for `targetDuration`?

brandonocasey: From what I understand `instantaneous` segment updates will only be released after waiting for the target duration of the last segment in the playlist to expire. So if the last segment is `X` with a target duration of 2s then we need to wait 2s from the current update for the next segment `Z` to be added.

src/dash-playlist-loader.js
```diff
@@ -752,6 +754,9 @@ export default class DashPlaylistLoader extends EventTarget {
     // can happen when a live video becomes VOD. skip timeout
     // creation.
     if (typeof mup !== 'number' || mup <= 0) {
+      if (mup < 0) {
+        this.logger_('minimumUpdatePeriod is less then 0 and invalid');
```
gesinger, line 758:
```
        this.logger_(`found invalid minimumUpdatePeriod of ${mup}, not setting a timeout`);
```

gkatsev:
Ran a live stream for ~20 minutes and it's working fine.

gkatsev (approved):

PR #1083, opened 2021-02-25
chore: use deferred scripts for faster demo startup

gkatsev:


gkatsev:
With a very unscientific test, with Fast 3G, the current system got to a player in ~9s and the new one was ~5s

gesinger (approved):

PR #1085, opened 2021-03-02
Skip over playback gaps that occur in the beginning of streams

### LINKED ISSUE #1084: Auto-played demuxed stream fails to play and infinitely stalls
## Description
The stream in question contains demuxed audio and video streams and there is a small timeline difference of around 1 second between the audio and video streams (see attached screenshot). The stream fails to play with auto-play enabled but plays correctly when auto-play is disabled and the stream is started by user interaction. See [this demo](https://videojs-http-streaming.netlify.app/?debug=false&autoplay=true&muted=false&minified=false&sync-workers=false&liveui=false&partial=false&url=https%3A%2F%2Flivectorprodmedia12-usea.licdn.com%2F0c13be2a-1820-4837-8318-29b9e6187219%2FL4D5ded7c97c980f000-livemanifest.ism%2Fmanifest(format%3Dm3u8-aapl)&type=application%2Fx-mpegURL&keysystems=&buffer-water=false&override-native=true&preload=auto&mirror-source=true) to reproduce the issue.

## Sources
https://livectorprodmedia12-usea.licdn.com/0c13be2a-1820-4837-8318-29b9e6187219/L4D5ded7c97c980f000-livemanifest.ism/manifest(format=m3u8-aapl)

## Steps to reproduce

1. Visit the [http-streaming demo page](https://videojs-http-streaming.netlify.app/)
2. Enter the source: https://livectorprodmedia12-usea.licdn.com/0c13be2a-1820-4837-8318-29b9e6187219/L4D5ded7c97c980f000-livemanifest.ism/manifest(format=m3u8-aapl)
3. Check the `AutoPlay` checkbox
4. Click the "load" button to load the source with the above settings

### Expected
The stream should play like any other demuxed stream.

### Error output
No error output, no failed network requests, nothing obvious found in verbose video.js log output

## Additional Information
I've spent some time digging into this myself and have the below notes. I believe the issue is a race condition, though I haven't gotten to the bottom of the issue yet.

Call stack from AutoPlay case:

1. `setupMasterPlaylistLoaderListeners_` is called
2. `setupFirstPlay` is called
3. MasterPlaylistController.load() called
4. mainSegmentLoader_.load() is called, and succeeds
5. audioSegmentLoader_.load() is called, but returns early because `this.playlist` is undefined (due to race condition, though I don't believe this is an issue as working demuxed follow this same path)
6. *race* media-groups.AUDIO.loadedplaylist event is handled, audio playlist is setup
7. *race* media-groups.AUDIO.loadedmetadata event is handled, audio playlist is setup

In the manual-play case, #6 and #7 happen before #3 and therefore the playlist plays. *NOTE* - Working demuxed sources follow the same path as the above and play fine. There is something happening after #6 and #7 that allows other sources to play, but I haven't figured out what that 'something' is yet.

Latest in repo

image

evanfarina:
## Description
Fixes #1084 . There is an edge case that is exposed when a gap exists at the beginning of the stream and no content is buffered by the time the first `waiting` event has emitted. This leaves the player in a state where it's no longer checking for playback stalls (or gaps).

## Specific Changes proposed
As mentioned in the linked issue, we have decided to proceed by checking for stalls when the first `play` event is emitted.

## Requirements Checklist
- [ x] Feature implemented / Bug fixed
  - [ x] Unit Tests updated or fixed
  - [ ] Docs/guides updated

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -86,6 +86,7 @@ export default class PlaybackWatcher {
 
     this.logger_('initialize');
 
+    const playHandler = () => this.monitorCurrentTime_();
     const canPlayHandler = () => this.monitorCurrentTime_();
     const waitingHandler = () => this.techWaiting_();
     const cancelTimerHandler = () => this.cancelTimer_();
@@ -118,6 +119,12 @@ export default class PlaybackWatcher {
     this.tech_.on('waiting', waitingHandler);
     this.tech_.on(timerCancelEvents, cancelTimerHandler);
     this.tech_.on('canplay', canPlayHandler);
+    // Catch an edge case that occurs when there is a gap at the start of a stream and no content has buffered by the time the first `waiting` event is emitted.
+    // In this case, a `waiting` event is followed by a `play` event. On first play we need to check that playback has not stalled due to a gap, and skip the gap
+    // if it has
+    if (this.tech_.paused()) {
+      this.tech_.one('play', playHandler);
+    }
 
     // Define the dispose function to clean up our events
     this.dispose = () => {
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -133,6 +133,7 @@ export default class PlaybackWatcher {
       this.tech_.off('waiting', waitingHandler);
       this.tech_.off(timerCancelEvents, cancelTimerHandler);
       this.tech_.off('canplay', canPlayHandler);
+      this.tech_.off('play', playHandler);
 
       loaderTypes.forEach((type) => {
         mpc[`${type}SegmentLoader_`].off('appendsdone', loaderChecks[type].updateend);
--- a/test/playback-watcher.test.js
+++ b/test/playback-watcher.test.js
@@ -42,6 +42,97 @@ QUnit.module('PlaybackWatcher', {
   }
 });
 
+QUnit.test('skips over gap at beginning of stream if player is paused', function(assert) {
+  let vhsGapSkipEvents = 0;
+  let hlsGapSkipEvents = 0;
+
+  this.player.autoplay(true);
+
+  this.player.tech_.on('usage', (event) => {
+    if (event.name === 'vhs-gap-skip') {
+      vhsGapSkipEvents++;
+    }
+    if (event.name === 'hls-gap-skip') {
+      hlsGapSkipEvents++;
+    }
+  });
+
+  // set an arbitrary source
+  this.player.src({
+    src: 'master.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+
+  // start playback normally
+  this.player.tech_.triggerReady();
+  this.clock.tick(1);
+  standardXHRResponse(this.requests.shift());
+  openMediaSource(this.player, this.clock);
+  this.player.tech_.trigger('play');
+  this.player.tech_.trigger('waiting');
+  // create a buffer with a gap of 2 seconds at beginning of stream
+  this.player.tech_.buffered = () => videojs.createTimeRanges([[2, 10]]);
+  // Playback watcher loop runs on a 250ms clock and needs run 6 consecutive stall checks before skipping the gap
+  this.clock.tick(250 * 6);
+  // Need to wait for the duration of the gap
+  this.clock.tick(2000);
+
+  assert.equal(vhsGapSkipEvents, 1, 'there is one skipped gap');
+  assert.equal(hlsGapSkipEvents, 1, 'there is one skipped gap');
+
+  // check that player jumped the gap
+  assert.equal(
+    Math.round(this.player.currentTime()),
+    2, 'Player seeked over gap after timer'
+  );
+});
+
+QUnit.test('Does NOT skip over gap at beginning of stream if player is playing', function(assert) {
+  let vhsGapSkipEvents = 0;
+  let hlsGapSkipEvents = 0;
+
+  this.player.autoplay(true);
+  this.player.tech_.paused = () => false;
+
+  this.player.tech_.on('usage', (event) => {
+    if (event.name === 'vhs-gap-skip') {
+      vhsGapSkipEvents++;
+    }
+    if (event.name === 'hls-gap-skip') {
+      hlsGapSkipEvents++;
+    }
+  });
+
+  // set an arbitrary source
+  this.player.src({
+    src: 'master.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+
+  // start playback normally
+  this.player.tech_.triggerReady();
+  this.clock.tick(1);
+  standardXHRResponse(this.requests.shift());
+  openMediaSource(this.player, this.clock);
+  this.player.tech_.trigger('play');
+  this.player.tech_.trigger('waiting');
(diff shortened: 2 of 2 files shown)


test/playback-watcher.test.js
```diff
@@ -42,6 +42,97 @@ QUnit.module('PlaybackWatcher', {
   }
 });
 
+QUnit.test('skips over gap at beginning of stream if player is paused', function(assert) {
```
gesinger, line 45: Minor, but we may want to rename this to something like "skips over gap at beginning of stream if played before content is buffered" as it might reflect the case a bit better.

test/playback-watcher.test.js
```diff
@@ -42,6 +42,97 @@ QUnit.module('PlaybackWatcher', {
...
+  let vhsGapSkipEvents = 0;
+  let hlsGapSkipEvents = 0;
+
+  this.player.autoplay(true);
```
gesinger, line 49: I don't think this is necessary, as the events are triggered ourselves, and the case can still present without autoplay (if someone has a slow enough connection and is fast enough clicking the play button).

src/playback-watcher.js
```diff
@@ -118,6 +119,12 @@ export default class PlaybackWatcher {
     this.tech_.on('waiting', waitingHandler);
     this.tech_.on(timerCancelEvents, cancelTimerHandler);
     this.tech_.on('canplay', canPlayHandler);
+    // Catch an edge case that occurs when there is a gap at the start of a stream and no content has buffered by the time the first `waiting` event is emitted.
+    // In this case, a `waiting` event is followed by a `play` event. On first play we need to check that playback has not stalled due to a gap, and skip the gap
+    // if it has
+    if (this.tech_.paused()) {
+      this.tech_.one('play', playHandler);
```
gesinger, line 126: We may also want to dispose this

test/playback-watcher.test.js
```diff
@@ -42,6 +42,97 @@ QUnit.module('PlaybackWatcher', {
...
+  assert.equal(hlsGapSkipEvents, 1, 'there is one skipped gap');
+
+  // check that player jumped the gap
+  assert.equal(
+    Math.round(this.player.currentTime()),
+    2, 'Player seeked over gap after timer'
+  );
+});
+
+QUnit.test('Does NOT skip over gap at beginning of stream if player is playing', function(assert) {
```
gesinger, line 90: Technically, I think we'd want this case to be handled by the gap skipper too, but the gap skipper is not set up yet at the time of these events. I'm not sure we'd run into it, since the player is marked as playing (paused set to false) before the source is set and ready is triggered.

I think if we ever run into a case where the player is playing before gap skipper is setup, we may want to remove the check for `if (this.tech_.paused()) {`, but until we run into that case, I think it's safer to leave that check there.

It may be best to remove this test and instead just add a test that checks that we don't listen to future `play` events after the first one (since we use the `one` listener) by checking that we don't reset the monitor if we get `play` events in-between.

gesinger (commented):
Thank you for the PR! A few minor comments.

test/playback-watcher.test.js
evanfarina: Good point, will remove.

test/playback-watcher.test.js
evanfarina: Agree.

src/playback-watcher.js
evanfarina: Yeah, had the same thought but wasn't sure it was necessary. Definitely the safer approach, will add that to existing dispose code

test/playback-watcher.test.js
evanfarina: Removed the test and added a new test which proves that additional `play` events do not speed up the timing of the  playback monitor. Also confirmed that using `.on('play')` results in this test failing.

test/playback-watcher.test.js
```diff
@@ -42,6 +42,97 @@ QUnit.module('PlaybackWatcher', {
...
+  assert.equal(hlsGapSkipEvents, 1, 'there is one skipped gap');
+
+  // check that player jumped the gap
+  assert.equal(
+    Math.round(this.player.currentTime()),
+    2, 'Player seeked over gap after timer'
+  );
+});
+
+QUnit.test('Multiple play events do not cause the gap-skipping logic to be called sooner than expected', function(assert) {
```
gesinger, line 88:
```suggestion
QUnit.test('multiple play events do not cause the gap-skipping logic to be called sooner than expected', function(assert) {
```

test/playback-watcher.test.js
```diff
@@ -42,6 +42,97 @@ QUnit.module('PlaybackWatcher', {
...
+  // Need to wait for the duration of the gap
+  this.clock.tick(2000);
+
+  assert.equal(vhsGapSkipEvents, 1, 'there is one skipped gap');
+  assert.equal(hlsGapSkipEvents, 1, 'there is one skipped gap');
+
+  // check that player jumped the gap
+  assert.equal(
+    Math.round(this.player.currentTime()),
+    2, 'Player seeked over gap after timer'
```
gesinger, line 84:
```suggestion
    2,
    'player seeked over gap after timer'
```

test/playback-watcher.test.js
```diff
@@ -42,6 +42,97 @@ QUnit.module('PlaybackWatcher', {
...
+  // Need to wait for the duration of the gap
+  this.clock.tick(2000);
+
+  assert.equal(vhsGapSkipEvents, 0, 'there is no skipped gap');
+  assert.equal(hlsGapSkipEvents, 0, 'there is no skipped gap');
+
+  // check that player did not skip the gap
+  assert.equal(
+    Math.round(this.player.currentTime()),
+    0, 'Player seeked over gap after timer'
```
gesinger, line 132:
```suggestion
    0,
    'player did not seek over gap'
```

test/playback-watcher.test.js
```diff
@@ -1111,6 +1202,8 @@ QUnit.module('PlaybackWatcher isolated functions', {
       tech: {
         on: () => {},
         off: () => {},
+        one: () => {},
+        paused: () => true,
```
gesinger, line 1206: I think it may make more contextual sense for this to be false for these cases

```suggestion
        paused: () => false,
```

test/playback-watcher.test.js
evanfarina: Good eyes. Updated.

test/playback-watcher.test.js
evanfarina: I guess you're right. I felt weird about defaulting to a playing state but since this is the playlist-watcher and only executes during playback it seems to make sense

gesinger (approved):

gesinger:
All the changes look good, thank you @evanfarina , but one thing I noticed during testing is that if you're using the test page with autoplay on and load the source again (the test source you provided in the issue) instead of refreshing the page, it won't skip the gap the second time. I don't think this should block us from pushing this PR through, but if you wanted to take a look at how source setting behavior might impact things, and if this will still be an issue for your use-case, we can either do it here, have another PR, or create a separate issue and handle it later.

gkatsev (approved):

PR #1087, opened 2021-03-04
Monitor playback for stalls due to gaps in the beginning of stream when a new source is loaded

evanfarina:
## Description
This is a follow-up fix for #[1084](https://github.com/videojs/http-streaming/issues/1084). @gesinger noticed an additional edge that's exposed when a new source is loaded via `player.src()` while the player was playing. This led to our `play` event handler not being called since it was wrapped by a check to ensure the player was in a paused state.

Additionally, I cleaned up a couple of tests by relying on Player API instead of firing HTML5Media events. This ensures the test more closely follows playback behavior in a production environment.

## Specific Changes proposed
Removed the condition around the play event listener in PlaybackWatcher. Since we're using `one` and not `on`, I expect this change to be safe and not cause gaps to be skipped at irregular intervals (a test was also added in my last change to ensure this is true).

Feel free to demo [here](http://localhost:9999/?debug=false&autoplay=true&muted=false&minified=false&sync-workers=false&liveui=false&partial=false&url=https%3A%2F%2Flivectorprodmedia12-usea.licdn.com%2F0c13be2a-1820-4837-8318-29b9e6187219%2FL4D5ded7c97c980f000-livemanifest.ism%2Fmanifest(format%3Dm3u8-aapl)&type=application%2Fx-mpegurl&keysystems=&buffer-water=false&override-native=true&preload=auto&mirror-source=true). Make sure the "AutoPlay" setting is checked. Allow the source to play and then enter the same source in the demo’s “load a URL” input and click “load”. Notice the source does not skip the gap and does not start.

## Requirements Checklist
- [x ] Feature implemented / Bug fixed
  - [ ] Docs/guides updated

### DIFF
--- a/test/playback-watcher.test.js
+++ b/test/playback-watcher.test.js
@@ -26,11 +26,13 @@ QUnit.module('PlaybackWatcher', {
     this.old = {};
 
     // setup a player
-    this.player = createPlayer({html5: {
-      vhs: {
-        overrideNative: true
+    this.player = createPlayer({
+      html5: {
+        vhs: {
+          overrideNative: true
+        }
       }
-    }});
+    });
     this.player.muted(true);
     this.player.autoplay(true);
   },
@@ -105,22 +107,17 @@ QUnit.test('multiple play events do not cause the gap-skipping logic to be calle
     type: 'application/vnd.apple.mpegurl'
   });
 
-  // start playback normally
   this.player.tech_.triggerReady();
   this.clock.tick(1);
   standardXHRResponse(this.requests.shift());
   openMediaSource(this.player, this.clock);
-  this.player.tech_.trigger('play');
-  this.player.tech_.trigger('waiting');
   // create a buffer with a gap of 2 seconds at beginning of stream
   this.player.tech_.buffered = () => videojs.createTimeRanges([[2, 10]]);
   // Playback watcher loop runs on a 250ms clock and needs run 6 consecutive stall checks before skipping the gap
-  // Start with three consecutive playback checks
-  this.clock.tick(250 * 3);
+  // Start with 5 consecutive playback checks
+  this.clock.tick(250 * 5);
   // and then simulate the playback monitor being called 'manually' by a new play event
   this.player.tech_.trigger('play');
-  // Simulate remaining time
-  this.clock.tick(250 * 2);
   // Need to wait for the duration of the gap
   this.clock.tick(2000);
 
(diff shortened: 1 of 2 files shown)


gkatsev (approved):

test/playback-watcher.test.js
```diff
@@ -105,22 +107,17 @@ QUnit.test('multiple play events do not cause the gap-skipping logic to be calle
     type: 'application/vnd.apple.mpegurl'
   });
 
-  // start playback normally
   this.player.tech_.triggerReady();
   this.clock.tick(1);
   standardXHRResponse(this.requests.shift());
   openMediaSource(this.player, this.clock);
-  this.player.tech_.trigger('play');
```
gesinger, line 113: Was the play event removed from this test for any reason?

test/playback-watcher.test.js
```diff
@@ -133,6 +130,102 @@ QUnit.test('multiple play events do not cause the gap-skipping logic to be calle
...
+  });
+  // start playback normally
+  this.player.tech_.triggerReady();
+  this.clock.tick(1);
+  standardXHRResponse(this.requests.shift());
+  openMediaSource(this.player, this.clock);
+  this.player.play();
+  this.player.tech_.trigger('waiting');
+  // create a buffer with a gap of 2 seconds at beginning of stream
+  this.player.tech_.buffered = () => videojs.createTimeRanges([[2, 10]]);
```
gesinger, line 187: Minor, but I think it make sense to have the gap reported before the waiting event.

test/playback-watcher.test.js
```diff
@@ -133,6 +130,102 @@ QUnit.test('multiple play events do not cause the gap-skipping logic to be calle
...
+  // Playback watcher loop runs on a 250ms clock and needs run 6 consecutive stall checks before skipping the gap
```
gesinger, line 188:
```suggestion
  // Playback watcher loop runs on a 250ms clock and needs 6 consecutive stall checks before skipping the gap
```

test/playback-watcher.test.js
```diff
@@ -133,6 +130,102 @@ QUnit.test('multiple play events do not cause the gap-skipping logic to be calle
...
+  this.player.currentTime(0);
+
+  this.player.src({
+    src: 'new-master.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+  openMediaSource(this.player, this.clock);
+  this.clock.tick(1);
+
+  // Playback watcher loop runs on a 250ms clock and needs run 6 consecutive stall checks before skipping the gap
```
gesinger, line 215:
```suggestion
  // Playback watcher loop runs on a 250ms clock and needs 6 consecutive stall checks before skipping the gap
```

test/playback-watcher.test.js
evanfarina: Yes, I found it wasn't necessary / had no impact on the test since autoplay is enabled. I mistakenly added it in my last PR

test/playback-watcher.test.js
evanfarina: Updated here and in 3 other locations.

gesinger (approved):

PR #1088, opened 2021-03-04
feat: usage on rendition change

brandonocasey:
## Description
Switch all `this.masterPlaylistLoader.media_` calls to `this.switchMedia_` which will log the cause behind the switch as a triggered usage event.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -301,16 +301,20 @@ export class MasterPlaylistController extends videojs.EventTarget {
     const nextPlaylist = this.selectPlaylist();
 
     if (this.shouldSwitchToMedia_(nextPlaylist)) {
-      this.switchMedia_(nextPlaylist, null, 'abr');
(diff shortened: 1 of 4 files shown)


src/master-playlist-controller.js
```diff
@@ -301,10 +301,18 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
     if (this.shouldSwitchToMedia_(nextPlaylist)) {
-      this.masterPlaylistLoader_.media(nextPlaylist);
+      this.switchMedia_(nextPlaylist, null, 'abr');
     }
   }
 
+  switchMedia_(playlist, delay, cause) {
+    const oldMedia = this.masterPlaylistLoader_.media();
+
+    this.logger_(`switch media ${oldMedia && (oldMedia.id || oldMedia.uri)} -> ${playlist.id || playlist.uri} from ${cause}`);
```
gesinger, line 311: Do we want to log only if the media is different from the prior media?

gkatsev (approved):

Issue #256, opened 2018-10-11
Dash live streaming stops playing after a few seconds


amtins:
## Description
Dash live streaming stops playing after a few seconds

## Sources
Livestreams url updated
~~https://srg2-accountdigitalinnovation-euwe.streaming.media.azure.net/3f17705e-954c-449f-b2f9-21786485668b/269fa71b-c34c-467a-ab3e-908cd37c640b.ism/manifest(format=mpd-time-csf)~~

https://wowzaec2demo.streamlock.net/live/bigbuckbunny/manifest_mpm4sav_mvtime.mpd

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Open https://amtins.github.io/videojs-dash-live-streaming/
2. Click play
3. Wait until it stops

### Expected

Should works as [videojs-contrib-dash](http://videojs.github.io/videojs-contrib-dash/)

Livestreams url updated
Url to test : ~~https://srg2-accountdigitalinnovation-euwe.streaming.media.azure.net/3f17705e-954c-449f-b2f9-21786485668b/269fa71b-c34c-467a-ab3e-908cd37c640b.ism/manifest(format=mpd-time-csf)~~

https://wowzaec2demo.streamlock.net/live/bigbuckbunny/manifest_mpm4sav_mvtime.mpd

## Additional Information
The given stream is a live stream with DVR. The issue stays the same with or without DVR.

### videojs-http-streaming version
videojs-http-streaming 1.2.6/1.5.0

### videojs version
video.js 7.2.4/7.4.0

### Browsers
All

### Platforms
All

albertoronu:
We're having the same issue. Our platform has been using video.js since version 6 to load Facebook DASH preview video and we can't make it work. Upgrading to latest version available v7.3.0. doesn't help neither. 

Video plays for few seconds and then stops triggering the "waiting" event. Putting the same live video on other players that support DASH works though. Please provide some guidance ASAP.

Thanks!

albertoronu:
@gkatsev @heff could you provide some guidance here? 

This is being urgent for our platform. We have been using videojs for the last year without issues but then DASH started having issues without doing any change, and updating to latest version of videojs 6.12.1 + dashjs 2.9.2 or upgrading to v 7.3.0 is not working at all. 

Stream starts playing but thenstops after few seconds and can't  resume.

bansalvks:
I am also facing some issues like
https://github.com/videojs/http-streaming/issues/260

[...]

PR #1092, opened 2021-03-11
fix: support timeshift dash manifests without startNumber attribute

ziddey:
## Description
This should bring support for live DASH assets that utilize timeShift (sliding window) and $Time$ as opposed to $Number$ (no startNumber attribute). Would fix https://github.com/videojs/http-streaming/issues/256

For such manifests, mpd-parser will always set `mediaSequence` to 1. Consequently, SegmentLoader will be unaware of any live window shifts.

## Specific Changes proposed
Instead, we need to derive an appropriate `mediaSequence` on each playlist update so the window can shift accordingly. We can take the first segment of the new playlist and find its index in the old playlist (we can search by the segment's `uri`). This is the number of shifts, or change in `mediaSequence`.

In the event the segment isn't found (e.g. network issues preventing manifest update), we can instead advance the entire length of existing segments, which should then simulate falling behind the live window and trigger a reseek.

This derivation only needs to occur for this type of DASH asset, but it should not cause issues otherwise since it should end up deriving the same value. It should generally be sufficient to check that the new playlist has a `mediaSequence` of 1, but maybe there's a more definitive way? We could then also check that the new sequence <= the old one (to exclude a playlist naturally going from 0 to 1).

## Requirements Checklist
  - [ ] Docs/guides updated


ziddey:
Can `segments` ever evaluate false? If it's always an array, even if empty, we could just check `b.mediaSequence === 1 && b.segments.length`. That would present the edge case where the playlist goes empty, and we'd maybe still want to increment the sequence?

In general, it may be worth considering scenarios where a playlist goes from not having segments to having segments and vice versa and how to handle setting the `mediaSequence`

gkatsev:
PRs are also built automatically via netlify, so, your changes are lives here https://deploy-preview-1092--videojs-http-streaming.netlify.app/ (see the netlify checks)

ziddey:
I believe I've identified the next issue. For an HLS stream, when changing quality, the media manifest for the old quality is no longer refreshed. Unfortunately, this is also being applied for DASH streams.

ApplicationFrameHost_m4MbM9Pm3Z

Note this isn't an issue with the Unified Streaming Live DASH stream since it only has one quality.

ziddey:
Added a `sourceType` check to `onGroupChanged` to not stop the playlist loader for dash sources.

What ramifications does this have on switching from demuxed to muxed audio? As it was, I'm not sure those conditions would have evaluated true.

PR #1093, opened 2021-03-11
fix: retry appends on QUOTA_EXCEEDED_ERR

gesinger:
## Requirements Checklist
  - [] Docs/guides updated


src/segment-loader.js
```diff
@@ -2044,34 +2063,138 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    }
+
+    if (error.code === QUOTA_EXCEEDED_ERR) {
+      this.handleQuotaExceededError_({segmentInfo, type, bytes});
+      // A quota exceeded error should be recoverable with a future re-append, so no need
+      // to trigger an append error.
+      return;
+    }
+
+    this.error(`${type} append of ${bytes.length}b failed for segment ` +
```
gkatsev, line 2162: should we log the error itself too? Maybe the message or callstack in it could have more info for us?

src/segment-loader.js
```diff
@@ -2044,34 +2063,138 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-    // haven't yet tested that, and prepending is how we have always done things.
-    const bytes = concatSegments({
-      bytes: byteLength,
-      segments
+    const audioBufferStart = audioBuffered.length ? audioBuffered.start(0) : 0;
+    const audioBufferEnd = audioBuffered.length ?
+      audioBuffered.end(audioBuffered.length - 1) : 0;
+    const videoBufferStart = videoBuffered.length ? videoBuffered.start(0) : 0;
+    const videoBufferEnd = videoBuffered.length ?
+      videoBuffered.end(videoBuffered.length - 1) : 0;
```
gkatsev, line 2087: any issues with just clearing from `0` to `now-MIN_BACK_BUFFER?`?

src/segment-loader.js
```diff
@@ -2044,34 +2063,138 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    const timeToRemoveUntil = currentTime - MIN_BACK_BUFFER;
+
+    this.logger_(`On QUOTA_EXCEEDED_ERR, removing video from 0 to ${timeToRemoveUntil}`);
+    this.sourceUpdater_.removeVideo(0, timeToRemoveUntil, () => {
+      this.logger_(`On QUOTA_EXCEEDED_ERR, removing audio from 0 to ${timeToRemoveUntil}`);
+      this.sourceUpdater_.removeAudio(0, timeToRemoveUntil, () => {
+        this.logger_(`On QUOTA_EXCEEDED_ERR, retrying append in ${MIN_BACK_BUFFER}s`);
+        this.waitingOnRemove_ = false;
+        // wait the length of time alotted in the back buffer to prevent wasted
+        // attempts (since we can't clear less than the minimum)
```
gkatsev, line 2139: shouldn't we be waiting based on the forward buffer? Though, waiting 1s each time makes sense. Just not sure it makes say to make it based on back buffer.

gkatsev: Also, can't we wait for `updateend` to know that stuff have been removed?

src/segment-loader.js
gkatsev: Ah, I see. This is to see if the segment is too large and removing the back buffer won't help.

test/source-updater.test.js
```diff
@@ -1313,3 +1314,38 @@ QUnit[testOrSkip]('audio appends are delayed until video append for the first ap
...
+    video: 'avc1.4D001E'
+  });
+
+  const videoBuffer = {
+    appendBuffer() {
+      const quotaExceededError = new Error();
+
+      quotaExceededError.code = QUOTA_EXCEEDED_ERR;
+
+      throw quotaExceededError;
```
gkatsev, line 1332: not that we should change this, but the alternative would've been `throw {code: QUOTA_EXCEEDED_ERR};`

gkatsev:
This PR makes Boat 4K play, fully, in Chrome without any errors in the console. In firefox, it now fails with `Playback cannot continue. No available working or supported playlists.`.

gkatsev:
Overall, definitely an improvement. We definitely should revisit this to try doing one or both of these:
- downswitch if we continue encountering this error
- break up large segments to make smaller appends. This potentially needs us to do a binary split appends or something to find optimal size. Like, first try and split segments in half, if we still got the error, split in fourths, then eights, etc until single  GOP. Though, at some point downswitching (if we have a lower rendition) would definitely be better. This may be something that we try harder if say a user has chosen a particular playlist.

test/source-updater.test.js
gesinger: The `vjsstandard` linter yelled at me for doing that at first (which I was surprised by, but changing to an actual error helped).

src/segment-loader.js
gesinger: The forward buffer may be changing in the case of demuxed content.

And `updateend` won't always fire for the removes (particularly if the buffer doesn't exist).

src/segment-loader.js
gesinger: Yeah, probably a good idea to add that in.

src/segment-loader.js
```diff
@@ -2044,34 +2063,138 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      // Can't remove enough buffer to make room for new segment (or the browser doesn't
+      // allow for appends of segments this size). In the future, it may be possible to
+      // split up the segment and append in pieces, but for now, error out this playlist
+      // in an attempt to switch to a more manageable rendition.
+      this.logger_('On QUOTA_EXCEEDED_ERR, single segment too large to append to ' +
+        'buffer, triggering an error. ' +
+        `Appended byte length: ${bytes.byteLength}, ` +
+        `audio buffer: ${timeRangesToArray(audioBuffered).join(', ')}, ` +
+        `video buffer: ${timeRangesToArray(videoBuffered).join(', ')}, `);
+      this.error({
```
brandonocasey, line 2102: I wonder if we should try splitting the single append into two appends and halving the segment.

gesinger: Yeah, I added that as a comment above. I think it would be good to try as a TODO, but I'm hesitant to do it just yet as I'm not sure how different browsers will behave if we append bytes that aren't split up on proper box boundaries.

brandonocasey (approved):

PR #1097, opened 2021-03-12
refactor: checkBuffer_/fillBuffer_/generateSegmentInfo

brandonocasey:
Refactor segment selection logic into (mostly) one place.


src/segment-loader.js
```diff
@@ -1505,9 +1458,40 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        startOfSegment,
+        buffered: this.buffered_(),
+        overrideCheck: this.isPendingTimestampOffset_
+      });
+      this.isPendingTimestampOffset_ = false;
+    }
+
+    const audioBufferedEnd = lastBufferedEnd(this.sourceUpdater_.audioBuffered());
+
+    if (typeof audioBufferedEnd === 'number') {
```
gkatsev, line 1478: why did the logic for this part change?

gkatsev: ah, I see. `lastBufferedEnd` already does a null check and so if we don't get a number from it, we don't have audio buffered but if we do we can just use that value directly in the next step.

src/vtt-segment-loader.js
```diff
@@ -229,12 +221,17 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
   skipEmptySegments_(segmentInfo) {
     while (segmentInfo && segmentInfo.segment.empty) {
-      segmentInfo = this.generateSegmentInfo_(
-        segmentInfo.playlist,
-        segmentInfo.mediaIndex + 1,
-        segmentInfo.startOfSegment + segmentInfo.duration,
-        segmentInfo.isSyncRequest
-      );
+      // stop at the last possible segmentInfo
+      if (segmentInfo.mediaIndex + 1 >= segmentInfo.playlist.segments.length) {
```
gkatsev, line 225: Looks like codecov is saying both these changes aren't covered by tests? Do we have any tests for this piece?

gkatsev (commented):
Looks like there's a merge conflict.
Most of my comments are around code coverage being missing. We should try and add tests if we can, if it's complicated to test a line, we can discuss it. It's definitely ok not to appease the code coverage just because if it's going to be a lot of work.

src/segment-loader.js
```diff
@@ -2325,41 +2304,11 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     if (!this.hasEnoughInfoToLoad_()) {
       this.loadQueue_.push(() => {
-        const buffered = this.buffered_();
-
-        if (typeof segmentInfo.timestampOffset === 'number') {
-          // The timestamp offset needs to be regenerated, as the buffer most likely
```
brandonocasey, line 2331: generateSegmentInfo, already did all of this.

test/vtt-segment-loader.test.js
```diff
@@ -676,17 +676,38 @@ QUnit.module('VTTSegmentLoader', function(hooks) {
...
+        this.requests[0].responseType = 'arraybuffer';
+        this.requests.shift().respond(200, null, new Uint8Array(10).buffer);
+
+        this.clock.tick(1);
+
+        assert.ok(playlist.segments[3].empty, 'marked empty segment as empty');
+
+        assert.ok(
+          !loader.pendingSegment_,
+          'no pending segment, and no error even though last two segments are empty'
```
gesinger, line 709: If there was an error I think it would be on the `loader` itself, which may be worth checking for the second part of this message.

test/segment-loader.test.js
```diff
@@ -62,6 +63,54 @@ SegmentLoader.prototype.addSegmentMetadataCue_ = function() {};
...
+
+  assert.equal(
+    getSyncSegmentCandidate(0, {segments: [
+      {timeline: 0},
+      {timeline: 1},
+      {timeline: 1},
+      {timeline: 1}
+    ]}),
+    0,
+    '0 timeline, four timeline 0 segments, try index 1'
```
gesinger, line 98: I think this comment was copied from the prior test.

src/segment-loader.js
```diff
@@ -2325,41 +2308,10 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-          //
-          // Since the transmuxer is using the actual timing values, but the buffer is
-          // adjusted by the timestamp offset, the value must be adjusted.
-          segmentInfo.audioAppendStart = audioBuffered.end(audioBuffered.length - 1) -
-            this.sourceUpdater_.audioTimestampOffset();
-        }
-
+        // regenerate the audioAppendStart, timestampOffset, etc as they
+        // may have changed since this function was added to the queue.
+        this.isPendingTimestampOffset_ = true;
```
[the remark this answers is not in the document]
brandonocasey: to force timestampoffset to regenerate in generateSegmentInfo.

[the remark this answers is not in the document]
gesinger: It feels a bit strange to set the property here and then clear it as a side effect of `generateSegmentInfo` (we definitely had side effects before, but in the interest of keeping the new function as side effect free as possible). I wonder if it would be better to set `isPendingTimestampOffset_` to `false` in `updateTransmuxerAndRequestSegment_` and then to have another option that can be passed into `generateSegmentInfo_` to regenerate relevant timing info.

src/segment-loader.js
```diff
@@ -1527,11 +1478,53 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      segmentTimeline: segment.timeline,
+      currentTimeline: this.currentTimeline_,
+      startOfSegment,
+      buffered: this.buffered_(),
+      overrideCheck
+    });
+
+    // if timestampoffset was set then we no longer have a timestampoffset
+    if (typeof segmentInfo.timestampOffset === 'number') {
+      this.isPendingTimestampOffset_ = false;
```
gesinger, line 1498: It may be worth moving this out of the function too to make the function have no side effects. I think `generateSegmentInfo` will only be used on `chooseNextRequest` (as part of `fillBuffer`), which has a check for `timestampOffset` right after, and in `loadSegment_` as part of the queue callback, where it was forced to update.

src/segment-loader.js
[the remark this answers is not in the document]
brandonocasey: This was a bit more complicated than I thought, but I generally agreed with your points here. There were two considerations: `vtt-segment-loader` never wants to set timestamp offsets on generated segment info. I got around that by making a member function that vtt-segment-loader will inherit and override to return null.

Second was that whenever `timestampOffset` is set using the `timestampOffsetForSegment` function we expect to set `isPendingTimestampOffset_` to `false`. So we will still have a small side effect no matter what we do, as our code currently expects us to do that. That being said I did add an option to force a timestamp update so that we don't have to set `isPendingTimestampOffset_` where the original comment was.

gesinger (approved):

PR #1098, opened 2021-03-12
Allow clients to limit the number of times a playlist attempts to reload following an error

evanfarina:
## Description

Addresses the issue of unlimited retries mentioned in issue #983 by allowing clients to control how often a playlist will try to reload following an error. Thanks to @gkatsev for the suggestion of using the `playlist` object to keep track of number of retries.

## Specific Changes proposed
Introduce a new option `maxPlaylistRetries` which sets a limit on the number of times a playlist will be retried. Once the number of retries is greater than (>) the value set by `maxPlaylistRetries`, `playlist.excludeUntil` is set to `Infinity`.

Note: A property `retryCount` is set and incremented on `Playlist` objects. I've found that playlist errors and segment errors follow different paths (`playlistRequestError ` and `blacklistCurrentPlaylist`, respectfully) so I'm incrementing this property in each of these places. I'd appreciate it if a reviewer could confirm that no error will result in both of these methods being called as that would cause a single error to increment the retryCount twice.

## Requirements Checklist
- [ x ] Feature implemented / Bug fixed
  - [ x ] Unit Tests updated or fixed
  - [ x ] Docs/guides updated


README.md
```diff
@@ -362,6 +363,13 @@ if a playlist is blacklisted, it will be blacklisted for a period of that
 customized duration. This enables the blacklist duration to be configured
 by the user.
 
+#### maxPlaylistRetries
+* Type: `number`
+* can be used as an initialization option
```
brandonocasey, line 368: Would be good to specify that the default for this is `Infinity`

src/manifest.js
```diff
@@ -132,6 +132,7 @@ export const forEachMediaGroup = (master, callback) => {
  */
 export const setupMediaPlaylist = ({ playlist, uri, id }) => {
   playlist.id = id;
+  playlist.retryCount = 0;
```
brandonocasey, line 135: perhaps it would be better to call this `retries` to match `maxPlaylistRetries`?

src/master-playlist-controller.js
```diff
@@ -1145,7 +1150,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
     }
 
     // Blacklist this playlist
-    currentPlaylist.excludeUntil = Date.now() + (blacklistDuration * 1000);
+    currentPlaylist.excludeUntil = currentPlaylist.retryCount > this.maxPlaylistRetries ? Infinity : Date.now() + (blacklistDuration * 1000);
```
brandonocasey, line 1153: perhaps we should just set `blacklistDuration` to `Infinity` above this rather than adding to this statement?

test/master-playlist-controller.test.js
```diff
@@ -5916,6 +5916,41 @@ QUnit.test('switch playlists if current playlist gets excluded and re-include if
   this.env.log.warn.callCount = 0;
 });
 
+QUnit.test('Playlist is blacklisted indefinitely if number of retries exceeds maxPlaylistRetries', function(assert) {
```
brandonocasey, line 5919: We are trying to move from `blacklist` to `exclude` in our wording. and while it is inconsistent for now, we have it on our roadmap to refactor everything. 
```suggestion
QUnit.test('Playlist is excluded indefinitely if number of retries exceeds maxPlaylistRetries', function(assert) {
```

brandonocasey (commented):
Sorry it took us so long to review. There has been a ton on our plate internally.

README.md
evanfarina: 👍

src/manifest.js
evanfarina: Updated here and throughout

brandonocasey (approved):

test/playlist-loader.test.js
```diff
@@ -1478,6 +1478,78 @@ QUnit.test('clears the update timeout when switching quality', function(assert)
...
+  loader.media(loader.master.playlists[0]);
+
+  this.requests[1].respond(
+    200, null,
+    '#EXTM3U\n' +
+    '#EXT-X-MEDIA-SEQUENCE:0\n' +
+    '#EXTINF:10,\n' +
+    '0.ts\n'
+  );
+  assert.equal(loader.master.playlists[0].retries, 0, 'retries resets to zero when a playlist sucessfully loads');
```
gkatsev, line 1514: Thanks for adding this case, an important one to check so that we don't increment forever.

test/playlist-loader.test.js
```diff
@@ -1478,6 +1478,78 @@ QUnit.test('clears the update timeout when switching quality', function(assert)
...
+
+  // master
+  this.requests.shift().respond(
+    200, null,
+    '#EXTM3U\n' +
+                                '#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=17\n' +
+                                'playlist/playlist.m3u8\n' +
+                                '#EXT-X-STREAM-INF:PROGRAM-ID=2,BANDWIDTH=170\n' +
+                                'playlist/playlist2.m3u8\n' +
+                                '#EXT-X-ENDLIST\n'
```
gkatsev, line 1494: Some weird whitespace here
```suggestion
    '#EXTM3U\n' +
    '#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=17\n' +
    'playlist/playlist.m3u8\n' +
    '#EXT-X-STREAM-INF:PROGRAM-ID=2,BANDWIDTH=170\n' +
    'playlist/playlist2.m3u8\n' +
    '#EXT-X-ENDLIST\n'
```

test/playlist-loader.test.js
```diff
@@ -1478,6 +1478,78 @@ QUnit.test('clears the update timeout when switching quality', function(assert)
...
```
gkatsev, line 1532:
```suggestion
    '#EXTM3U\n' +
    '#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=17\n' +
    'playlist/playlist.m3u8\n' +
    '#EXT-X-STREAM-INF:PROGRAM-ID=2,BANDWIDTH=170\n' +
    'playlist/playlist2.m3u8\n' +
    '#EXT-X-ENDLIST\n'
```

gkatsev (approved):
Aside from the weird test whitespace, LGTM

gkatsev:
The playback tests look a bit suspicious, they don't seem like the standard timeout issue we see.

gkatsev (dismissed):

src/master-playlist-controller.js
```diff
@@ -1132,6 +1135,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return;
     }
 
+    currentPlaylist.retries++;
```
gesinger, line 1138: I think this value may not reflect exactly what we mean. Right now, if we try to request a playlist once and it fails, we will set `retries` to 1. Technically it wasn't a retry, it was just the first attempt. We may actually never retry the playlist if the selection algorithm chooses a different one later on.

In addition, this doesn't take into consideration the blacklist reason, meaning `retries` will be incremented for some cases where we haven't actually retried a playlist (e.g., excessive segment downloads, an audio track failed to load, and early aborts)

gkatsev: > I think this value may not reflect exactly what we mean. Right now, if we try to request a playlist once and it fails, we will set `retries` to 1. Technically it wasn't a retry, it was just the first attempt. We may actually never retry the playlist if the selection algorithm chooses a different one later on.

I think that is fine for this pass, as this retry is playlist dependent. A playlist should be excluded if we have retried it a couple of times, it doesn't matter whether we switched away or not. The name may not be ideal, but I don't think it should stop us from getting this in.

> In addition, this doesn't take into consideration the blacklist reason, meaning `retries` will be incremented for some cases where we haven't actually retried a playlist (e.g., excessive segment downloads, an audio track failed to load, and early aborts)

That's a good point. Are those things we have reasons available for in this method?

gesinger: For some we do, for others we don't. I think we may be best off determining exactly what type of reason(s) we'd want to count for this option, ensure we are passing that reason from callers of `blacklistCurrentPlaylist`, and while we're at it maybe update the `retries` property to something more precise or use `retries` but change our logic around it, since the playlist object is public and we are not marking this property as private. Alternatively, we can mark it as private.

gkatsev: So, the places we call `blacklistCurrentPlaylist` are:
- playback watcher for excessive download (Infinity)
- media groups, error encountered with an audio group
- MPC, on a playlist loader error
- MPC, on playlistunchanged if we determine it's an outdated playlist
- MPC, segment loader error
- MPC, earlyabort (this one is going away)
- MPC, unknown codecs (Infinity)
- MPC, unsupported codecs (Infinity)
- MPC, codec switching not supported (Infinity)

While the name `retires` may not necessarily apply to all those conditions, I think it's reasonable to increment on all of these. Maybe it's unnecessary to increment if we are excluding playlists indefinitely, but I don't think it's harmful.
I think marking `retries` as private is fine too.

gkatsev: Spoke offline, we'll rename the property to `playlistErrors_` to make it pseudo-private and better represent what's actually happening.

gesinger (approved):

PR #1100, opened 2021-03-15
fix: Handle audio only audio track change correctly

brandonocasey:
Allows audio only playlists to change audio tracks. IE an audio only playlist can change from english to spanish correctly now. Would be good to get https://github.com/videojs/mpd-parser/pull/123 in with this.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -4,7 +4,7 @@
 import window from 'global/window';
 import PlaylistLoader from './playlist-loader';
 import DashPlaylistLoader from './dash-playlist-loader';
-import { isEnabled, isLowestEnabledRendition } from './playlist.js';
+import { isEnabled, isLowestEnabledRendition, isAudioOnly } from './playlist.js';
 import SegmentLoader from './segment-loader';
 import SourceUpdater from './source-updater';
 import VTTSegmentLoader from './vtt-segment-loader';
@@ -331,12 +331,13 @@ export class MasterPlaylistController extends videojs.EventTarget {
   }
 
   getAudioTrackPlaylists_() {
-    const playlists = [];
     const master = this.master();
 
-    if (!master && !master.mediaGroups && !master.mediaGroups.AUDIO) {
-      return playlists;
+    if (isAudioOnly(master)) {
+      return [];
     }
+
+    const playlists = [];
 
     const AUDIO = master.mediaGroups.AUDIO;
     const groupKeys = Object.keys(AUDIO);
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -330,6 +330,53 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.abrTimer_ = null;
   }
 
+  getAudioTrackPlaylists_() {
+    const playlists = [];
+    const master = this.master();
+
+    if (!master && !master.mediaGroups && !master.mediaGroups.AUDIO) {
+      return playlists;
+    }
+
+    const AUDIO = master.mediaGroups.AUDIO;
+    const groupKeys = Object.keys(AUDIO);
+    let track;
+
+    // get the current active track
+    if (Object.keys(this.mediaTypes_.AUDIO.groups).length) {
+      track = this.mediaTypes_.AUDIO.activeTrack();
+    // or get the default track from master if mediaTypes_ isn't setup yet
+    } else {
+      // default group is `main` or just the first group.
+      const defaultGroup = AUDIO.main || groupKeys.length && AUDIO[groupKeys[0]];
+
+      for (const label in defaultGroup) {
+        if (defaultGroup[label].default) {
+          track = {label};
+          break;
+        }
+      }
+    }
+
+    if (!track) {
+      return playlists;
+    }
+
+    for (const group in AUDIO) {
+      if (AUDIO[group][track.label]) {
+        const properties = AUDIO[group][track.label];
+
+        if (properties.playlists) {
+          playlists.push.apply(playlists, properties.playlists);
+        } else {
+          playlists.push(properties);
+        }
+      }
+    }
+
+    return playlists;
+  }
+
   /**
    * Register event handlers on the master playlist loader. A helper
    * function for construction time.
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -4,7 +4,7 @@
 import window from 'global/window';
 import PlaylistLoader from './playlist-loader';
 import DashPlaylistLoader from './dash-playlist-loader';
-import { isEnabled, isLowestEnabledRendition, isAudioOnly } from './playlist.js';
+import { isEnabled, isLowestEnabledRendition } from './playlist.js';
 import SegmentLoader from './segment-loader';
 import SourceUpdater from './source-updater';
 import VTTSegmentLoader from './vtt-segment-loader';
@@ -342,14 +342,19 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.abrTimer_ = null;
   }
 
+  /**
+   * Get a list of playlists for the currently selected audio playlist
+   *
+   * @return {Array} the array of audio playlists
+   */
   getAudioTrackPlaylists_() {
     const master = this.master();
 
-    if (isAudioOnly(master)) {
+    // if we don't have any9 demuxed audio tracks
+    // return an empty array
+    if (!master && !master.mediaGroups && !master.mediaGroups.AUDIO) {
       return [];
     }
-
-    const playlists = [];
 
     const AUDIO = master.mediaGroups.AUDIO;
     const groupKeys = Object.keys(AUDIO);
@@ -371,10 +376,15 @@ export class MasterPlaylistController extends videojs.EventTarget {
       }
     }
 
+    // no active track no playlists.
     if (!track) {
-      return playlists;
+      return [];
     }
 
+    const playlists = [];
+
+    // get all of the playlists that are possible for the
+    // active track.
     for (const group in AUDIO) {
       if (AUDIO[group][track.label]) {
         const properties = AUDIO[group][track.label];
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -350,7 +350,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
   getAudioTrackPlaylists_() {
     const master = this.master();
 
-    // if we don't have any9 demuxed audio tracks
+    // if we don't have any demuxed audio tracks
     // return an empty array
     if (!master && !master.mediaGroups && !master.mediaGroups.AUDIO) {
       return [];
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -352,8 +352,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // if we don't have any demuxed audio tracks
     // return an empty array
-    if (!master && !master.mediaGroups && !master.mediaGroups.AUDIO) {
-      return [];
+    if (!master || !master.mediaGroups || !master.mediaGroups.AUDIO) {
+      return master && master.playlists;
     }
 
     const AUDIO = master.mediaGroups.AUDIO;
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -350,10 +350,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
   getAudioTrackPlaylists_() {
     const master = this.master();
 
-    // if we don't have any demuxed audio tracks
-    // return an empty array
+    // if we don't have any audio groups then we can only
+    // assume that the audio tracks are contained in masters
+    // playlist array, use that or an empty array.
     if (!master || !master.mediaGroups || !master.mediaGroups.AUDIO) {
-      return master && master.playlists;
+      return master && master.playlists || [];
     }
 
     const AUDIO = master.mediaGroups.AUDIO;
--- a/src/media-groups.js
+++ b/src/media-groups.js
@@ -769,6 +769,16 @@ export const activeTrack = {
   }
 };
 
+export const getActiveGroup = (type, {mediaTypes}) => () => {
+  const activeTrack_ = mediaTypes[type].activeTrack();
+
+  if (!activeTrack_) {
+    return null;
+  }
+
+  return mediaTypes[type].activeGroup(activeTrack_);
+};
+
 /**
  * Setup PlaylistLoaders and Tracks for media groups (Audio, Subtitles,
  * Closed-Captions) specified in the master manifest.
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -60,9 +60,9 @@
     "@videojs/vhs-utils": "^3.0.0",
     "aes-decrypter": "3.1.2",
     "global": "^4.4.0",
-    "m3u8-parser": "4.6.0",
-    "mpd-parser": "0.15.4",
-    "mux.js": "5.11.0",
+    "m3u8-parser": "4.5.2",
+    "mpd-parser": "0.16.0",
```
gkatsev, line 64: this probably caused a merge conflict

src/playlist-selectors.js
```diff
@@ -143,6 +143,8 @@ export const comparePlaylistResolution = function(left, right) {
  *        Current height of the player element (should account for the device pixel ratio)
  * @param {boolean} limitRenditionByPlayerDimensions
  *        True if the player width and height should be used during the selection, false otherwise
+ * @param {Playlist} [currentMedia]
+ *        The currently selected playlist if it exists.
```
gkatsev, line 147: this doesn't match what's actually passed in

src/playlist-selectors.js
```diff
@@ -166,8 +169,16 @@ export const simpleSelector = function(
...
     limitRenditionByPlayerDimensions
   };
+
+  let playlists = master.playlists;
+
+  // if playlist is audio only, select between currently active audio group playlists.
+  if (Playlist.isAudioOnly(master)) {
+    playlists = masterPlaylistController.getAudioTrackPlaylists_();
+    options.audioOnly = true;
+  }
```
gkatsev, line 179: 👍🏻

gkatsev (commented):
Dont really have much to add. I think it's worth having @gesinger look over this.

src/playlist-selectors.js
```diff
@@ -166,8 +169,16 @@ export const simpleSelector = function(
     height: playerHeight,
     limitRenditionByPlayerDimensions
   };
+
+  let playlists = master.playlists;
+
+  // if playlist is audio only, select between currently active audio group playlists.
+  if (Playlist.isAudioOnly(master)) {
+    playlists = masterPlaylistController.getAudioTrackPlaylists_();
+    options.audioOnly = true;
```
gesinger, line 178: Is this used anywhere?

src/master-playlist-controller.js
```diff
@@ -330,6 +330,53 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.abrTimer_ = null;
   }
 
+  getAudioTrackPlaylists_() {
+    const playlists = [];
+    const master = this.master();
+
+    if (!master && !master.mediaGroups && !master.mediaGroups.AUDIO) {
+      return playlists;
```
gesinger, line 338: When first reading through this function I didn't get why all playlists are returned if there's no audio groups, then later realized it's for audio only playlists. Should we add a check for isAudioOnly as part of this conditional or add a comment to the method to note that this assumes either an active audio group or an audio only playlist?

src/master-playlist-controller.js
```diff
@@ -330,6 +330,53 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      for (const label in defaultGroup) {
+        if (defaultGroup[label].default) {
+          track = {label};
+          break;
+        }
+      }
+    }
+
+    if (!track) {
+      return playlists;
```
gesinger, line 362: Might be a bit easier to read (here and above) to `return []` and create `playlists` after.

src/media-groups.js
```diff
@@ -149,6 +156,18 @@ export const onTrackChanged = (type, settings) => () => {
     return;
   }
 
+  if (activeGroup.masterPlaylist) {
+    if (activeTrack && lastTrack && activeTrack.id !== lastTrack.id) {
+      const mpc = settings.vhs.masterPlaylistController_;
```
gesinger, line 161: I think `masterPlaylistController` is a property in settings.

src/media-groups.js
```diff
@@ -149,6 +156,18 @@ export const onTrackChanged = (type, settings) => () => {
...
+
+      mediaType.logger_(`track change. Switching master audio from ${lastTrack.id} to ${activeTrack.id}`);
+      masterPlaylistLoader.pause();
+      mainSegmentLoader.resetEverything();
+      mpc.fastQualityChange_();
```
gesinger, line 166: `fastQualityChange` should reset everything unless the media is the same. And if the media is the same, would we ever resume playback (the loader would pause but `fastQualityChange` would return early)?

src/media-groups.js
```diff
@@ -658,17 +663,28 @@ export const activeGroup = (type, settings) => (track) => {
...
+
+  if (!variants && groupKeys.length === 1) {
+    // use the main group if it exists
+    if (groups.main) {
+      variants = groups.main;
+    // only one group, use that one
+    } else if (groupKeys.length === 1) {
+      variants = groups[groupKeys[0]];
+    }
+  }
```
gesinger, line 681: It looks like there's a conditional within a repeated check within the conditional. Should the `groupKeys.length === 1` be removed from the outer conditional?

gesinger (commented):
We should probably add some audio only related tests.

src/playlist-selectors.js
brandonocasey: We print `options` to the debug log. I think it would be helpful to know that we are selecting an audio only playlist.

src/media-groups.js
brandonocasey: We will only get in here if the track id changed. which will cause us to select from a new set of playlists in fastQualityChange which will mean that the playlists should never be the same.

brandonocasey: updated the code to check that the playlist will actually change.

src/media-groups.js
```diff
@@ -395,36 +444,19 @@ export const initialize = {
...
-        }
-
         let playlistLoader;
 
-        // if vhs-json was provided as the source, and the media playlist was resolved,
-        // use the resolved media playlist object
-        if (sourceType === 'vhs-json' && properties.playlists) {
+        if (audioOnlyMaster) {
+          logger_(`AUDIO group '${groupId}' label '${variantLabel}' is a master playlist`);
+          properties.masterPlaylist = true;
```
gesinger, line 454: When reading from the top I was a bit confused until I saw that `masterPlaylist` was a boolean. Maybe `isMasterPlaylist`?

src/media-groups.js
```diff
@@ -78,12 +80,21 @@ export const onGroupChanged = (type, settings) => () => {
...
+  }
+
+  mediaType.lastGroup_ = activeGroup;
+  mediaType.lastTrack_ = activeTrack;
 
   stopLoaders(segmentLoader, mediaType);
 
-  if (!activeGroup) {
+  if (!activeGroup || activeGroup.masterPlaylist) {
     // there is no group active
```
gesinger, line 98:
```suggestion
    // there is no group active or active group is the main playlist and won't change
```

src/playlist.js
```diff
@@ -541,6 +542,74 @@ export const isLowestEnabledRendition = (master, media) => {
...
+const playlistMatch = (a, b) => {
+  if (a.id && b.id && a.id === b.id) {
+    return true;
+  }
+
+  if (a.uri && b.uri && a.uri === b.uri) {
+    return true;
+  }
+
+  if (a.resolvedUri && b.resolvedUri && a.resolvedUri === b.resolvedUri) {
```
gesinger, line 554: Can we have a `resolvedUri` without a `uri`?

test/rendition-mixin.test.js
```diff
@@ -57,20 +60,25 @@ const makeMockVhsHandler = function(playlistOptions, handlerOptions) {
...
-
+  vhsHandler.masterPlaylistController_ = mpc;
   vhsHandler.playlists = new videojs.EventTarget();
-  vhsHandler.playlists.master = { playlists: [] };
+
+  vhsHandler.playlists.master = master;
+
+  if (!vhsHandler.playlists.master.playlists) {
+    vhsHandler.playlists.master.playlists = [];
+
```
gesinger, line 80:
```suggestion
```

src/playlist.js
brandonocasey: we can't I am going to move the resolvedUri check above the uri check and add some comments.

gesinger (commented):
Looks good. I think some of the new and modified functions may need some new tests.

test/media-groups.test.js
```diff
@@ -18,1208 +18,1426 @@ QUnit.module('MediaGroups', {
...
-    );
-    assert.equal(
-      result['CLOSED-CAPTIONS'].onTrackChanged.toString(), noopToString,
-      'created noop function for CLOSED-CAPTIONS onTrackChanged'
-    );
-  }
-);
+};
+
+QUnit.module('MediaGroups', function() {
```
brandonocasey, line 23: made all media groups tests live under the same base module.

test/media-groups.test.js
```diff
@@ -18,1208 +18,1426 @@ QUnit.module('MediaGroups', {
...
-    mediaTypes: MediaGroups.createMediaTypes(),
-    masterPlaylistLoader: {
-      media: () => media
-    }
-  };
-  const groups = settings.mediaTypes[type].groups;
-  const tracks = settings.mediaTypes[type].tracks;
-  const activeTrack = MediaGroups.activeTrack[type](type, settings);
-  const activeGroup = MediaGroups.activeGroup(type, settings);
+  ['AUDIO', 'SUBTITLES'].forEach(function(groupType) {
```
brandonocasey, line 253: move all `activeGroup` tests to here and added a few for cases we were missing.

test/media-groups.test.js
```diff
@@ -18,1208 +18,1426 @@ QUnit.module('MediaGroups', {
...
+      assert.equal(
+        segmentLoaderResyncCalls, 1,
+        'resync changing to group with playlist loader'
+      );
+      assert.strictEqual(
+        mediaType.activePlaylistLoader, groups.main[1].playlistLoader,
+        'sets the correct active playlist loader'
+      );
+
+      mediaType.lastGroup_ = null;
```
brandonocasey, line 489: add tests to onGroupChanged for `isMasterPlaylist` type group changes.

src/master-playlist-controller.js
```diff
@@ -352,8 +352,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     // if we don't have any demuxed audio tracks
     // return an empty array
-    if (!master && !master.mediaGroups && !master.mediaGroups.AUDIO) {
-      return [];
+    if (!master || !master.mediaGroups || !master.mediaGroups.AUDIO) {
+      return master && master.playlists;
```
gesinger, line 356: If there's no master, is it OK that this returns `undefined`? Would it be better to return `[]`?

The comment above may need to be changed too otherwise. And would returning `master.playlists` end up returning non audio only tracks?

src/playlist.js
```diff
@@ -563,6 +564,101 @@ export const isLowestEnabledRendition = (master, media) => {
...
+
+      if (audioGroupFound) {
+        continue;
+      }
+    }
+
+    // if we make it here this playlist isn't audio and we
+    // are not audio only
+    return false;
+  }
```
gesinger, line 657: If we can, since this ends up in a nesting of three `for` loops, it may be better to move this to a separate function to check if there's an audio playlist match for a given playlist.

src/master-playlist-controller.js
brandonocasey: I will make it return an empty array without a master. It will could end up returning non-audio playlists but if we don't have audio groups and this function is called, which should only happen if we detect that the playlist is audio only, Then we should just return the main playlists.

src/playlist.js
brandonocasey: added as `someAudioVariant` a function similar to `Array.some`.

test/media-groups.test.js
```diff
@@ -18,1208 +18,1426 @@ QUnit.module('MediaGroups', {
...
+      assert.equal(
+        mainSegmentLoaderResetCalls, 0,
+        'no reset changing from no active playlist loader to group with no playlist loader'
+      );
+      assert.equal(
+        segmentLoaderResyncCalls, 0,
+        'no resync changing to group with no playlist loader'
+      );
+
+      mediaType.lastGroup_ = null;
```
brandonocasey, line 455: pretend that we didn't have a last group.

test/media-groups.test.js
```diff
@@ -18,1208 +18,1426 @@ QUnit.module('MediaGroups', {
...
+        'no reset when active group hasn\'t changed'
+      );
+      assert.equal(segmentLoaderSetAudioCalls.length, 1, 'set audio on track change');
+      assert.ok(segmentLoaderSetAudioCalls[0], 'enabled audio on track change');
+      assert.strictEqual(
+        mediaType.activePlaylistLoader, groups.main[1].playlistLoader,
+        'sets the correct active playlist loader'
+      );
+
+      tracks.fr.enabled = false;
```
brandonocasey, line 622: This test previously relied on weird logic to change tracks, now we actually change the tracks around before each step.

test/media-groups.test.js
```diff
@@ -18,1208 +18,1426 @@ QUnit.module('MediaGroups', {
...
+        assert.equal(mock.calls[key], 0, `${name} ${key} not called`);
+      });
+
+      Object.keys(mock.args).forEach(function(key) {
+        assert.equal(mock.args[key].length, 0, `${name} ${key} not called`);
+      });
+    });
+  };
+
+  QUnit.test('onTrackChanged with isMasterPlaylist', function(assert) {
```
brandonocasey, line 733: The previous `onTrackChanged` test was very brittle, so I didn't want to add anything else to it. I instead added the this test here that tests the new `isMasterPlaylist` functionality.

test/media-groups.test.js
```diff
@@ -18,1208 +18,1427 @@ QUnit.module('MediaGroups', {
...
-    assert.ok(mainSegmentLoaderSetAudioCalls[0], 'main audio set to true');
-    assert.equal(
-      segmentLoaderResetCalls, 0,
-      'no reset changing to group with no playlist loader'
-    );
-    assert.equal(
-      segmentLoaderSetAudioCalls.length, 0,
-      'no audio set when changing to group with no playlist loader'
-    );
+    // verify that all othre mocks are zero
```
gkatsev, line 786:
```suggestion
    // verify that all other mocks are zero
```

test/playlist.test.js
```diff
@@ -1317,4 +1317,95 @@ QUnit.module('Playlist', function() {
...
+
+    a.id = 'bar';
+    assert.equal(Playlist.playlistMatch(a, b), true, 'resolved uri match');
+
+    a.resolvedUri += '?nope';
+    assert.equal(Playlist.playlistMatch(a, b), true, 'uri match');
+
+    a.uri += '?nope';
+
+    assert.equal(Playlist.playlistMatch(a, b), false, 'no match');
```
gkatsev, line 1341: should query parameters result in a uri match?

test/playlist.test.js
```diff
@@ -1317,4 +1317,95 @@ QUnit.module('Playlist', function() {
       'no segments, live delay can\'t be calculated'
     );
   });
+
+  QUnit.test('playlistMatch', function(assert) {
```
gkatsev, line 1321: if playlistMatch returns a boolean, might be better to use `assert.true` and `assert.false`. Same for isAudioOnly.

test/playlist.test.js
gkatsev: I see, resolveUri query param succeeds because the uri still matches but if both have a query param it fails.

gkatsev (approved):

test/playlist.test.js
brandonocasey: i'll switch to that.

gkatsev (approved):

PR #1101, opened 2021-03-16
fix: catch remove errors, remove all data on QUOTA_EXCEEDED

brandonocasey:
## Description
Call remove on QUOTA_EXCEEDED, fix coverage unit tests, catch source updater remove errors

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1101,7 +1101,7 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @param {Function} [done] - an optional callback to be executed when the remove
    * operation is complete
    */
-  remove(start, end, done = () => {}) {
+  remove(start, end, done = () => {}, force = false) {
     // clamp end to duration if we need to remove everything.
     // This is due to a browser bug that causes issues if we remove to Infinity.
     // videojs/videojs-contrib-hls#1225
@@ -1124,7 +1124,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       }
     };
 
-    if (!this.audioDisabled_) {
+    if (force || !this.audioDisabled_) {
       removesRemaining++;
       this.sourceUpdater_.removeAudio(start, end, removeFinished);
     }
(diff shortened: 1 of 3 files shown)


src/segment-loader.js
```diff
@@ -2129,25 +2129,23 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-        // attempts (since we can't clear less than the minimum)
-        this.quotaExceededErrorRetryTimeout_ = window.setTimeout(() => {
-          this.logger_('On QUOTA_EXCEEDED_ERR, re-processing call queue');
-          this.quotaExceededErrorRetryTimeout_ = null;
-          this.processCallQueue_();
-        }, MIN_BACK_BUFFER * 1000);
-      });
-    });
+    this.logger_(`On QUOTA_EXCEEDED_ERR, removing audio/video from 0 to ${timeToRemoveUntil}`);
+    this.remove(0, timeToRemoveUntil, () => {
```
brandonocasey, line 2133: I think we want to remove everything, and it's much easier to use our `remove` function that already does all of that.

src/source-updater.js
```diff
@@ -170,7 +170,11 @@ const actions = {
     }
 
     sourceUpdater.logger_(`Removing ${start} to ${end} from ${type}Buffer`);
-    sourceBuffer.remove(start, end);
+    try {
+      sourceBuffer.remove(start, end);
```
brandonocasey, line 174: Apparently newer versions of chrome can fail to remove on a source buffer if there isn't anything to remove.

test/segment-loader.test.js
```diff
@@ -4174,22 +4174,23 @@ QUnit.module('SegmentLoader', function(hooks) {
...
-        const playlist = playlistWithDuration(40);
-
-        loader.playlist(playlist);
-        loader.load();
-        this.clock.tick(1);
-
-        loader.sourceUpdater_.appendBuffer = ({type, bytes}, callback) => {
-          callback({type: 'QUOTA_EXCEEDED_ERR', code: QUOTA_EXCEEDED_ERR});
-        };
+        return new Promise((resolve, reject) => {
```
brandonocasey, line 4177: We need to add listeners before we do the action for sync worker

src/segment-loader.js
```diff
@@ -1101,7 +1101,7 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @param {Function} [done] - an optional callback to be executed when the remove
    * operation is complete
    */
-  remove(start, end, done = () => {}) {
+  remove(start, end, done = () => {}, force = false) {
```
gesinger, line 1104: Minor, but we may want to update the jsdoc

PR #1105, opened 2021-03-25
fix: better time to first frame for live playlists

brandonocasey:
Does the following:
* ~Fixes some off by 1 errors~
* Loops backwards over segments, rather than forwards when looking for a sync-point in live videos
* Actually uses live delay timing values as live delays, rather than doing it in a round-about fashion.


src/playlist.js
```diff
@@ -104,7 +104,7 @@ const forwardDuration = function(playlist, endSequence) {
   */
 const intervalDuration = function(playlist, endSequence, expired) {
   if (typeof endSequence === 'undefined') {
-    endSequence = playlist.mediaSequence + playlist.segments.length;
+    endSequence = playlist.mediaSequence + (playlist.segments.length - 1);
```
brandonocasey, line 107: endSequence is invalid if we use `playlist.segments.length` as it refers to a zero based index.

src/playlist.js
```diff
@@ -293,13 +293,18 @@ export const playlistEnd = function(playlist, expired, useSafeLiveEnd, liveEdgeP
 
   expired = expired || 0;
 
-  const endSequence = useSafeLiveEnd ? safeLiveIndex(playlist, liveEdgePadding) : playlist.segments.length;
```
brandonocasey, line 296: It doesn't really make sense to get a safe live index, based on how many seconds we want to start behind the playback head. Then get how many seconds that index starts at from behind the end of the playlist. We should instead get the timing information for the end of the playlist. Then subtract the live edge padding from that.

This was causing us to download 3/4 segments before a frame of video would even be shown. Now we download 1.

src/sync-controller.js
```diff
@@ -6,6 +6,12 @@ import {sumDurations} from './playlist';
 import videojs from 'video.js';
 import logger from './util/logger';
 
+const getSegmentIndex = (i, playlist, currentTime = 0) => {
+  const segments = playlist.segments;
+
+  return (playlist.endList || currentTime === 0) ? i : segments.length - (i + 1);
```
brandonocasey, line 12: Loop from the back of the playlist when looking for a live sync point, as we are much more likely to be closer to the end.

src/playlist.js
```diff
@@ -293,13 +293,18 @@ export const playlistEnd = function(playlist, expired, useSafeLiveEnd, liveEdgeP
...
-  return intervalDuration(
+  let lastSegmentTime = intervalDuration(
     playlist,
-    playlist.mediaSequence + endSequence,
+    playlist.mediaSequence + playlist.segments.length - 1,
     expired
   );
+
+  if (useSafeLiveEnd) {
+    liveEdgePadding = typeof liveEdgePadding === 'number' ? liveEdgePadding : playlist.targetDuration * 3;
```
gesinger, line 303:
```suggestion
    // "The client SHALL choose which Media Segment to play first from the
    // Media Playlist when playback starts.  If the EXT-X-ENDLIST tag is not
    // present and the client intends to play the media normally, the client
    // SHOULD NOT choose a segment which starts less than three target
    // durations from the end of the Playlist file.  Doing so can trigger
    // playback stalls."
    // https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-6.3.3

    liveEdgePadding = typeof liveEdgePadding === 'number' ? liveEdgePadding : playlist.targetDuration * 3;
```

src/sync-controller.js
```diff
@@ -38,7 +44,8 @@ export const syncPointStrategies = [
       currentTime = currentTime || 0;
 
       for (let i = 0; i < segments.length; i++) {
-        const segment = segments[i];
+        const segmentIndex = getSegmentIndex(i, playlist, currentTime);
```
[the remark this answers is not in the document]
brandonocasey: I don't think it's that much code and we share this bit between the two loops. I think that we will end up with more code if we change this to looping backwards.

gkatsev:
I have to say I'm a bit leery of removing the `safeLiveIndex`, though, it's probably fine. Would like to hear back from @gesinger if he has more thoughts on the comment he brought up.

brandonocasey:
So the reason I removed safeLiveIndex is because we only used it to get the index of a segment based on a period of time that we want to be behind. Then we use that index to calculate the duration at which that segment would start, which will put `seekableEnd` at an estimated segment boundary. If that estimated segment boundary is wrong we will create a gap. 
Example:
* playlist is `3000s` 
* targetDuration is `6` 
* last segment durations from the manifest are `-1 = 5.597`, `-2 = 5.5`, `-3 = 5.9`, -4 = 5.5`.
1. safeLiveDelay will be `17.9s` behind live (targetDuration * 2, plus last segment duration)
2. safeLiveIndex will return segment at index `-4`
3. intervalDuration will return `2977.503` 
4. this will be a nearly exact segment boundary for the start to segment at index `-4`

With this code  just start at the specific distance away from the current end of the playlist that we want to be. 
Example:
* playlist is `3000s` 
* targetDuration is `6` 
* last segment durations from the manifest are `-1 = 5.597`, `-2 = 5.5`, `-3 = 5.9`, -4 = 5.5`.

1. We get the duration of the entire playlist, `3000s`
2. We subtract our safeLiveDistance of `18s` (targetDuration * 3) to get `2982s`
3. This will be most of the way through the segment at index `-4` but not on a boundary.

test/playlist.test.js
```diff
@@ -6,1177 +6,1236 @@ import { useFakeEnvironment } from './test-helpers';
...
-    segmentDuration,
-    bandwidth,
-    playlist,
-    bytesReceived
-  );
+    const playlistEnd = Playlist.playlistEnd(playlist, 0, true, 25);
 
-  assert.ok(isNaN(estimate), 'returns NaN when no BANDWIDTH information on playlist');
+    assert.equal(playlistEnd, 175, 'playlist end is 175');
+  });
```
gesinger, line 724: Looks like a copy of the previous test.

test/playlist.test.js
```diff
@@ -6,1427 +6,1236 @@ import { useFakeEnvironment } from './test-helpers';
...
-  assert.ok(
-    Playlist.hasAttribute('BANDWIDTH', playlist),
-    'true for playlist with specified attribute'
-  );
-});
+    assert.equal(playlistEnd, 175, 'playlist end is 175');
+  });
 
-QUnit.module('Playlist estimateSegmentRequestTime');
+  QUnit.test('playlistEnd uses given live edge padding with useSafeLiveEnd true', function(assert) {
```
gesinger, line 707: Looks like this test is included twice.

gesinger (approved):

PR #1106, opened 2021-03-26
feat: stats for timeToLoadedData, appendsToLoadedData, mainAppendsToLoadedData, audioAppendsToLoadedData, and mediaAppends

brandonocasey:
## Description
Add the mentioned stats


src/master-playlist-controller.js
```diff
@@ -289,6 +291,23 @@ export class MasterPlaylistController extends videojs.EventTarget {
     } else {
       this.masterPlaylistLoader_.load();
     }
+    const timeToCanPlayStart = Date.now();
+
+    this.timeToFirstFrame__ = 0;
+    this.appendsToFirstFrame__ = 0;
+
+    this.tech_.one('canplay', () => {
```
brandonocasey, line 299: once we 'canplay' a frame will be showing.

gkatsev: Apparently this is close to a "time to first byte" since if we are paused, the video may not get rendered immediately. `playing` or first `timeupdate` is potentially closer to a time to first rendered frame if we are autoplaying/not paused.

gkatsev: There's this experimental API that can better represent the frame getting rendered https://wicg.github.io/video-rvfc/

brandonocasey: switched it to `loadeddata` https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadeddata_event
as it specifically says: "The loadeddata event is fired when the frame at the current playback position of the media has finished loading; often the first frame."

gkatsev: that seems more like time to first byte rather than time to first frame, the latter implies that the frame has rendered which loadeddata doesn't guarantee it.

gkatsev: Yup, basically, this isn't a simple change. We should figure out what we care about and do our best to capture that.

Preload value also plays into account here.

Also, I believe artificial timeupdates we fire only fire on the player and not the tech.

brandonocasey: I switched this over to using timeupdate and took into account preload.

gkatsev: I guess what do you consider `timeToFirstFrame`? Is it the data has been loaded into the video element or the frame has rendered visibly to the user?

brandonocasey: Since we don't have a good way to determine if the frame has been rendered visibly to the user (except for the experimental API), I think time to first frame should be when the video element reports that it has a `buffered` region containing `currentTime`. That might be a few ms before the frame is visible to the user, but I think it's the best cross browser solution.

gkatsev: Makes sense. I would probably want to rename the field then. We should also make sure to document exactly what it means so that we don't get confused in the future and have a place to reference it.
Also, does live have anything special for this beyond being equivalent to preload=none?

brandonocasey: maybe `timeToLoadedData` and `appendsToLoadedData`? Live shouldn't be any different for VOD for these stats.

src/segment-loader.js
```diff
@@ -2968,6 +2969,8 @@ export default class SegmentLoader extends videojs.EventTarget {
     // used for testing
     this.trigger('appended');
 
+    this.mediaAppends++;
```
gesinger, line 3008: This will show extra appends for:
* segments with no data to be appended
* sync requests that did not get appended

It will also under-report total number of appends, since we potentially append multiple times per segment.

Do we want to consider those cases and think about naming as "segmentAppends" since that may denote full segments?

gkatsev (approved):

src/segment-loader.js
brandonocasey: sync requests that do not get appended will return much earlier in the function. We could wrap `mediaAppends` increment in `segmentInfo.hasAppendedData`?

brandonocasey: the other problem with calling it segment appends is that we now append parts too.

gesinger: 👍 on sync requests

On wrapping and part appends, what is our use-case for this data? Do we want to know actual appends, or are we more concerned about requests?

brandonocasey: I think we want to know actual appends, we already have stats for requests.

gesinger (approved):

PR #1117, opened 2021-04-09
refactor: offload mp4/ts probe to web worker

brandonocasey:
## Description
offloads ts / mp4 probing to the web worker saving us ~600 gzipped bytes.


test/segment-loader.test.js
```diff
@@ -911,8 +936,7 @@ QUnit.module('SegmentLoader', function(hooks) {
           this.clock.tick(1);
 
           // decryption tick for syncWorker
-          this.clock.tick(1);
-
+          this.clock.tick(2);
```
gesinger, line 939: Might be worth adding a comment for the extra tick

src/util/worker-callback.js
```diff
@@ -0,0 +1,42 @@
...
+  const callback = options.callback;
+  const message = Object.assign({}, options, {endAction: null, transmuxer: null, callback: null});
+
+  const listenForEndEvent = (event) => {
+    if (event.data.action !== endAction) {
+      return;
+    }
+    transmuxer.removeEventListener('message', listenForEndEvent);
+
+    // transfer ownership of bytes back to us.
```
gesinger, line 13: Since the ownership is transferred here, do we need to transfer the ownership again in the media-segment-request users of `workerCallback`?

brandonocasey: We do for anywhere else the data is referenced.

src/media-segment-request.js
```diff
@@ -379,6 +360,47 @@ const transmuxAndNotify = ({
...
     }
   });
+
+  // Check to see if we are appending a full segment.
+  if (!isPartial && !segment.lastReachedChar) {
+    // In the full segment transmuxer, we don't yet have the ability to extract a "proper"
+    // start time. Meaning cached frame data may corrupt our notion of where this segment
+    // really starts. To get around this, full segment appends should probe for the info
+    // needed.
+    workerCallback({
```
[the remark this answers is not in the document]
brandonocasey: We don't send anything to the transmuxer until the probe comes back, so I wouldn't think so.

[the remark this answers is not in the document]
gesinger: Missed that, you're right 👍

gesinger (approved):

PR #1122, opened 2021-04-27
feat: add initialBandwidth option at the tech level

gkatsev:


brandonocasey (approved):

brandonocasey:
I wonder if we should deprecate `bandwidth` during the next major

test/videojs-http-streaming.test.js
```diff
@@ -2913,6 +2913,36 @@ QUnit.test(
...
+
+  this.player.src({
+    src: 'http://example.com/media.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+
+  this.clock.tick(1);
+
+  openMediaSource(this.player, this.clock);
+  assert.equal(this.player.tech_.vhs.bandwidth, 0, 'set bandwidth to 0');
```
gesinger, line 2928: Kinda minor, but here and below, the test values being only 0 may run into issues if bandwidth is ever initialized in the future.

src/videojs-http-streaming.js
```diff
@@ -511,6 +511,12 @@ class VhsHandler extends Component {
       videojs.log.warn('Using hls options is deprecated. Use vhs instead.');
     }
 
+    // if a tech level `initialBandwidth` option was passed
```
gesinger, line 514: Probably worth documentating and mentioning that it takes precedence over VHS level `bandwidth` but not source level `bandwidth`.

PR #1124, opened 2021-04-29
fix: add master referenced id/uri audio playlists. Add playlists to hls media groups

brandonocasey:
Fixes issues with rendition switching on audio only hls media.

### DIFF
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -275,9 +275,9 @@ export const addPropertiesToMaster = (master, uri) => {
       // we mark the first playlist uri without prepending the index as we used to
       // ideally we would do all of the playlists the same way.
       if (i === 0) {
-        p.uri = p.uri || id;
-      } else {
         p.uri = p.uri || groupId;
+      } else {
+        p.uri = p.uri || id;
       }
 
       p.id = p.id || id;
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -270,7 +270,16 @@ export const addPropertiesToMaster = (master, uri) => {
     properties.playlists.forEach(function(p, i) {
       const id = createPlaylistID(i, groupId);
 
-      p.uri = p.uri || id;
+      // DEPRECATED, this has been added to prevent a breaking change.
+      // previously we only ever had a single media group playlist, so
+      // we mark the first playlist uri without prepending the index as we used to
+      // ideally we would do all of the playlists the same way.
+      if (i === 0) {
+        p.uri = p.uri || id;
+      } else {
+        p.uri = p.uri || groupId;
+      }
+
       p.id = p.id || id;
 
       p.resolvedUri = p.resolvedUri || resolveUrl(master.uri, p.uri);
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -270,19 +270,22 @@ export const addPropertiesToMaster = (master, uri) => {
     properties.playlists.forEach(function(p, i) {
       const id = createPlaylistID(i, groupId);
 
-      // DEPRECATED, this has been added to prevent a breaking change.
-      // previously we only ever had a single media group playlist, so
-      // we mark the first playlist uri without prepending the index as we used to
-      // ideally we would do all of the playlists the same way.
-      if (i === 0) {
-        p.uri = p.uri || groupId;
+      if (p.uri) {
+        p.resolvedUri = p.resolvedUri || resolveUrl(master.uri, p.uri);
       } else {
-        p.uri = p.uri || id;
+        // DEPRECATED, this has been added to prevent a breaking change.
+        // previously we only ever had a single media group playlist, so
+        // we mark the first playlist uri without prepending the index as we used to
+        // ideally we would do all of the playlists the same way.
+        p.uri = i === 0 ? groupId : id;
+
+        // don't resolve a placeholder uri to an absolute url, just use
+        // the placeholder again
+        p.resolvedUri = p.uri;
       }
 
       p.id = p.id || id;
 
-      p.resolvedUri = p.resolvedUri || resolveUrl(master.uri, p.uri);
       // add an empty attributes object, all playlists are
       // expected to have this.
       p.attributes = p.attributes || {};
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -105,7 +105,13 @@ export const parseManifest = ({
  *        Callback to call for each media group
  */
 export const forEachMediaGroup = (master, callback) => {
+  if (!master.mediaGroups) {
+    return;
+  }
   ['AUDIO', 'SUBTITLES'].forEach((mediaType) => {
+    if (!master.mediaGroups[mediaType]) {
+      return;
+    }
     for (const groupKey in master.mediaGroups[mediaType]) {
       for (const labelKey in master.mediaGroups[mediaType][groupKey]) {
         const mediaProperties = master.mediaGroups[mediaType][groupKey][labelKey];
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -261,22 +261,28 @@ export const addPropertiesToMaster = (master, uri) => {
   }
 
   forEachMediaGroup(master, (properties, mediaType, groupKey, labelKey) => {
-    if (!properties.playlists ||
-        !properties.playlists.length ||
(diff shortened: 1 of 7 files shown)


src/playlist-loader.js
```diff
@@ -429,13 +429,16 @@ export default class PlaylistLoader extends EventTarget {
...
 
     // switch to fully loaded playlists immediately
-    if (this.master.playlists[playlist.id].endList ||
+    if (masterPlaylistRef && masterPlaylistRef.endList ||
         // handle the case of a playlist object (e.g., if using vhs-json with a resolved
         // media playlist or, for the case of demuxed audio, a resolved audio media group)
         (playlist.endList && playlist.segments.length)) {
       // abort outstanding playlist requests
+
+      updateMaster(this.master, playlist);
```
brandonocasey, line 441: Make sure that the master playlistRef and the playlist are up to date.

src/media-groups.js
```diff
@@ -662,6 +662,20 @@ export const initialize = {
...
 };
 
+const groupMatch = (list, media) => {
+  for (let i = 0; i < list.length; i++) {
+    if (playlistMatch(media, list[i])) {
+      return true;
+    }
+
+    if (list[i].playlists && groupMatch(list[i].playlists, media)) {
+      return true;
```
gkatsev, line 672: should this be something like the following instead?
```suggestion
    if (list[i].playlists) {
      return groupMatch(list[i].playlists, media);
```

since, if we have a `playlists`, groupMatch is whatever recursing is.

src/manifest.js
```diff
@@ -261,22 +261,28 @@ export const addPropertiesToMaster = (master, uri) => {
...
     }
 
-    // Set up phony URIs for the media group playlists since playlists are referenced by
-    // their URIs throughout VHS, but some formats (e.g., DASH) don't have external URIs
-    const phonyUri = `placeholder-uri-${mediaType}-${groupKey}-${labelKey}`;
-    const id = createPlaylistID(0, phonyUri);
+    properties.playlists.forEach(function(p, i) {
+      const id = createPlaylistID(i, groupId);
+
+      p.uri = p.uri || id;
```
gkatsev, line 273: is `p.uri` expected to be the `groupId` (what used to be called `phonyUri`)?
Should `uri` continue to be set to `groupId`?

src/media-groups.js
brandonocasey: I don't think so because we want to consider other playlists as well. If we return the value of `groupMatch` here it might be false, while the next playlist in the list matches.

gkatsev: ah, we don't want to return false here yet. I sort of missed the for loop. Cool, makes sense.

src/manifest.js
brandonocasey: Now that we can have more than 1 playlist in a group the groupId alone would conflict

gkatsev: Wonder how big of an issue this is, since, it's potentially breaking.

brandonocasey: I guess we could keep playlist at index 0 as `groupId`, and use all others as `id`

src/manifest.js
```diff
@@ -261,22 +261,37 @@ export const addPropertiesToMaster = (master, uri) => {
...
+      // ideally we would do all of the playlists the same way.
+      if (i === 0) {
+        p.uri = p.uri || groupId;
+      } else {
+        p.uri = p.uri || id;
+      }
+
+      p.id = p.id || id;
+
+      p.resolvedUri = p.resolvedUri || resolveUrl(master.uri, p.uri);
```
gesinger, line 285: If the playlist doesn't have a URI, would this lead to a weird result (given that it would use the ID)?

src/manifest.js
```diff
@@ -261,22 +261,37 @@ export const addPropertiesToMaster = (master, uri) => {
...
   forEachMediaGroup(master, (properties, mediaType, groupKey, labelKey) => {
-    if (!properties.playlists ||
-        !properties.playlists.length ||
-        properties.playlists[0].uri) {
-      return;
+    const groupId = `placeholder-uri-${mediaType}-${groupKey}-${labelKey}`;
+
+    if (!properties.playlists || !properties.playlists.length) {
+      properties.playlists = [Object.assign({}, properties)];
     }
```
gesinger, line 268: This may impact this check: https://github.com/videojs/http-streaming/blob/552b012035b6f2fd8da8440a41928668d85e25a8/src/media-groups.js#L471

Which looks for the existence of playlists.

test/media-groups.test.js
```diff
@@ -320,6 +320,76 @@ QUnit.module('MediaGroups', function() {
...
+          }
+        };
+
+        this.groups.main = [{ id: 'en', uri: 'en.ts'}, { id: 'fr', uri: 'fr.ts' }];
+        this.groups.foo = [{ id: 'en-foo', uri: 'en-foo.ts' }, { id: 'fr-foo', uri: 'fr-foo.ts' }];
+        this.groups.bar = [{ id: 'en-bar', uri: 'en-foo.ts' }, { id: 'fr-bar', uri: 'fr-bar.ts' }];
+
+        assert.deepEqual(this.activeGroup(), this.groups.bar, 'selected matching group');
+      });
+      QUnit.test('dash audio only playlist returns correct group', function(assert) {
```
gesinger, line 342:
```suggestion

      QUnit.test('dash audio only playlist returns correct group', function(assert) {
```

src/playlist-loader.js
```diff
@@ -429,13 +429,22 @@ export default class PlaylistLoader extends EventTarget {
...
 
     // switch to fully loaded playlists immediately
-    if (this.master.playlists[playlist.id].endList ||
+    if (masterPlaylistRef && masterPlaylistRef.endList ||
         // handle the case of a playlist object (e.g., if using vhs-json with a resolved
         // media playlist or, for the case of demuxed audio, a resolved audio media group)
         (playlist.endList && playlist.segments.length)) {
       // abort outstanding playlist requests
+
+      const update = updateMaster(this.master, playlist);
```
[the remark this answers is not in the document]
brandonocasey: To update uri references for media playlists. We can get into a state where they are behind with audio only playlists with altenative audio tracks.

src/manifest.js
brandonocasey: It will give us weird results, but there are a few places in hls that expect a resolved uri so our media playlist needs to have one.

src/manifest.js
brandonocasey: I think this might be okay since it is looking for dash sourceType specifically here and those will always have playlists. It does change our notions around a bit though.

src/manifest.js
gesinger: Should we just use the ID as-is in those cases? Might be easier to understand as a property if it's not an actual URI (that doesn't exist).

src/playlist-loader.js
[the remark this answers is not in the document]
gesinger: Is there a better place we can put that logic? Maybe on refreshes we make sure to update everything? It would help prevent the `media` function from having more potential side effects.

src/manifest.js
gesinger: If DASH always has `playlists` in this case, I wonder why I included that case at all back when.

src/playlist-loader.js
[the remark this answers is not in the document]
brandonocasey: good point, let me see if there is a better spot.

src/manifest.js
brandonocasey: maybe it didn't at the time this was added

brandonocasey: Moved the code to `updateMaster` where we update the media group playlist with changes from the requested playlist (usually just segments)

gesinger (commented):
Looks good. Might be worth adding tests around `addPropertiesToMaster` for media group IDs.

src/playlist-loader.js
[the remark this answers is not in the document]
brandonocasey: Moved the code to updateMaster where we update the media group playlist with changes from the requested playlist (usually just segments)

gesinger (approved):

PR #1125, opened 2021-04-30
fix: llhls syncing fixes

brandonocasey:
* Adds partIndex to `getMediaInfoForTime`, `sync-controller` and other Playlist functions.
* Adds part level timing info and fixes segment level timing info with parts


src/playlist.js
```diff
@@ -261,12 +261,13 @@ export const duration = function(playlist, endSequence, expired) {
...
-  * @param {number} startIndex
-  * @param {number} endIndex
+  * @param {Object} durationList list to iterate over for durations.
+  * @param {number} defaultDuration duration to use for elements before or after the durationList
+  * @param {number} startIndex partsAndSegments index to start
+  * @param {number} endIndex partsAndSegments index to end.
   * @return {number} the number of seconds between startIndex and endIndex
   */
-export const sumDurations = function(playlist, startIndex, endIndex) {
+export const sumDurations = function({defaultDuration, durationList, startIndex, endIndex}) {
```
brandonocasey, line 270: changes to allow passing in a list and a default duration. This allows us to use `partsAndSegments`, just `parts`, or just `segments`

src/playlist.js
```diff
@@ -261,12 +261,13 @@ export const duration = function(playlist, endSequence, expired) {
   * playlist in which case, the targetDuration of the playlist is used
   * to approximate the durations of the segments
   *
-  * @param {Object} playlist a media playlist object
-  * @param {number} startIndex
-  * @param {number} endIndex
+  * @param {Object} durationList list to iterate over for durations.
+  * @param {number} defaultDuration duration to use for elements before or after the durationList
+  * @param {number} startIndex partsAndSegments index to start
+  * @param {number} endIndex partsAndSegments index to end.
```
gkatsev, line 267: Suggestion: since this changed to an options object, we should rename the params to reflect that
```suggestion
  * @param {Object} options.durationList list to iterate over for durations.
  * @param {number} options.defaultDuration duration to use for elements before or after the durationList
  * @param {number} options.startIndex partsAndSegments index to start
  * @param {number} options.endIndex partsAndSegments index to end.
```

src/playlist.js
```diff
@@ -374,30 +375,54 @@ export const seekable = function(playlist, expired, liveEdgePadding) {
  * @param {number} startTime
  * @return {Object}
  */
-export const getMediaInfoForTime = function(
+export const getMediaInfoForTime = function({
```
gkatsev, line 378: suggestion: the jsdoc for this method should be updated as well to reflect the new options object.
```
* @param {Object} options.playlist the media playlist to query
* @param {number} options.currentTime The number of seconds since the earliest
* possible position to determine the containing segment for
* @param {number} options.startIndex
* @param {number} options.startTime
* @param {number} options.segmentIndex
* @param {number} options.partIndex
* ```

File: src/playlist.js [hunk 4]
```diff
@@ -374,30 +375,54 @@ export const seekable = function(playlist, expired, liveEdgePadding) {
...
+
+  for (let i = 0; i < partsAndSegments.length; i++) {
+    const partAndSegment = partsAndSegments[i];
+
+    if (segmentIndex !== partAndSegment.segmentIndex) {
+      continue;
+    }
+
+    // skip this if part index does not match.
+    if (typeof partIndex === 'number' && typeof partAndSegment.partIndex === 'number' && partIndex !== partAndSegment.partIndex) {
```
gkatsev, line 399: question: why would `partIndex` not be a number? Is it if it isn't present?

gkatsev: I guess below I can see that it is default to `null`. Maybe we should default it to `-1` instead, that way we would only need to check `partIndex !== partAndSegment.partIndex`?

src/playlist.js
gkatsev: I guess right now `partIndex should be `{?number}`

brandonocasey: I think it's `{number} [partIndex]` to denote optional.

src/playlist.js
brandonocasey: We ran into problems with mediaIndex being negative previously during live playlist updates. I wanted to avoid the whole issue with partIndex. The issue was something like: If we get a playlist update at just the right time we can end up downloading the same segment again because `-1 + 1 = 0`. Using null we prevent choosing an incorrect segment via incrementing.

src/playlist.js
gkatsev: yeah, `[partIndex]` to be option, but given that it can be `null` or a number, it should have the `?`.

gkatsev (approved):
Aside from my nitpick above, LGTM

test/sync-controller.test.js
```diff
@@ -226,6 +229,50 @@ QUnit.test('uses separate date time to display time mapping for each timeline',
...
+
+  assert.equal(syncPoint, null, 'no syncpoint when no date time to display time mapping');
+
+  playlist.segments.forEach((segment, index) => {
+    segment.dateTimeObject = new Date(2012, 11, 12, 12, 12, 12 + (index * 10));
+  });
+
+  this.syncController.setDateTimeMappingForStart(playlist);
+
+  const newPlaylist = playlistWithDuration(200, {llhls: true});
```
gesinger, line 249: Any reason the new playlist is created versus using the old?

test/sync-controller.test.js
```diff
@@ -226,6 +229,50 @@ QUnit.test('uses separate date time to display time mapping for each timeline',
...
+
+  syncPoint = strategy.run(this.syncController, newPlaylist, duration, timeline);
+
+  assert.equal(syncPoint, null, 'no syncpoint when datetimeObject not set on playlist');
```
gesinger, line 253: Should this be when no dateTimeObjects for segments?

src/sync-controller.js
```diff
@@ -473,15 +491,26 @@ export default class SyncController extends videojs.EventTarget {
...
     } else {
       return false;
     }
 
+    if (part) {
+      part.start = start;
+      part.end = end;
+    }
+
+    if (!segment.start) {
```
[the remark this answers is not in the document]
brandonocasey: because for parts that are associated with a segment, we don't want to use the part start value after the first part is appended.

src/sync-controller.js
```diff
@@ -37,15 +32,15 @@ export const syncPointStrategies = [
...
       let lastDistance = null;
+      const partsAndSegments = getPartsAndSegments(playlist);
 
       currentTime = currentTime || 0;
-
-      for (let i = 0; i < segments.length; i++) {
-        const segmentIndex = getSegmentIndex(i, playlist, currentTime);
-        const segment = segments[segmentIndex];
+      for (let i = 0; i < partsAndSegments.length; i++) {
+        const index = (playlist.endList || currentTime === 0) ? i : partsAndSegments.length - (i + 1);
```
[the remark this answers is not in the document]
brandonocasey: added

src/playlist.js
```diff
@@ -367,37 +368,63 @@ export const seekable = function(playlist, expired, liveEdgePadding) {
...
- * @param {Object} playlist the media playlist to query
- * @param {number} currentTime The number of seconds since the earliest
+ * @param {Object} options.playlist the media playlist to query
+ * @param {number} options.currentTime The number of seconds since the earliest
  * possible position to determine the containing segment for
- * @param {number} startIndex
- * @param {number} startTime
- * @return {Object}
+ * @param {number} options.startTime the time where the segment/part starts
+ * @param {number} options.segmentIndex the segment index to start looking at.
```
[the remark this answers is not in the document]
brandonocasey: changed

test/sync-controller.test.js
brandonocasey: not entirely sure, I updated this test to just use the old playlist.

src/sync-controller.js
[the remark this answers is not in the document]
gesinger: 👍 Maybe worth a comment (since I missed it when reading through)

test/sync-controller.test.js
```diff
@@ -246,30 +246,20 @@ QUnit.test('ProgramDateTime strategy finds nearest llhls sync point', function(a
...
-    segment.dateTimeObject = new Date(2012, 11, 12, 12, 12, 22 + (index * 10));
-  });
-
-  syncPoint = strategy.run(this.syncController, newPlaylist, duration, timeline, 194);
+  syncPoint = strategy.run(this.syncController, playlist, duration, timeline, 194);
 
   assert.deepEqual(syncPoint, {
     time: 192,
-    segmentIndex: 18,
+    segmentIndex: 19,
```
gesinger, line 253: Do you know why this and below changed even though the playlists look like they should be the same?

brandonocasey: This happened because newPlaylist had a `22 + (index * 10)` instead of a `12 + (index * 10)` for its `dateTimeObject`. I think that using a weird date like that makes it much harder to understand what the test is doing. The way it's setup now its easy to see why the sync points are where they are.

src/sync-controller.js
```diff
@@ -510,7 +510,12 @@ export default class SyncController extends videojs.EventTarget {
       part.end = end;
     }
 
-    if (!segment.start) {
+    // if we don't have a segment start yet or the start value we got
+    // is less than our current segment.start value. save a new start value.
```
gesinger, line 514:
```suggestion
    // If we don't have a segment start yet or the start value we got
    // is less than our current segment.start value, save a new start value.
```

src/sync-controller.js
```diff
@@ -510,7 +510,12 @@ export default class SyncController extends videojs.EventTarget {
...
+    // We have to do this because parts will have segment timining info saved
```
gesinger, line 515:
```suggestion
    // We have to do this because parts will have segment timing info saved
```

gesinger (approved):

gesinger (approved):

PR #1127, opened 2021-05-05
fix: more accurate segment choices and logging

brandonocasey:
## Description
This pull request refactors two functions that are the cause of latency drift in live streams during rendition changes. It also vastly improves the segmentInfoString logging logic.

### `SegmentLoader#getSyncSegmentCandidate`
What we do right now is choose a segment that will never be the one we want, still append it to the source buffer and then ignore the final "end result" (saving the media/part index as the current one).

I have improved this logic in the following ways:
1. Try to choose an accurate segment that will be appended
2. If the segment isn't accurate, save the timing info but don't append saving us a lot of source buffer latency.
3. If the segment is accurate append it and save it's media/part index as the current media/part index.

### `Playlist#getMediaInfoForTime`
Currently we add `TIME_FUDGE_FACTOR` for each segment and this causes us to choose the wrong segment more often then not for playlists with > 15 segments as we will be off by more than half a second in those cases. Instead we should only add `TIME_FUDGE_FACTOR` a single time when comparing against the time we want. We might even want to consider not using `TIME_FUDGE_FACTOR` here.


src/playlist.js
```diff
@@ -374,9 +374,9 @@ export const getMediaInfoForTime = function(
       for (let i = startIndex - 1; i >= 0; i--) {
         const segment = partsAndSegments[i];
 
-        time += (segment.duration + TIME_FUDGE_FACTOR);
+        time += segment.duration;
 
-        if (time > 0) {
+        if ((time + TIME_FUDGE_FACTOR) > 0) {
```
brandonocasey, line 379: Should we keep `TIME_FUDGE_FACTOR` at all?

test/loader-common.js
```diff
@@ -870,7 +870,7 @@ export const LoaderCommonFactory = ({
       });
 
       QUnit.test('drops partIndex if playlist update drops parts', function(assert) {
-        assert.timeout(100000000000000000000);
+        loader.duration_ = () => Infinity;
```
brandonocasey, line 873: ie 11 has the duration as NaN here, so sync-controller gets a sync point of `1` for a vod video.

src/playlist.js
gkatsev: question: What is `TIME_FUDGE_FACTOR` used for here? Is using it causing timing issues down the line?

src/segment-loader.js
```diff
@@ -34,27 +34,27 @@ import {lastBufferedEnd} from './ranges.js';
...
-    return segmentIndexArray[Math.min(segmentIndexArray.length - 1, 1)];
+      if (time > targetTime) {
+        return i;
+      }
+    }
   }
 
-  return Math.max(segments.length - 1, 0);
+  // default to grabbing the last timeline segment or zero.
+  return Math.max(0, timelineSegments[timelineSegments.length - 1] || 0);
```
gkatsev, line 57: question: so, because of how `timelineSegments` are generated (based on whether they're in the current timeline or not), when we get here we either have timelineSegments or we have an empty list and so we end up returning `0` as a default like we did previously when `currentTimeline === -1`?

src/segment-loader.js
```diff
@@ -2031,6 +2034,27 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      });
+
+      const next = this.chooseNextRequest_();
+
+      // if the sync request isn't the segment that would be request next
+      // after taking into account it's data. do not append it.
+      if (next.mediaIndex !== segmentInfo.mediaIndex || next.partIndex !== segmentInfo.partIndex) {
+        this.logger_('sync segment was incorrect, not appending');
+        return;
+      }
```
gkatsev, line 2053: question: this was causing us issues because we got a sync request for a location that was different from where we wanted but previously we were appending the sync segment but this caused problems?

test/segment-loader.test.js
```diff
@@ -2967,6 +2954,65 @@ QUnit.module('SegmentLoader', function(hooks) {
       });
     });
 
+    QUnit.test('sync request can be thrown away', function(assert) {
```
gkatsev, line 2957: issue: looks like this test is failing under IE11. https://github.com/videojs/http-streaming/pull/1127/checks?check_run_id=2659611689#step:13:478

gkatsev (commented):
Trying out https://conventionalcomments.org/ again.

test/segment-loader.test.js
```diff
@@ -79,51 +79,38 @@ SegmentLoader.prototype.addSegmentMetadataCue_ = function() {};
...
+
+  playlist.segments = [
+    {timeline: 0, duration: 4},
+    {timeline: 0, duration: 4},
+    {timeline: 0, duration: 4},
+    {timeline: 0, duration: 4}
+  ];
+
+  assert.equal(getSyncSegmentCandidate(-1, playlist, 0), 0, '-1 timeline, 4x 0 segments, 0 target');
+  assert.equal(getSyncSegmentCandidate(0, playlist, 1), 0, '0 timeline, 4x 0 segments, 4 target');
```
gesinger, line 95: Should this be `1 target`?

test/segment-loader.test.js
```diff
@@ -79,51 +79,38 @@ SegmentLoader.prototype.addSegmentMetadataCue_ = function() {};
...
+  assert.equal(getSyncSegmentCandidate(0, playlist, 4), 1, '0 timeline, 4x 0 segments, 4 target');
+  assert.equal(getSyncSegmentCandidate(-1, playlist, 8), 0, '-1 timeline, 4x 0 segments, 8 target');
+  assert.equal(getSyncSegmentCandidate(0, playlist, 8), 2, '0 timeline, 4x 0 segments, 8 target');
+  assert.equal(getSyncSegmentCandidate(0, playlist, 20), 3, '0 timeline, 4x 0 segments, 20 target');
+
+  playlist.segments = [
```
gesinger, line 101: I don't think this would be a valid segments array. Might be able to skip this test.

src/playlist.js
brandonocasey: I think this change basically encompasses what you are saying. We only use `getMediaIndexForTime` when we are not incrementing mediaIndex/partIndex. This change makes it so that cumulatively we can only be off by `TIME_FUDGE_FACTOR` rather than every single part/segments duration being incremented or decremented by `TIME_FUDGE_FACTOR`. For the sake of time I think that we might want to add a TODO here to see if we should even use TIME_FUDGE_FACTOR at all. I lean towards not using it, but it will require a bit more testing and could easily enough be an isolated change.

src/segment-loader.js
brandonocasey: yes thats correct.

src/segment-loader.js
brandonocasey: We download an entire segment for sync purposes. The segment used to always be inaccurate but appended anyway costing us a lot of time and back/forward buffer space. Now we only append if the segment would have been the next segment we choose anyway. Otherwise we save the timing info and throw away the segment than choose the real next segment.

test/segment-loader.test.js
brandonocasey: I think it might still be a valid test for this code in isolation.

test/segment-loader.test.js
brandonocasey: yes

src/segment-loader.js
```diff
@@ -2031,6 +2034,27 @@ export default class SegmentLoader extends videojs.EventTarget {
     // as we use the start of the segment to offset the best guess (playlist provided)
     // timestamp offset.
     this.updateSourceBufferTimestampOffset_(segmentInfo);
+
+    // throw away the isSyncRequest segment if
+    // it wouldn't have been the next segment we request.
+    if (segmentInfo.isSyncRequest) {
+      this.syncController_.saveSegmentTimingInfo({
```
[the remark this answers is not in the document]
brandonocasey: Should we call this function within this block? From a brief glance it seems like it only estimates the segment end point using start/duration unless we have an `end`. Seems like we should call it here.

src/segment-loader.js
```diff
@@ -34,27 +34,27 @@ import {lastBufferedEnd} from './ranges.js';
  * @param {Object} playlist - the playlist object to look for a
  * @return {number} An index of a segment from the playlist to load
  */
-export const getSyncSegmentCandidate = function(currentTimeline, {segments = []} = {}) {
-  // if we don't currently have a real timeline yet.
-  if (currentTimeline === -1) {
-    return 0;
-  }
+export const getSyncSegmentCandidate = function(currentTimeline, playlist, targetTime) {
+  const segments = playlist.segments || [];
```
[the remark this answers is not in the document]
brandonocasey: done

src/playlist.js
gesinger: 👍 a TODO would be good for now. This definitely is better than before.

src/segment-loader.js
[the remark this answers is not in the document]
gesinger: I think we should probably call it. On that note we may also want to consider adding a property to the segment whether the last time we requested that segment it was a sync request (and clear it if it's requested on a non sync request). It may help us for debugging, particularly now that it's no longer being appended.

src/playlist.js
gkatsev: Probably worth making an issue/a note in our major/refactoring epic

src/segment-loader.js
gkatsev: Makes sense. Maybe one day we'll be able to use `fetch` and only download enough of that segment to get timing info rather than waiting for the entire segment.

src/playlist.js
brandonocasey: added

src/segment-loader.js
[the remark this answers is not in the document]
brandonocasey: called

src/playlist.js
```diff
@@ -376,6 +376,9 @@ export const getMediaInfoForTime = function(
 
         time += segment.duration;
 
+        // TODO: We should consider not using TIME_FUDGE_FACTOR at all here, but at
+        // the time where this change was made there wasn't enough time to test without it.
+        // We should make some time to see if we can remove TIME_FUDGE_FACTOR here.
```
gesinger, line 381:
```suggestion
        // TODO: consider not using TIME_FUDGE_FACTOR at all here
```

gesinger (dismissed):

src/segment-loader.js
```diff
@@ -162,14 +162,10 @@ const segmentInfoString = (segmentInfo) => {
...
-  const partCount = segment.parts ? segment.parts.length : 0;
-  const preloadPartCount = segment.preloadHints ?
-    segment.preloadHints.filter((h) => h.type === 'PART').length :
-    0;
-  const zeroBasedPartCount = partCount + preloadPartCount - 1 - (preloadPartCount > 0 ? 1 : 0);
+  const totalParts = hasPartIndex ? getKnownPartCount({preloadSegment: segment}) - 1 : 0;
 
   return `${name} [${seq + index}/${seq + segmentLen}]` +
-    (hasPartIndex ? ` part [${partIndex}/${zeroBasedPartCount}]` : '') +
+    (hasPartIndex ? ` part [${partIndex}/${totalParts}]` : '') +
```
gesinger, line 168: Should this take into account the 0 based indexing vs total. Maybe we can just say `[index 0, 3 total]

brandonocasey: I think we should keep the name `zeroBasedPartCount` here I guess. I do subtract 1 to make it zero based.

gesinger (approved):

PR #1128, opened 2021-05-11
fix: only append init segment on change

brandonocasey:
## Description
Only append init segments on change.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -2367,8 +2367,11 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     this.logger_(`Requesting ${segmentInfoString(segmentInfo)}`);
 
-    // if init segment is being requested
-    // we will have to re-append the init segment.
+    // If there's an init segment associated with this segment, but it is not cached (identified by a lack of bytes),
+    // then this init segment has never been seen before and should be appended.
+    //
+    // At this point the content type (audio/video or both) is not yet known, but it should be safe to set
+    // both to true and leave the decision of whether to append the init segment to append time.
     if (simpleSegment.map && !simpleSegment.map.bytes) {
       this.logger_('going to request init segment.');
       this.appendInitSegment_ = {
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -2098,7 +2098,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       this.playlistOfLastInitSegment_[type] = playlist;
       // we should only be appending the next init segment if we detect a change, or if
       // the segment has a map
-      this.appendInitSegment_[type] = map ? true : false;
+      this.appendInitSegment_[type] = false;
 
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -2098,7 +2098,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       this.playlistOfLastInitSegment_[type] = playlist;
       // we should only be appending the next init segment if we detect a change, or if
       // the segment has a map
-      this.appendInitSegment_[type] = map ? true : false;
+      this.appendInitSegment_[type] = false;
```
gesinger, line 2101: Maybe we should update the comment to something like: "disable future init segments appends for this type (until a change is necessary)"

src/segment-loader.js
```diff
@@ -2368,6 +2368,16 @@ export default class SegmentLoader extends videojs.EventTarget {
...
     this.logger_(`Requesting ${segmentInfoString(segmentInfo)}`);
 
+    // if init segment is being requested
+    // we will have to re-append the init segment.
+    if (simpleSegment.map && !simpleSegment.map.bytes) {
+      this.logger_('going to request init segment.');
+      this.appendInitSegment_ = {
+        video: true,
+        audio: true
+      };
```
gesinger, line 2378: Should this be dependent on the type that this loader is handling?

test/segment-loader.test.js
```diff
@@ -3142,6 +3142,7 @@ QUnit.module('SegmentLoader', function(hooks) {
           appends[1].initSegment,
           'appended a different init segment'
         );
+        loader.appendInitSegment_.audio = true;
```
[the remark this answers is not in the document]
gesinger: It's probably worth adding those details to the comment (here and below) to ensure we don't forget why we force it in the future.

src/segment-loader.js
brandonocasey: I don't think it should because if we get an audio or video init segment on this segment loader, we should re-append it. It won't likely happen for a video only loader or an audio only loader, but at this point we won't actually know what type of content we are requesting, as we get that in trackinfo.

src/segment-loader.js
```diff
@@ -2368,6 +2367,16 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     this.logger_(`Requesting ${segmentInfoString(segmentInfo)}`);
 
+    // if init segment is being requested
+    // we will have to re-append the init segment.
```
gesinger, line 2371:
```suggestion
    // If there's an init segment associated with this segment, but it is not cached (identified by a lack of bytes),
    // then this init segment has never been seen before and should be appended.
    //
    // At this point the content type (audio/video or both) is not yet known, but it should be safe to set
    // both to true and leave the decision of whether to append the init segment to append time.
```

gesinger (approved):

PR #1133, opened 2021-05-20
fix: add tests for data uri, fix data uri in demo page

brandonocasey:
## Description
Adds tests for, and as soon as we get a vhs-utils update, fixes data-uri type urls.


scripts/index-demo-page.js
```diff
@@ -400,7 +400,7 @@
           }
 
           if (source.src) {
-            stateEls.url.value = source.src;
+            stateEls.url.value = encodeURI(source.src);
```
brandonocasey, line 403: without this data-uri's break the demo on reload.

gkatsev (approved):
LGTM, can be a `chore:`

gkatsev (approved):

PR #1137, opened 2021-06-03
Fix average bandwidth value from decaying when the system bandwidth being constant

KVPasupuleti:
## Description
In the `movingAverageBandwidthSelector` method, we are using this line to calculate the average.

```
average = decay * this.systemBandwidth + (1 - decay) * average;
```

I see a problem here.

To explain with an example,

I have an average value of `1000` currently and a new chunk got downloaded with a bandwidth of `500`. And the next chunk is downloaded after `1 second`.

The `systemBandwidth` is updated after each chunk gets downloaded.

So, my expectation is this `500` should be contributed to the average only once until the next chunk gets downloaded.

But, as the `checkABR` method is getting called for every `250ms`, this `500` is getting contributed `4 times` until the next chunk gets downloaded, by which the average value is getting decreased gradually because the `systemBandwidth`  is being constant until then.

You can check the logs attached below.

[Video-js-ABR.log](https://github.com/videojs/http-streaming/files/6593667/Video-js-ABR.log)

## Specific Changes proposed
- Change the average value only if the `systemBandwidth` changes and is greater than zero.

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev:
Wrote a test for this https://github.com/videojs/http-streaming/pull/1141
Once the other PR passes review (I'm doing something I'm unsure about) I'll get all the stuff merged in.

brandonocasey (approved):

PR #1138, opened 2021-06-04
feat: Caption services (608/708) metadata

gkatsev:
Add an option for caption services metadata in case the user wants to specify labels for 608/708 captions, override the ones provided in the manifest, or needs to add more information like character encoding (this isn't currently available but will be added some time in the future).
For HLS, an `EXT-X-MEDIA` tag can be specified with an `INSTREAM-ID` attribute. We already support this. https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-08#section-4.4.6.1
This PR updated mpd-parser which uses the ANSI 214 supplamental spec section 7.2 to parse out the same information from MPD files. https://github.com/videojs/mpd-parser/pull/131.

Adds a property called `captionServices` which has properties of the caption service IDs like `CC1` or `SERVICE1` and allows a user to specify a `language` and `label`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,7 +472,7 @@ This option defaults to `false`.
 ##### captionServices
 * Type: `object`
 * Default: undefined
-* Allow to override instream captions like 608 and 708 to provide labels or languaages.
+* Provide extra information, like a label or a language, for instream (608 and 708) captions.
 
 The captionServices options object has properties that map to the caption services. Each property is an object itself that includes several properties, like a label or language.
 
--- a/README.md
+++ b/README.md
@@ -58,6 +58,8 @@ Video.js Compatibility: 6.0, 7.0
       - [handlePartialData](#handlepartialdata)
       - [liveRangeSafeTimeDelta](#liverangesafetimedelta)
       - [captionServices](#captionservices)
+        - [Format](#format)
+        - [Example](#example)
   - [Runtime Properties](#runtime-properties)
     - [vhs.playlists.master](#vhsplaylistsmaster)
     - [vhs.playlists.media](#vhsplaylistsmedia)
@@ -475,8 +477,8 @@ This option defaults to `false`.
 The captionServices options object has properties that map to the caption services. Each property is an object itself that includes several properties, like a label or language.
 
 For 608 captions, the service names are `CC1`, `CC2`, `CC3`, and `CC4`. For 708 captions, the service names are `SERVICEn` where `n` is a digit between `1` and `63`.
-Format:
-```json
+###### Format
+```js
 {
   vhs: {
     captionServices: {
@@ -488,8 +490,8 @@ Format:
   }
 }
 ```
-Example:
-```json
+###### Example
+```js
 {
   vhs: {
     captionServices: {
--- a/README.md
+++ b/README.md
@@ -483,8 +483,9 @@ For 608 captions, the service names are `CC1`, `CC2`, `CC3`, and `CC4`. For 708 
   vhs: {
     captionServices: {
       [serviceName]: {
-        language: String,
-        label: String
+        language: String, // optional
+        label: String, // optional
+        default: boolean // optional
       }
     }
   }
@@ -501,7 +502,8 @@ For 608 captions, the service names are `CC1`, `CC2`, `CC3`, and `CC4`. For 708 
       },
       SERVICE1: {
         langauge: 'kr',
-        label: 'Korean'
+        label: 'Korean',
+        default: true
       }
     }
   }
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Video.js Compatibility: 6.0, 7.0
       - [cacheEncryptionKeys](#cacheencryptionkeys)
       - [handlePartialData](#handlepartialdata)
       - [liveRangeSafeTimeDelta](#liverangesafetimedelta)
+      - [captionServices](#captionservices)
   - [Runtime Properties](#runtime-properties)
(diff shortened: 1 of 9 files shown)


src/util/text-tracks.js
```diff
@@ -24,12 +24,28 @@ export const createCaptionsTrackIfNotExists = function(inbandTextTracks, tech, c
       // in the m3u8 for us to use
       inbandTextTracks[captionStream] = track;
     } else {
+      const captionServices = tech.options_.vhs && tech.options_.vhs.captionServices || {};
```
gkatsev, line 27: This section gets used when we have caption services, but they aren't being specified in the manifests. When specified in the manifest, we will apply the properties in media groups https://github.com/videojs/http-streaming/blob/ad9aa2ef67fa90557af6afa7878d911b5f1093ac/src/media-groups.js#L641-L662
which will need an update to use this captionServices data

src/util/text-tracks.js
```diff
@@ -16,20 +16,41 @@ export const createCaptionsTrackIfNotExists = function(inbandTextTracks, tech, c
...
+      }
+
       // Otherwise, create a track with the default `CC#` label and
       // without a language
       inbandTextTracks[captionStream] = tech.addRemoteTextTrack({
         kind: 'captions',
-        id: captionStream,
-        label: captionStream
+        id: instreamId,
+        default: def,
```
gkatsev, line 51: For whatever reason this doesn't seem to do anything, though, I'd expect it to. It does let us override it above in the media groups sections.

README.md
```diff
@@ -466,6 +469,47 @@ This option defaults to `false`.
 * Default: [`SAFE_TIME_DELTA`](https://github.com/videojs/http-streaming/blob/e7cb63af010779108336eddb5c8fd138d6390e95/src/ranges.js#L17)
 * Allow to  re-define length (in seconds) of time delta when you compare current time and the end of the buffered range.
 
+##### captionServices
+* Type: `object`
+* Default: undefined
+* Allow to override instream captions like 608 and 708 to provide labels or languaages.
```
gesinger, line 475:
```suggestion
* Allow overriding of instream captions like 608 and 708 to provide labels and/or languages.
```

src/util/text-tracks.js
```diff
@@ -16,20 +16,41 @@ export const createCaptionsTrackIfNotExists = function(inbandTextTracks, tech, c
...
+      const captionServices = tech.options_.vhs && tech.options_.vhs.captionServices || {};
+      let label = captionStream;
+      let language = captionStream;
+      let def = false;
+
+      if (captionServices[instreamId]) {
+        label = captionServices[instreamId].label;
+        language = captionServices[instreamId].language;
+        def = captionServices[instreamId].default;
+      }
```
gesinger, line 44:
```suggestion
      const captionService = captionServices[instreamId];

      if (captionService) {
        label = captionService.label;
        language = captionService.language;
        def = captionService.default;
      }
```

src/util/text-tracks.js
gesinger: Might be worth adding a comment here mentioning that.

src/util/text-tracks.js
gesinger: Worth a comment to look into it?

test/master-playlist-controller.test.js
```diff
@@ -3260,15 +3260,77 @@ QUnit.test(
       .map(cap => Object.assign({name: cap.id}, cap));
 
     assert.equal(capsArr.length, 4, '4 closed-caption tracks defined in playlist');
-    assert.equal(addedCaps.length, 2, '2 CEA608 tracks added internally');
+    assert.equal(addedCaps.length, 4, '4 tracks added internally');
```
gesinger, line 3263:
```suggestion
    // 2 608 tracks, 2 708 tracks
    assert.equal(addedCaps.length, 4, '4 tracks added internally');
```

test/master-playlist-controller.test.js
```diff
@@ -3260,15 +3260,77 @@ QUnit.test(
...
+    // media
+    this.standardXHRResponse(this.requests.shift());
+
+    const master = masterPlaylistController.masterPlaylistLoader_.master;
+    const caps = master.mediaGroups['CLOSED-CAPTIONS'].CCs;
+    const capsArr = Object.keys(caps).map(key => Object.assign({name: key}, caps[key]));
+    const addedCaps = masterPlaylistController.mediaTypes_['CLOSED-CAPTIONS'].groups.CCs
+      .map(cap => Object.assign({name: cap.id}, cap));
+
+    assert.equal(capsArr.length, 4, '4 closed-caption tracks defined in playlist');
```
gesinger, line 3327:
```suggestion
    // 2 608 tracks, 2 708 tracks
    assert.equal(capsArr.length, 4, '4 closed-caption tracks defined in playlist');
```

gkatsev: Added it as part of the test message

test/master-playlist-controller.test.js
gkatsev: Added it as part of the test message

README.md
gkatsev: Reworded entirely.

src/util/text-tracks.js
gkatsev: added

src/util/text-tracks.js
gkatsev: Yeah, makes sense. Added

gesinger (approved):

PR #1141, opened 2021-06-18
test: add a test moving average decaying even without new data

gkatsev:


test/playlist-selectors.test.js
```diff
@@ -64,6 +65,50 @@ test('Exponential moving average has a configurable decay parameter', function(a
   assert.equal(playlist.attributes.BANDWIDTH, 50, 'selected the middle playlist');
 });
 
+test('Calling exponential moing average wont decay average unless new bandwidth data was provided', function(assert) {
```
brandonocasey, line 68:
```suggestion
test('Calling exponential moving average won\'t decay average unless new bandwidth data was provided', function(assert) {
```

scripts/rollup.config.js
```diff
@@ -57,6 +59,11 @@ const options = {
     }
     defaults.module.unshift('replace');
 
+    if (NO_TEST_BUNDLE) {
+      defaults.module.unshift('strip');
```
brandonocasey, line 63: should we use `rollup-plugin-stub` in tests to prevent having to remove stuff from the main build? This is what we do in video.js https://github.com/videojs/video.js/blob/main/test/unit/player.test.js#L947-L955

gkatsev: Looked into this, I think that we don't want to use stub anymore. The repo is archived. Also, it adds a lot of time to the build.

brandonocasey: Don't we want to always strip this code from the module/browser build? I think we can just remove the if statement and always add the strip plugin.

gkatsev: yeah, seem reasonable. I think I wasn't sure exactly how the test/mobule/browser builds stuff worked when I first added this.

brandonocasey (approved):

PR #1144, opened 2021-06-22
fix: ts segments that don't define all streams in the first pmt

brandonocasey:
All code was in mux.js


gkatsev (approved):
Fixes videojs/video.js#7230

PR #1147, opened 2021-06-23
chore: do not strip debugger statements

brandonocasey:
## Description
`debugger;` statements are currently being stripped.

### DIFF
--- a/scripts/rollup.config.js
+++ b/scripts/rollup.config.js
@@ -77,7 +77,8 @@ const options = {
       }),
       strip: strip({
         functions: ['TEST_ONLY_*'],
-        debugger: false
+        debugger: false,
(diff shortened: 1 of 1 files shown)


scripts/rollup.config.js
```diff
@@ -76,7 +76,8 @@ const options = {
         compress: {passes: 2}
       }),
       strip: strip({
-        functions: ['TEST_ONLY_*']
+        functions: ['TEST_ONLY_*'],
+        debugger: false
```
gkatsev, line 80: can we add `sourceMap: false` while we're at it?
```suggestion
        debugger: false,
        sourceMap: false
```

gkatsev (approved):

PR #1148, opened 2021-06-23
chore: skip playback tests in forks

brandonocasey:
## Description
Testing to see if we can skip playback tests for forks like this.


gkatsev (approved):
🎉

PR #1152, opened 2021-06-29
fix: remove invalid media groups

brandonocasey:
Fixes `Song - ts Audio only` which doesn't work because it has invalid media groups (without uris)

### DIFF
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -11,6 +11,33 @@ export const createPlaylistID = (index, uri) => {
 };
 
 /**
+ * Loops through all supported media groups in master and calls the provided
+ * callback for each group
+ *
+ * @param {Object} master
+ *        The parsed master manifest object
+ * @param {Function} callback
+ *        Callback to call for each media group
+ */
+export const forEachMediaGroup = (master, callback) => {
+  if (!master.mediaGroups) {
+    return;
+  }
+  ['AUDIO', 'SUBTITLES'].forEach((mediaType) => {
+    if (!master.mediaGroups[mediaType]) {
+      return;
+    }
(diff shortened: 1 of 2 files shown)


src/manifest.js
```diff
@@ -105,34 +132,18 @@ export const parseManifest = ({
...
-/**
- * Loops through all supported media groups in master and calls the provided
- * callback for each group
- *
- * @param {Object} master
- *        The parsed master manifest object
- * @param {Function} callback
- *        Callback to call for each media group
- */
-export const forEachMediaGroup = (master, callback) => {
```
brandonocasey, line 120: moved this function up so that it could be used.

src/manifest.js
```diff
@@ -105,34 +132,18 @@ export const parseManifest = ({
...
-    if (!master.mediaGroups[mediaType]) {
-      return;
-    }
-    for (const groupKey in master.mediaGroups[mediaType]) {
-      for (const labelKey in master.mediaGroups[mediaType][groupKey]) {
-        const mediaProperties = master.mediaGroups[mediaType][groupKey][labelKey];
+  forEachMediaGroup(manifest, (properties, mediaType, groupKey, labelKey) => {
+    if (!properties.uri && (!properties.playlists || !properties.playlists[0].uri)) {
+      delete manifest.mediaGroups[mediaType][groupKey][labelKey];
+      onwarn(`removing media group ${mediaType} - ${groupKey} - ${labelKey} as it lacks a uri`);
```
gesinger, line 138: This may need a check for `onwarn`.

gesinger (commented):
Might be good to add a test.

PR #1155, opened 2021-07-02
chore: use transmuxer debug/warn/error events to the debug log

brandonocasey:
Log debug/warn/error from the transmuxer.
https://github.com/videojs/mux.js/pull/391

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "global": "^4.4.0",
     "m3u8-parser": "4.7.0",
     "mpd-parser": "0.17.0",
-    "mux.js": "5.12.0",
+    "mux.js": "5.12.1",
     "video.js": "^6 || ^7"
   },
   "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "global": "^4.4.0",
     "m3u8-parser": "4.7.0",
     "mpd-parser": "0.17.0",
-    "mux.js": "5.11.3",
+    "mux.js": "5.12.0",
     "video.js": "^6 || ^7"
   },
   "peerDependencies": {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -81,7 +81,7 @@
     "lodash-compat": "^3.10.0",
     "nomnoml": "^0.3.0",
     "rollup": "^2.36.1",
-    "rollup-plugin-worker-factory": "0.5.5",
+    "rollup-plugin-worker-factory": "0.5.7",
```
brandonocasey, line 84: Had to update `rollup-plugin-worker-factory` to call `onmessage` in `dispatchEvent` as we use `onmessage` in this scenario in tests.

src/media-segment-request.js
```diff
@@ -415,7 +427,10 @@ const handleSegmentBytes = ({
   isEndOfTimeline,
   endedTimelineFn,
   dataFn,
-  doneFn
+  doneFn,
```
brandonocasey, line 430: Might do a refactor after this pull request, 90% of these functions don't care about these handlers, why do we have to call each one out in the parameters every single time...

src/transmuxer-worker.js
```diff
@@ -155,6 +155,18 @@ const wireTransmuxerEvents = function(self, transmuxer) {
...
+    self.postMessage({action: 'debug', message});
+  });
+
+  transmuxer.on('error', function({message}) {
+    self.postMessage({action: 'error', message});
+  });
+
+  transmuxer.on('warn', function({message}) {
+    self.postMessage({action: 'warn', message});
+  });
```
gesinger, line 168: I know the mux.js commit is already in, but I think we may be better off using `log` with a level and message. `error` usually is a special case separate from standard log messages, and it would also allow us to only have one function all the way through the chain.

gkatsev (dismissed):

gkatsev (approved):

PR #1157, opened 2021-07-06
chore: use the new npm cache option when setting up node

brandonocasey:
## Description
Testing with the new cache option.


gkatsev (approved):
Much nicer! Now we just need to propagate it everywhere else, heh.

Issue #1057, opened 2021-01-28
Tracking for a Major version of VHS


brandonocasey:
[See a similar story for Video.js](https://github.com/videojs/video.js/issues/7068)

gyanex:
How to check if HLS is in use before using 
var vhs = player.tech().vhs;

PR #1162, opened 2021-07-12
fix: Use segmentInfo.trackInfo to get segment information.

brandonocasey:
## Description
This prevents an issue shown via the test that I added. Mainly that a playlist change that happens before we append data from a `pendingSegment_` will cause reset `this.currentMediaInfo_` to be null. Thus causing `handleData_`'s call to `hasEnoughInfoToAppend_`  to throw on `this.currentMediaInfo_.hasAudio`

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -811,12 +811,14 @@ export default class SegmentLoader extends videojs.EventTarget {
    *         TimeRange object representing the current buffered ranges
    */
   buffered_() {
-    if (!this.sourceUpdater_ || !this.startingMediaInfo_) {
+    const trackInfo = this.getMediaInfo_();
+
+    if (!this.sourceUpdater_ || !trackInfo) {
       return videojs.createTimeRanges();
     }
 
     if (this.loaderType_ === 'main') {
-      const { hasAudio, hasVideo, isMuxed } = this.startingMediaInfo_;
+      const { hasAudio, hasVideo, isMuxed } = trackInfo;
 
       if (hasVideo && hasAudio && !this.audioDisabled_ && !isMuxed) {
         return this.sourceUpdater_.buffered();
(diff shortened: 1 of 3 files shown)


src/segment-loader.js
```diff
@@ -1917,15 +1927,13 @@ export default class SegmentLoader extends videojs.EventTarget {
     }
 
     const segmentInfo = this.pendingSegment_;
+    const trackInfo = this.getCurrentMediaInfo_();
 
-    // no segment to append any data for or
-    // we do not have information on this specific
-    // segment yet
```
gkatsev, line 1923: is this comment no longer accurate?

src/segment-loader.js
```diff
@@ -1905,6 +1907,14 @@ export default class SegmentLoader extends videojs.EventTarget {
     return true;
   }
 
+  getCurrentMediaInfo_(segmentInfo = this.pendingSegment_) {
+    return segmentInfo && segmentInfo.trackInfo || this.currentMediaInfo_;
```
gkatsev, line 1911: Am I reading this right that basically `currentMediaInfo_` is the trackInfo, but that there's a sync issue between what the SegmentLoader has (or doesn't have) and what the actual segmentInfo is. Does it make sense to even have `currentMediaInfo_` at all? Should we remove it and rely only on segmentInfo.trackInfo?

src/segment-loader.js
brandonocasey: this was removed in error.

src/segment-loader.js
brandonocasey: I was wondering the same myself during the creation of this pull request but was running into issues in a few areas trying to get segmentInfo.trackInfo working across the board. I think it's likely that we should do this in the future, but for now we should stick with this. I will add a task to #1057 to think about why we need current/starting media info and wether we should just be using `segmentInfo.trackInfo`.

gkatsev: Sounds good.

gkatsev (approved):

PR #1163, opened 2021-07-12
test: cleanup transmuxers in tests

brandonocasey:
## Description
Currently leaking some web workers between tests that we need to cleanup.


test/media-segment-request.test.js
```diff
@@ -62,20 +62,25 @@ const sharedHooks = {
...
     };
   },
   afterEach(assert) {
     this.realDecrypter.terminate();
     this.env.restore();
 
-    if (this.transmuxer) {
-      this.transmuxer.terminate();
-    }
+    this.transmuxers.forEach(function(transmuxer) {
```
brandonocasey, line 81: seemed safer to keep track of every transmuxer we create and clean them all up.

gkatsev (approved):

PR #1167, opened 2021-07-15
fix: do not add placeholder properties for AUDIO groups without a uri

brandonocasey:
Fixes Song - ts Audio only which doesn't work because it has invalid media groups (without uris)

### DIFF
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -2,7 +2,7 @@ import videojs from 'video.js';
 import window from 'global/window';
 import { Parser as M3u8Parser } from 'm3u8-parser';
 import { resolveUrl } from './resolve-url';
-import { getLastParts } from './playlist.js';
+import { getLastParts, isAudioOnly } from './playlist.js';
 
 const { log } = videojs;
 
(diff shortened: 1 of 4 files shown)


gkatsev (approved):
Verified fixed

src/manifest.js
```diff
@@ -279,11 +279,26 @@ export const addPropertiesToMaster = (master, uri) => {
...
+  const audioOnlyMaster = isAudioOnly(master);
 
   forEachMediaGroup(master, (properties, mediaType, groupKey, labelKey) => {
     const groupId = `placeholder-uri-${mediaType}-${groupKey}-${labelKey}`;
 
+    // add a playlist array under properties
     if (!properties.playlists || !properties.playlists.length) {
+      // if we have an audio only master and this media group does not have a uri.
+      // check if it is located in the real list. If it is don't add placeholder
+      // properties as we should never switch to it.
```
gesinger, line 291: I think the not switching to it confused me a bit. Would you say this statement is accurate?

```suggestion
      // If the manifest is audio only and this media group does not have a uri, check
      // if the media group is located in the main list of playlists. If it is, don't add
      // placeholder properties as it shouldn't be considered an alternate audio track.
```

gesinger (approved):

PR #1171, opened 2021-07-21
fix: exclude playlists on DRM key status other than usable

brandonocasey:
## Description
This prevents us from trying to play drm content that is going to fail to play.


src/videojs-http-streaming.js
```diff
@@ -999,6 +999,18 @@ class VhsHandler extends Component {
       audioMedia: audioPlaylistLoader && audioPlaylistLoader.media()
     });
 
+    this.player_.tech_.on('keystatuschange', (e) => {
+      // typically 'output-restricted', but anything other than usable will
+      // result in a failure to play.
+      if (e.status !== 'usable') {
```
gesinger, line 1005: I think we should be careful here. Since there are various statuses: https://w3c.github.io/encrypted-media/#dom-mediakeystatus , some of which we handle, i.e., `expired`: https://github.com/videojs/videojs-contrib-eme/blob/5c441dca1896fc73dff2eeb855dd8e60f99c5da4/src/eme.js#L121-L126

It might be better to only blacklist on specific statuses.

test/videojs-http-streaming.test.js
```diff
@@ -4503,6 +4503,56 @@ QUnit.test('configures eme for HLS on source buffer creation', function(assert)
   }, 'set source eme options');
 });
 
+QUnit.test('eme handles keystatuschange where status is not usable', function(assert) {
```
gesinger, line 4506: Would be good to have one more test where status is usable.

scripts/sources.json
```diff
@@ -393,5 +393,49 @@
...
+    "features": [],
+    "keySystems": {
+      "com.widevine.alpha": {
+        "url": "https://proxy.uat.widevine.com/proxy?video_id=GTS_SW_SECURE_CRYPTO_HDCP_V2&provider=widevine_test"
+      }
+    }
+  },
+  {
+    "name": "HDCP v2.1 DRM dash",
+    "uri": "https://storage.googleapis.com/wvmedia/cenc/h264/tears/tears.mpd@21",
```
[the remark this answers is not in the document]
brandonocasey: I added the `#` so that the source will still work, but we will have a url to sync the dropdown with when we reload the page.

src/videojs-http-streaming.js
brandonocasey: switched this to only exclude on output-restricted.

test/videojs-http-streaming.test.js
brandonocasey: added.

src/videojs-http-streaming.js
```diff
@@ -1000,12 +1000,11 @@ class VhsHandler extends Component {
     });
 
     this.player_.tech_.on('keystatuschange', (e) => {
-      // typically 'output-restricted', but anything other than usable will
-      // result in a failure to play.
-      if (e.status !== 'usable') {
+      // If there's an error, the status is 'output-restricted'
```
gesinger, line 1003:
```suggestion
```

gesinger (dismissed):

PR #1174, opened 2021-07-26
test: fix tests in chrome 92 by disposing media elements/ blob url

brandonocasey:
## Description
Chrome 92 requires media elements to be torn down completely before they are removed from memory. This includes unsetting the src property and removing the src attribute. 

This is only an issue when a lot of players are created and disposed on a page. For us this mostly happens in tests.

The error in question:
Blocked attempt to create a WebMediaPlayer as there are too many WebMediaPlayers already in existence. See crbug.com/1144736#c27

### DIFF
--- a/test/segment-loader.test.js
+++ b/test/segment-loader.test.js
@@ -15,7 +15,6 @@ import mp4probe from 'mux.js/lib/mp4/probe';
 import {
   playlistWithDuration,
   standardXHRResponse,
-  setupMediaSource,
   MockTextTrack
 } from './test-helpers.js';
 import {
@@ -831,7 +830,7 @@ QUnit.module('SegmentLoader', function(hooks) {
     });
 
     QUnit.test('only appends one segment at a time', function(assert) {
-      return setupMediaSource(loader.mediaSource_, loader.sourceUpdater_).then(() => {
+      return this.setupMediaSource(loader.mediaSource_, loader.sourceUpdater_).then(() => {
 
         loader.playlist(playlistWithDuration(10));
         loader.load();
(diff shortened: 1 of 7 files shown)


gkatsev (approved):

test/segment-loader.test.js
```diff
@@ -831,7 +830,7 @@ QUnit.module('SegmentLoader', function(hooks) {
     });
 
     QUnit.test('only appends one segment at a time', function(assert) {
-      return setupMediaSource(loader.mediaSource_, loader.sourceUpdater_).then(() => {
+      return this.setupMediaSource(loader.mediaSource_, loader.sourceUpdater_).then(() => {
```
gkatsev, line 833: where does `this.setupMediaSource` is coming from here? I'm not seeing where it is defined, and `setupMediaSource` is no longer being imported.

gkatsev (commented):
Just one question

test/segment-loader.test.js
brandonocasey: its defined in the beforeEach/afterEach to handle the creation of the video element and then the removal.

brandonocasey: https://github.com/videojs/http-streaming/pull/1174/files#diff-76e6f47cbe1dfa39fb03a86becd7cc8d79ab123c00f4c1a74aec433201ae0115R70

gkatsev: aaah, cool. Missed that this file used the default hooks.

gkatsev (approved):
:shipit:

PR #1176, opened 2021-07-29
fix: Keep media update timeout alive so live playlists can recover

brandonocasey:
## Description
Right now for live hls streams we can be stuck buffering even though we should have recovered because we don't attempt to retry requests. You can reproduce this:

1. Select `Unified Streaming Live HLS` 
2. Play
3. Set network to offline in dev tools
4. Wait for the player to stall
5. Set network back to normal in dev tools
6. Player will never recover

This happens because we exclude and pause the current playlist loader when it fails a request.  Then we switch to same playlist again and since it is the same playlist we don't try to request it again. So now we have no mediaUpdateTimeout and we did not request the playlist again.

## The fix
1. We need a way to make sure that we start a mediaUpdateTimeout in `media`
2. We need mediaUpdateTimeouts to create media update timeouts 
3. We do not need to **always** return `true` in `shouldSwitchToMedia` for live playlists if the playlist is the same. This gave us state we have now as we were relying on abr to refresh live playlists in some scenarios.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -63,18 +63,23 @@ const shouldSwitchToMedia = function({
 
   const sharedLogLine = `allowing switch ${currentPlaylist && currentPlaylist.id || 'null'} -> ${nextPlaylist.id}`;
 
+  if (!currentPlaylist) {
+    log(`${sharedLogLine} as current playlist is not set`);
+    return true;
+  }
+
+  // no need to switch if playlist is the same
+  if (nextPlaylist.id === currentPlaylist.id) {
+    return false;
+  }
+
   // If the playlist is live, then we want to not take low water line into account.
   // This is because in LIVE, the player plays 3 segments from the end of the
   // playlist, and if `BUFFER_LOW_WATER_LINE` is greater than the duration availble
   // in those segments, a viewer will never experience a rendition upswitch.
-  if (!currentPlaylist) {
-    log(`${sharedLogLine} as current playlist ` + (!currentPlaylist ? 'is not set' : 'is live'));
+  if (!currentPlaylist.endList) {
+    log(`${sharedLogLine} as current playlist is live`);
     return true;
-  }
-
-  // no need to switch playlist is the same
-  if (nextPlaylist.id === currentPlaylist.id) {
-    return false;
   }
 
   const maxBufferLowWaterLine = experimentalBufferBasedABR ?
@@ -351,7 +356,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
   checkABR_() {
     const nextPlaylist = this.selectPlaylist();
 
-    if (this.shouldSwitchToMedia_(nextPlaylist)) {
+    if (nextPlaylist && this.shouldSwitchToMedia_(nextPlaylist)) {
       this.switchMedia_(nextPlaylist, 'abr');
     }
   }
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -67,7 +67,7 @@ const shouldSwitchToMedia = function({
   // This is because in LIVE, the player plays 3 segments from the end of the
   // playlist, and if `BUFFER_LOW_WATER_LINE` is greater than the duration availble
   // in those segments, a viewer will never experience a rendition upswitch.
-  if (!currentPlaylist || !currentPlaylist.endList) {
+  if (!currentPlaylist) {
     log(`${sharedLogLine} as current playlist ` + (!currentPlaylist ? 'is not set' : 'is live'));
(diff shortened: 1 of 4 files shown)


src/playlist-loader.js
```diff
@@ -705,14 +705,20 @@ export default class PlaylistLoader extends EventTarget {
...
+      window.clearTimeout(this.mediaUpdateTimeout);
+      this.mediaUpdateTimeout = null;
+    }
     const media = this.media();
 
     if (shouldDelay) {
       const delay = media ? ((media.partTargetDuration || media.targetDuration) / 2) * 1000 : 5 * 1000;
 
-      this.mediaUpdateTimeout = window.setTimeout(() => this.load(), delay);
+      this.mediaUpdateTimeout = window.setTimeout(() => {
```
brandonocasey, line 717: imo this should be called something else, but I think we should save it for the playlist-loader base class refactor as dash-playlist-loader does the same thing.

gkatsev (approved):
Haven't tested yet but changes make sense to me.

src/master-playlist-controller.js
```diff
@@ -67,7 +67,7 @@ const shouldSwitchToMedia = function({
   // This is because in LIVE, the player plays 3 segments from the end of the
   // playlist, and if `BUFFER_LOW_WATER_LINE` is greater than the duration availble
   // in those segments, a viewer will never experience a rendition upswitch.
-  if (!currentPlaylist || !currentPlaylist.endList) {
+  if (!currentPlaylist) {
     log(`${sharedLogLine} as current playlist ` + (!currentPlaylist ? 'is not set' : 'is live'));
```
gesinger, line 71: Per the comment above, will this mean that we should have a way of ignoring the low water for live in the rest of this function? 

And we may need to update this line and the comment above

src/playlist-loader.js
```diff
@@ -619,6 +613,8 @@ export default class PlaylistLoader extends EventTarget {
       return;
     }
 
+    this.updateMediaUpdateTimeout_(refreshDelay(playlist, true));
```
gesinger, line 616: May be worth a comment above here saying why we update the timeout even if it's switching to the active playlist

test/master-playlist-controller.test.js
```diff
@@ -2033,36 +2033,6 @@ QUnit.test(
...
-    this.masterPlaylistController.mainSegmentLoader_.trigger('bandwidthupdate');
-    assert.equal(
-      mediaChanges.length,
-      3,
-      'changes live media when sufficient forward buffer and higher ' +
-               'bandwidth playlist'
-    );
-
-    mediaChanges.length = 0;
-
```
gesinger, line 2065: Do we want to remove these tests?

brandonocasey: These only tested that live content always switches as far as I could tell. We might have to add one more back in depending on changes that I have to make in shouldSwitchToMedia.

src/master-playlist-controller.js
```diff
@@ -351,7 +356,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
   checkABR_() {
     const nextPlaylist = this.selectPlaylist();
 
-    if (this.shouldSwitchToMedia_(nextPlaylist)) {
+    if (nextPlaylist && this.shouldSwitchToMedia_(nextPlaylist)) {
```
brandonocasey, line 359: minor fix here. I saw that we throw an error if `nextPlaylist` is undefined.

src/playlist-loader.js
brandonocasey: added

src/playlist-loader.js
```diff
@@ -705,14 +710,20 @@ export default class PlaylistLoader extends EventTarget {
...
+    }
     const media = this.media();
 
     if (shouldDelay) {
       const delay = media ? ((media.partTargetDuration || media.targetDuration) / 2) * 1000 : 5 * 1000;
 
-      this.mediaUpdateTimeout = window.setTimeout(() => this.load(), delay);
+      this.mediaUpdateTimeout = window.setTimeout(() => {
+        this.mediaUpdateTimeout = null;
+        this.load();
```
gkatsev, line 724: is it hard to add a test for these lines?

gkatsev (commented):
Can confirm that this fixes the issue of going offline and coming back!

src/playlist-loader.js
brandonocasey: i'll see what I can do.

brandonocasey: added

gkatsev (approved):

gesinger (approved):

PR #1182, opened 2021-08-11
chore: add a github-release github action

brandonocasey:
## Description
Automates the github release changelog process.

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -34,12 +34,16 @@ jobs:
       - name: build
         run: npm run build-prod
 
-      - name: Generate current release changelog
-        run: ./node_modules/.bin/conventional-changelog -p videojs -r 2 -o RELEASE-CHANGELOG.md
-
       - name: Check if this is a pre-release
-        run: echo ::set-output name=IS_PRE_RELEASE::$(./node_modules/.bin/is-prerelease && echo "true" || echo "false")
+        run: echo ::set-output name=IS_PRE_RELEASE::$(npx is-prerelease && echo "true" || echo "false")
         id: pre-release-check
+
+      - name: delete prerelease tags to build changelog
+        run: git tag -l | grep -vE '^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$' | xargs git tag -d
+        if: ${{ steps.pre-release-check.outputs.IS_PRE_RELEASE == 'false' }}
+
+      - name: Generate current release changelog
+        run: npx conventional-changelog -p videojs -r 2 -o RELEASE-CHANGELOG.md
 
       - name: get dashed package version for netlify
         run: echo ::set-output name=VERSION::$(node -e "process.stdout.write(require('./package.json').version.split('.').join('-'))")
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -8,6 +8,8 @@ on:
     - "v[0-9]+.[0-9]+.[0-9]+-*"
 jobs:
   github-release:
+    env:
+      NETLIFY_BASE: 'videojs-http-streaming.netlify.app'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/github-release.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Generate current release changelog
+        run: ./node_modules/.bin/conventional-changelog -p videojs -r 2 -o RELEASE-CHANGELOG.md
+
+      - name: Create Github release
+        uses: softprops/action-gh-release@v1
+        with:
+          body_path: RELEASE-CHANGELOG.md
+          token: ${{github.token}}
+          files:
```
gkatsev, line 37: probably want to list the dist files here

.github/workflows/github-release.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: npm i --prefer-offline --no-audit
+
+      - name: Generate current release changelog
+        run: ./node_modules/.bin/conventional-changelog -p videojs -r 2 -o RELEASE-CHANGELOG.md
+
+      - name: Create Github release
+        uses: softprops/action-gh-release@v1
+        with:
+          body_path: RELEASE-CHANGELOG.md
+          token: ${{github.token}}
```
gkatsev, line 36: seems like this is the default, is it worth noting?

.github/workflows/github-release.yml
gkatsev: I guess we'd need to run the build step first, though.

.github/workflows/github-release.yml
```diff
@@ -0,0 +1,37 @@
+name: github-release
+on:
+  push:
+    tags:
+    - "v[0-9]+.[0-9]+.[0-9]+"
```
gkatsev, line 5: We don't really publish pre-releases often, but I wonder if it'll be better to use the `if: startsWith(github.ref, 'refs/tags/')` that's mentioned in the gh-release action instead?

gkatsev (commented):
Awesome!

.github/workflows/github-release.yml
brandonocasey: Seems to work, from my testing on a separate repo, not sure if we need new permissions.

gkatsev: github adds a token by default for the current repo as well, so, it should be fine. A separate one is only needed if you need special permissions or something.

.github/workflows/github-release.yml
brandonocasey: It might be easier to just do it in ci yeah, but I do see benefits to doing it in it's own file/action as some of our projects won't use this action at all.

.github/workflows/github-release.yml
```diff
@@ -0,0 +1,57 @@
...
+        id: get-version
+        shell: bash
+
+      - name: Generate current release changelog
+        run: ./node_modules/.bin/conventional-changelog -p videojs -r 2 -o RELEASE-CHANGELOG.md
+
+      - name: add netlify preview to release notes
+        run: |
+          echo "" >> RELEASE-CHANGELOG.md
+          echo "[netlify preview](https://v${{steps.get-version.outputs.VERSION}--videojs-http-streaming.netlify.app)" >> RELEASE-CHANGELOG.md
```
gkatsev, line 50: cool idea!
Does this get sluggified properly? Since the preview url is something like https://v2-10-0--videojs-http-streaming.netlify.app/ rather than https://v2.10.0--videojs-http-streaming.netlify.app/

.github/workflows/github-release.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  github-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        # We neeed to fetch the entire history as conventional-changelog needs
+        # access to any number of git commits to build the changelog.
+        with:
+          fetch-depth: 0
```
gkatsev, line 15: I guess the alternative is to use a specific number and hope that we never actually cross that threshold between releases haha. Not sure that's worth doing, though.

.github/workflows/github-release.yml
gkatsev: Oh, I see, this piece is required otherwise we'll have to run on every push or at best to every push to the main branch.
Maybe we can add a second match for builds something like
```suggestion
    # match semver versions
    - "v[0-9]+.[0-9]+.[0-9]+"
    # match semver pre-releases
    - "v[0-9]+.[0-9]+.[0-9]+-*?"
```

.github/workflows/github-release.yml
brandonocasey: It does, i convert it to a dashed version in the `node -e` command above via `split('.').join('-')`

.github/workflows/github-release.yml
brandonocasey: Yeah I think we just want to download it all, If grabbing the entire repo takes longer for releases to be generated, it isn't the end of the world. As long as it works.

gkatsev: Yeah, if we see it ends up being slow, we can cap this at that point.

.github/workflows/github-release.yml
gkatsev: ah, cool, missed that piece.

gkatsev (approved):

gkatsev (dismissed):
nice!

gkatsev (approved):

gkatsev (approved):

PR #1186, opened 2021-08-17
chore: various demo page fixes and enhancements

brandonocasey:
## Description
* Adds back in the run tests link (which is removed on netlify).
* Correctly disposes of stats on player dispose.
* Defaults to the new liveui 
* Saves the state of `player.muted()` to the test page state so that muting/unmuting the player is saved across reloads.
* Moves the preloaded source dropdown to the top of the form

### DIFF
--- a/index.html
+++ b/index.html
@@ -69,7 +69,7 @@
   <div class="tab-content container-fluid">
 
     <div class="tab-pane active" id="sources" role="tabpanel">
-      <div class="input-group">
+      <div class="input-group mb-2">
         <span class="input-group-text"><label for=load-source>Preloaded Sources</label></span>
         <select id=load-source class="form-select">
           <optgroup label="hls">
--- a/index.html
+++ b/index.html
@@ -69,6 +69,23 @@
   <div class="tab-content container-fluid">
 
     <div class="tab-pane active" id="sources" role="tabpanel">
+      <div class="input-group">
+        <span class="input-group-text"><label for=load-source>Preloaded Sources</label></span>
+        <select id=load-source class="form-select">
+          <optgroup label="hls">
+          </optgroup>
+          <optgroup label="dash">
+          </optgroup>
+          <optgroup label="drm">
+          </optgroup>
+          <optgroup label="live">
+          </optgroup>
+          <optgroup label="low latency live">
+          </optgroup>
+          <optgroup label="json manifest object">
+          </optgroup>
+        </select>
+      </div>
 
       <label for=url class="form-label">Source URL</label>
       <div class="input-group">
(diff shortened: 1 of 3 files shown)


scripts/index.js
```diff
@@ -338,6 +338,12 @@
   };
 
   var setupPlayerStats = function(player) {
+    player.on('dispose', () => {
+      if (window.statsTimer) {
+        window.clearInterval(window.statsTimer);
+        window.statsTimer = null;
+      }
+    });
```
gkatsev, line 346: ~this does get cleared out when we make a new player.~
It was supposed to be cleared here but I accidentally put the lines backwards https://github.com/videojs/http-streaming/blob/e8230a9b22b9a1b3ccb8f0926798c5ef42f53ed6/scripts/index.js#L547-L551

index.html
```diff
@@ -69,6 +69,23 @@
   <div class="tab-content container-fluid">
 
     <div class="tab-pane active" id="sources" role="tabpanel">
+      <div class="input-group">
```
gkatsev, line 72:
```suggestion
      <div class="input-group mb-2">
```

index.html
```diff
@@ -221,9 +221,19 @@
...
 
+  <div class="text-center p-3" id=unit-test-link>
+    <a href="test/debug.html">Run unit tests</a>
+  </footer>
+
+  <script>
+    var unitTestLink = document.getElementById('unit-test-link');
+
+    if ((/netlify.app/).test(window.location.href.host)) {
+      unitTestLink.parentElement.removeChild(unitTestLink);
```
gkatsev, line 234: since we don't care about IE in this file, we can also use [`remove()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/remove)
```suggestion
      unitTestLink.remove();
```

index.html
```diff
@@ -122,7 +122,7 @@
         </div>
 
         <div class="form-check">
-          <input id=liveui type="checkbox" class="form-check-input">
+          <input id=liveui type="checkbox" class="form-check-input" checked>
```
gkatsev, line 125: 👍🏻

index.html
```diff
@@ -221,9 +221,19 @@
         <ul id="segment-metadata" class="col-8"></ul>
       </div>
     </div>
-
   </div>
 
+  <div class="text-center p-3" id=unit-test-link>
```
gkatsev, line 226:
```suggestion
  <footer class="text-center p-3" id=unit-test-link>
```

scripts/index.js
brandonocasey: probably better to do it in the actual function on dispose anyway. Keep it all contained. I will remove those lines.

index.html
brandonocasey: fancy

scripts/index.js
gkatsev: sounds good

gkatsev (approved):

Issue #1000, opened 2020-11-17
Player stucks with buffer empty


pmendozav:
##  Description
I have a live streaming with several renditions in 60fps. This is a simple description of these (id - resolution - segment size aprox)
- A - 256x144   - 200kb
- B - 416x234   - 320kb
- C - 640x360  - 500kb
- D - 960x540  - 1mb
- E - 1280x720  - 2mb
- F - 1280x720  - 3mb

When I tried to use the manifest I observed that player stucks observing the next situations:
- Keeps downloading manifests: from a single variant that sometimes is not the lower
- Keeps downloading manifests + segments: but the video does not play 
(buffer always is empty)

Some observations:
- I found that sometimes when the video is in buffering, player tries to change the rendition but keeps selecting the same and during switching abort() from segmentLoader is called and the segments are not loaded in buffer (keeping empty).
- Sometimes using the tool https://videojs-http-streaming.netlify.app/utils/stats/ I observed that if I force to use A/B/C the CPU in between 10-30% and heap uses 30mb aprox but if I force to use F and wait some minutes the memory demand is between 40-80% and heap grows also (without playback)
- In the case of F (major bandwidth), Safari requires only a couple of seconds to start and never stucks

## Sources
This is a sample of the live streaming I'm using:
https://hs-video-testing.s3.amazonaws.com/master_abc_60.m3u8

## Steps to reproduce
1. Go to https://videojs-http-streaming.netlify.app/utils/stats/
2. Use https://hs-video-testing.s3.amazonaws.com/master_abc_60.m3u8
3. Force to use F_60.m3u8
4. Wait

### Expected
Player could to stuck a couple of seconds for buffering but should to recover even if the major rendition is forced to use.

### Error output
No errors/warnings reported in the console

### videojs-http-streaming version
what version of videojs-http-streaming does this occur with?

### videojs version
what version of videojs does this occur with?
videojs 7.8.2, 7.8.4, 7.10.2

### Browsers
what browsers are affected? please include browser and version for each
* Chrome

### Platforms
what platforms are affected? please include operating system and version or device and version for each
* macOS Catalina

### Other Plugins
are any other videojs plugins being used on the page? If so, please list them with version below.
*

### Other JavaScript
are you using any other javascript libraries or frameworks on the page? if so please list them below.
*

Gregory-Gerard:
Maybe related to my own issue? #999 

Using an old version of video.js (7.2.3) "resolved" this issue in my case.

[...]

PR #1187, opened 2021-08-17
fix: fix seekable by using timestampOffset to convert expired/start to relative time

brandonocasey:
Lets us recover from the issue mentioned in #1000 but there still appears to be other issues.


brandonocasey:
This isn't the correct solution.

PR #1192, opened 2021-08-31
fix: playback watcher should fix seeking into a gap.

brandonocasey:
## Description
playback watcher should seek us out of a gap that was seeked into.

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -36,22 +36,42 @@ const timerCancelEvents = [
  * @return {boolean}
  *         Whether the current time should be considered close to the buffer
  */
-export const closeToBufferedContent = ({ buffered, targetDuration, currentTime }) => {
-  if (!buffered.length) {
-    return false;
+export const closeToBufferedContent = ({ buffered, audioBuffered, videoBuffered, targetDuration, currentTime }) => {
+  const twoSegmentDurations = (targetDuration - Ranges.TIME_FUDGE_FACTOR) * 2;
+  const bufferedToCheck = [audioBuffered, videoBuffered];
+
+  for (let i = 0; i < bufferedToCheck.length; i++) {
+    // skip null buffered
+    if (!bufferedToCheck[i]) {
+      continue;
+    }
+
+    const timeAhead = Ranges.timeAheadOf(bufferedToCheck[i], currentTime);
+
+    // if we are less than two video/audio segment durations behind,
+    // we haven't append enough to be close to buffered content.
+    if (timeAhead < twoSegmentDurations) {
+      return false;
+    }
   }
 
-  // It's possible that, on seek, a remove hasn't completed and the buffered range is
-  // somewhere past the current time. In that event, don't consider the buffered content
-  // close.
-  if (currentTime > buffered.start(0)) {
+  // default to using buffered, but if we don't have one
+  // use video or audio buffered
+  buffered = buffered && buffered.length ? buffered : null;
+  buffered = buffered || videoBuffered || audioBuffered;
+
+  const nextRange = Ranges.findNextRange(buffered, currentTime);
+
+  // if we don't have a next buffered range, we cannot be close to
+  // buffered content.
+  if (!nextRange.length) {
     return false;
   }
 
   // Since target duration generally represents the max (or close to max) duration of a
   // segment, if the buffer is within a segment of the current time, the gap probably
   // won't be closed, and current time should be considered close to buffered content.
-  return buffered.start(0) - currentTime < targetDuration;
+  return nextRange.start(0) - currentTime < (targetDuration + Ranges.TIME_FUDGE_FACTOR);
 };
 
 /**
@@ -109,33 +129,6 @@ export default class PlaybackWatcher {
       this.tech_.on(['seeked', 'seeking'], loaderChecks[type].reset);
     });
 
-    // handle seeking into a gap
-    this.tech_.on('seeking', () => {
-      let clear;
-      const checkForGaps = () => {
-        if (this.fixesBadSeeks_()) {
-          clear();
-        }
-      };
-
-      clear = () => {
-        this.tech_.off('seeking', clear);
-        this.tech_.off('seeked', clear);
-        this.tech_.off('dispose', clear);
-
-        ['main', 'audio'].forEach(function(type) {
-          mpc[`${type}SegmentLoader_`].off('appended', checkForGaps);
-        });
-      };
-
-      this.tech_.on('seeking', clear);
-      this.tech_.on('seeked', clear);
-      this.tech_.on('dispose', clear);
-
-      ['main', 'audio'].forEach(function(type) {
-        mpc[`${type}SegmentLoader_`].on('appended', checkForGaps);
-      });
-    });
     this.tech_.on('seekablechanged', fixesBadSeeksHandler);
     this.tech_.on('waiting', waitingHandler);
     this.tech_.on(timerCancelEvents, cancelTimerHandler);
@@ -274,7 +267,7 @@ export default class PlaybackWatcher {
    * @private
    */
   checkCurrentTime_() {
-    if (this.tech_.seeking() && this.fixesBadSeeks_()) {
+    if (this.fixesBadSeeks_()) {
       this.consecutiveUpdates = 0;
       this.lastRecordedTime = this.tech_.currentTime();
       return;
@@ -377,17 +370,22 @@ export default class PlaybackWatcher {
       return true;
     }
 
+    const sourceUpdater = this.masterPlaylistController_.sourceUpdater_;
     const buffered = this.tech_.buffered();
 
     if (
       closeToBufferedContent({
         buffered,
+        audioBuffered: sourceUpdater.audioBuffer ? sourceUpdater.audioBuffered() : null,
+        videoBuffered: sourceUpdater.videoBuffer ? sourceUpdater.videoBuffered() : null,
         targetDuration: this.media().targetDuration,
         currentTime
       })
     ) {
-      seekTo = buffered.start(0) + Ranges.SAFE_TIME_DELTA;
-      this.logger_(`Buffered region starts (${buffered.start(0)}) ` +
+      const nextRange = Ranges.findNextRange(buffered, currentTime);
+
+      seekTo = nextRange.start(0) + Ranges.SAFE_TIME_DELTA;
+      this.logger_(`Buffered region starts (${nextRange.start(0)}) ` +
                    ` just beyond seek point (${currentTime}). Seeking to ${seekTo}.`);
 
       this.tech_.setCurrentTime(seekTo);
@@ -447,7 +445,7 @@ export default class PlaybackWatcher {
     const seekable = this.seekable();
     const currentTime = this.tech_.currentTime();
 
-    if (this.tech_.seeking() && this.fixesBadSeeks_()) {
+    if (this.fixesBadSeeks_()) {
       // Tech is seeking or bad seek fixed, no action needed
       return true;
     }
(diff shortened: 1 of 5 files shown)


src/playback-watcher.js
```diff
@@ -115,6 +115,24 @@ export default class PlaybackWatcher {
...
+      const clear = () => {
+        this.tech_.off('seeking', clear);
+        this.tech_.off('seeked', clear);
+        window.clearTimeout(seekTimeout);
+      };
+
+      this.tech_.on('seeking', clear);
+      this.tech_.on('seeked', clear);
+
+      seekTimeout = window.setTimeout(() => {
```
brandonocasey, line 131: We might want to listen for segment loader appends here, rather than waiting 3s as that might not be enough time on slow connections. I also think that we should be checking if the player is buffering or not, but techWaiting might do that.

gesinger: I think we should go with what you said here about waiting for segment loader appends rather than a timeout. I think a timeout alone might get us into trouble with some of our potential resolution strategies.

src/playback-watcher.js
```diff
@@ -115,6 +115,24 @@ export default class PlaybackWatcher {
...
+        clear();
+        this.techWaiting_(false);
```
gesinger, line 133: I wonder if there's a better way of whether we should consider seeking or not rather than overriding whether we're actually seeking. Technically we are in a seeking state, but maybe we can just have something like `ignoreSeeks`.

src/playback-watcher.js
```diff
@@ -253,7 +267,7 @@ export default class PlaybackWatcher {
    * @private
    */
   checkCurrentTime_() {
-    if (this.tech_.seeking() && this.fixesBadSeeks_()) {
+    if (this.fixesBadSeeks_()) {
```
brandonocasey, line 270: `fixesBadSeeks_` already returns false if `seeking` is `false` on `tech` so checking `seeking` is redundant.

src/playlist-loader.js
```diff
@@ -340,7 +340,8 @@ export const updateMaster = (master, newMedia, unchangedCheck = isPlaylistUnchan
  *         The time in ms to wait before refreshing the live playlist
  */
 export const refreshDelay = (media, update) => {
-  const lastSegment = media.segments[media.segments.length - 1];
+  const segments = media.segments || [];
```
brandonocasey, line 343: Fixes an issue I was seeing in our multi period source where sometimes media does not have segments. I can make this a different pull request, but it was easy enough.

src/playback-watcher.js
```diff
@@ -356,17 +370,22 @@ export default class PlaybackWatcher {
...
         buffered,
+        audioBuffered: sourceUpdater.audioBuffer ? sourceUpdater.audioBuffered() : null,
+        videoBuffered: sourceUpdater.videoBuffer ? sourceUpdater.videoBuffered() : null,
         targetDuration: this.media().targetDuration,
         currentTime
       })
     ) {
-      seekTo = buffered.start(0) + Ranges.SAFE_TIME_DELTA;
-      this.logger_(`Buffered region starts (${buffered.start(0)}) ` +
+      const nextRange = Ranges.findNextRange(buffered, currentTime);
```
brandonocasey, line 385: Example values for why we need this:
```
buffered = [[0, 10], [21, 40]]
currentTime = 20
````
Previously we would just take `buffered.start(0)` which is in this case is before currentTime. Now we find the `TimeRange` that is after `currentTime` and use that.

File: src/ranges.js [hunk 6]
```diff
@@ -445,3 +445,32 @@ export const lastBufferedEnd = function(a) {
 
   return a.end(a.length - 1);
 };
+
+export const timeAheadOf = function(range, startTime) {
```
gesinger, line 449: This could probably use a JSDoc if only to describe its purpose.

src/ranges.js
```diff
@@ -445,3 +445,32 @@ export const lastBufferedEnd = function(a) {
...
+  let time = 0;
+
+  if (!range || !range.length || !range.end) {
```
gesinger, line 452: Is `!range.end` necessary?

src/playback-watcher.js
```diff
@@ -36,28 +36,42 @@ const timerCancelEvents = [
...
+    }
   }
 
-  // It's possible that, on seek, a remove hasn't completed and the buffered range is
-  // somewhere past the current time. In that event, don't consider the buffered content
-  // close.
-  if (currentTime > buffered.start(0)) {
+  // default to using buffered, but if we don't have one
+  // use video or audio buffered
+  buffered = buffered && buffered.length ? buffered : null;
```
brandonocasey, line 60: I saw a case were we had no bufferIntersection but we did have a video/audio buffer. Seeking to the start of the next videoBuffer fixed it for us.

src/playback-watcher.js
```diff
@@ -36,28 +36,42 @@ const timerCancelEvents = [
...
+  // buffered content.
+  if (!nextRange.length) {
     return false;
   }
 
   // Since target duration generally represents the max (or close to max) duration of a
   // segment, if the buffer is within a segment of the current time, the gap probably
   // won't be closed, and current time should be considered close to buffered content.
-  return buffered.start(0) - currentTime < targetDuration;
+  return nextRange.start(0) - currentTime < (targetDuration + Ranges.TIME_FUDGE_FACTOR);
```
brandonocasey, line 74: Added time fudge factor here because segment durations are never exact.

test/playback-watcher.test.js
```diff
@@ -1006,6 +1006,12 @@ QUnit.test('jumps to buffered content if seeking just before', function(assert)
     currentTime: () => currentTime,
     buffered: () => buffered
   };
+
+  Object.assign(playbackWatcher.masterPlaylistController_.sourceUpdater_, {
```
brandonocasey, line 1010: had to update this test to pass videoBuffered, other than that all tests continue to work.

src/playback-watcher.js
```diff
@@ -36,28 +36,42 @@ const timerCancelEvents = [
  * @return {boolean}
  *         Whether the current time should be considered close to the buffer
  */
-export const closeToBufferedContent = ({ buffered, targetDuration, currentTime }) => {
-  if (!buffered.length) {
-    return false;
-  }
+export const closeToBufferedContent = ({ buffered, audioBuffered, videoBuffered, targetDuration, currentTime }) => {
```
gesinger, line 39: JSDoc will need updating

test/playback-watcher.test.js
```diff
@@ -1627,32 +1682,88 @@ QUnit.test('respects allowSeeksWithinUnsafeLiveWindow flag', function(assert) {
...
     'returned false'
   );
 });
 
-QUnit.test('false if buffer is beyond target duration from current time', function(assert) {
+QUnit.test('false if zero length audioBuffered and videoBuffered', function(assert) {
   assert.notOk(
     closeToBufferedContent({
-      buffered: videojs.createTimeRanges([[14.1, 30]]),
+      audioBuffered: videojs.createTimeRanges(),
```
gesinger, line 1710:
```suggestion
      audioBuffered: videojs.createTimeRanges(),
      videoBuffered: videojs.createTimeRanges(),
```

test/playback-watcher.test.js
```diff
@@ -1663,21 +1774,55 @@ QUnit.test('false if buffer is beyond target duration from current time', functi
...
       targetDuration: 4,
-      currentTime: 14
+      currentTime: 23
     }),
-    'returned false'
+    'returned true'
+  );
+});
+
+QUnit.test('complex gaps with enough buffer ahead.', function(assert) {
```
gesinger, line 1796:
```suggestion
QUnit.test('complex gaps with enough buffer ahead', function(assert) {
```

test/playback-watcher.test.js
```diff
@@ -1663,21 +1774,55 @@ QUnit.test('false if buffer is beyond target duration from current time', functi
...
+      audioBuffered,
+      buffered,
+      targetDuration: 7,
+      currentTime: 3095.45
+    }),
+    'returned true'
+  );
+});
+
+QUnit.test('another complex gaps with enough buffer ahead.', function(assert) {
```
gesinger, line 1813:
```suggestion
QUnit.test('another complex gaps with enough buffer ahead', function(assert) {
```

src/playback-watcher.js
```diff
@@ -36,28 +36,42 @@ const timerCancelEvents = [
...
   }
 
-  // It's possible that, on seek, a remove hasn't completed and the buffered range is
-  // somewhere past the current time. In that event, don't consider the buffered content
-  // close.
-  if (currentTime > buffered.start(0)) {
+  // default to using buffered, but if we don't have one
+  // use video or audio buffered
+  buffered = buffered && buffered.length ? buffered : null;
+  buffered = buffered || videoBuffered || audioBuffered;
```
gkatsev, line 61:
```suggestion
  if (!buffered || buffered.length === 0) {
    buffered = videoBuffered || audioBuffered
  }
```

src/ranges.js
brandonocasey: not sure we do it in other functions in this file.

src/playback-watcher.js
```diff
@@ -356,17 +370,22 @@ export default class PlaybackWatcher {
...
+        videoBuffered: sourceUpdater.videoBuffer ? sourceUpdater.videoBuffered() : null,
         targetDuration: this.media().targetDuration,
         currentTime
       })
     ) {
-      seekTo = buffered.start(0) + Ranges.SAFE_TIME_DELTA;
-      this.logger_(`Buffered region starts (${buffered.start(0)}) ` +
+      const nextRange = Ranges.findNextRange(buffered, currentTime);
+
+      seekTo = nextRange.start(0) + Ranges.SAFE_TIME_DELTA;
```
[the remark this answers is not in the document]
brandonocasey: There are three parts.
1. We have to consider all content (including gaps) after current time to determine if we have appended two segments. Previously we just used the first buffer to determine this. This is in `closeToBufferedContent`
2. Next we need to find the range that is ahead of currentTime, rather than the first buffer, and determine if it is within one segment duration away from currentTime. This is also in `closeToBufferedContent`
3. Finally we need to again get the `nextRange` rather than the first one to determine where we should seek.

src/playback-watcher.js
```diff
@@ -29,35 +29,54 @@ const timerCancelEvents = [
...
   }
 
-  // At least two to three segments worth of content should be buffered before there's a
-  // full enough buffer to consider taking any actions.
-  if (buffered.end(0) - buffered.start(0) < targetDuration * 2) {
-    return false;
+  // default to using buffered, but if we don't have one
+  // use video or audio buffered
+  if (!buffered || buffered.length === 0) {
+    buffered = videoBuffered || audioBuffered
```
gkatsev, line 65: oops forgot a semicolon
```suggestion
    buffered = videoBuffered || audioBuffered;
```

src/playback-watcher.js
```diff
@@ -29,35 +29,54 @@ const timerCancelEvents = [
  *        Options object
  * @param {TimeRange} options.buffered
  *        Current buffer
+ * @param {TimeRange} options.videoBuffered
+ *        Current buffer from the media sources video buffer
+  * @param {TimeRange} options.audioBuffered
+  *       Current buffer from the media sources audio buffer
```
gesinger, line 35:
```suggestion
 * @param {TimeRange} options.videoBuffered
 *        Current buffered from the media source's video buffer
 * @param {TimeRange} options.audioBuffered
 *       Current buffered from the media source's audio buffer
 ```

src/ranges.js
gesinger: Maybe it was a mistake? I think `end` is a method.

gkatsev (approved):

gesinger (approved):

PR #1195, opened 2021-09-08
fix: Only check for bad seeks after seeking and an append

brandonocasey:
## Description
In our current code we run `fixesBadSeeks_()` on tech waiting and our emulated tech waiting. This is problematic because at the time of tech waiting we won't have removed content from our segment loaders buffers. This means that any seek backwards within `targetDuration` will be skipped as a gap.

Example:
```
1. seek to 300 tech waiting fires, but there is nothing in the buffer at this time.
2. seek back to 290, tech waiting fires and we have buffer starting at 299s
3. playback watcher calls `fixedBadSeeks_` and determines that there is a gap from 290-299 as we have yet to append content.
4. playback watcher skips to 299
```

Previously we check to see if an append had happened by checking the buffer, but that isn't specific enough as large target durations will have issues as seen above. Now we use seeking events.

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -306,6 +306,10 @@ export default class PlaybackWatcher {
       return false;
     }
 
+    // TODO: It's possible that these seekable checks should be moved out of this function
+    // and into a function that runs on seekablechange. It's also possible that we only need
+    // afterSeekableWindow as the buffered check at the bottom is good enough to handle before
+    // seekable range.
     const seekable = this.seekable();
     const currentTime = this.tech_.currentTime();
     const isAfterSeekableRange = this.afterSeekableWindow_(
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -78,9 +78,9 @@ export default class PlaybackWatcher {
     /**
      * We check if a seek was into a gap through the following steps:
      * 1. We get a seeking event and we do not get a seeked event. This means that
-     *    we seeking was attempted but not completed.
+     *    a seek was attempted but not completed.
      * 2. We run `fixesBadSeeks_` on segment loader appends. This means that we already
-     *    removed everything from our buffer and appeded a segment, and should be ready
+     *    removed everything from our buffer and appended a segment, and should be ready
      *    to check for gaps.
      */
     const setSeekingHandlers = (fn) => {
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -22,64 +22,6 @@ const timerCancelEvents = [
 ];
 
 /**
- * Returns whether or not the current time should be considered close to buffered content,
- * taking into consideration whether there's enough buffered content for proper playback.
- *
- * @param {Object} options
- *        Options object
- * @param {TimeRange} options.buffered
- *        Current buffer
- * @param {TimeRange} options.videoBuffered
- *        Current buffered from the media source's video buffer
- * @param {TimeRange} options.audioBuffered
- *        Current buffered from the media source's audio buffer
- * @param {number} options.targetDuration
- *        The active playlist's target duration
- * @param {number} options.currentTime
- *        The current time of the player
- * @return {boolean}
- *         Whether the current time should be considered close to the buffer
- */
-export const closeToBufferedContent = ({ buffered, audioBuffered, videoBuffered, targetDuration, currentTime }) => {
-  const twoSegmentDurations = (targetDuration - Ranges.TIME_FUDGE_FACTOR) * 2;
-  const bufferedToCheck = [audioBuffered, videoBuffered];
-
-  for (let i = 0; i < bufferedToCheck.length; i++) {
-    // skip null buffered
-    if (!bufferedToCheck[i]) {
-      continue;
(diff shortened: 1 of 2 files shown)


src/playback-watcher.js
```diff
@@ -134,7 +75,38 @@ export default class PlaybackWatcher {
       this.tech_.on(['seeked', 'seeking'], loaderChecks[type].reset);
     });
 
-    this.tech_.on('seekablechanged', fixesBadSeeksHandler);
+    /**
+     * We check if a seek was into a gap through the following steps:
+     * 1. We get a seeking event and we do not get a seeked event. This means that
+     *    we seeking was attempted but not completed.
```
gesinger, line 81:
```suggestion
     *    a seek was attempted but not completed.
```

src/playback-watcher.js
```diff
@@ -134,7 +75,38 @@ export default class PlaybackWatcher {
...
+     * 2. We run `fixesBadSeeks_` on segment loader appends. This means that we already
+     *    removed everything from our buffer and appeded a segment, and should be ready
```
gesinger, line 83:
```suggestion
     *    removed everything from our buffer and appended a segment, and should be ready
```

src/playback-watcher.js
```diff
@@ -134,7 +75,38 @@ export default class PlaybackWatcher {
...
+     *    to check for gaps.
+     */
+    const setSeekingHandlers = (fn) => {
+      ['main', 'audio'].forEach((type) => {
+        mpc[`${type}SegmentLoader_`][fn]('appended', this.seekingAppendCheck_);
```
gesinger, line 88: Since this is now only run on appends, I think we will miss some cases where the seek was before or after the seekable range.

src/playback-watcher.js
```diff
@@ -377,27 +345,45 @@ export default class PlaybackWatcher {
...
 
-      this.tech_.setCurrentTime(seekTo);
-      return true;
+    const nextRange = Ranges.findNextRange(buffered, currentTime);
+
+    // we have appended enough content, but we don't have anything buffered
+    // to seek over the gap
+    if (nextRange.length === 0) {
+      return false;
     }
```
gesinger, line 377: Any specific reason for moving all of this out of a separate function?

test/playback-watcher.test.js
```diff
@@ -1097,6 +1068,145 @@ QUnit.test('jumps to correct range with gaps', function(assert) {
...
+    videoBuffered: () => buffered
+  });
+
+  currentTime = 40;
+  buffered = videojs.createTimeRanges([[19, 39], [41, 70]]);
+  this.player.tech_.trigger('seeking');
+  this.player.tech_.trigger('seeking');
+  mainSegmentLoader.trigger('appended');
+  assert.equal(seeks.length, 1, 'seeked');
+  assert.equal(seeks[0], 41.1, 'player seeked to the start of the closer buffer');
```
gesinger, line 1115:
```suggestion
  assert.equal(seeks[0], 41.1, 'player seeked to the start of the next buffer');
```

src/playback-watcher.js
brandonocasey: Maybe we can run on append or seekablechange?

src/playback-watcher.js
brandonocasey: After reducing the function to just the `twoSegmentDurations` check and the `nextRange` check it seemed silly to  call `findNextRange` in the helper function and then directly afterwards since we need access to it. So I think it made more sense to move the helper function in here.

gkatsev:
We have another test stream that this PR seems to fix https://github.com/videojs/http-streaming/issues/1202

harvister:
Hi, I was having a similar issue over at #1202.  I built and tested your most recent commit c6144ad58ffd4e05a47ed14ee020b3034c6c919c, as @gkatsev mentioned it could also fix my issue.

I can confirm that it does indeed fix the dramatic several second skip-back and/or infinite looping that I was experiencing, so thank you for that.

However, it still does not play as cleanly as all the other players I mentioned in my post.  After your fix, I now experience a stutter instead of the jump-back, basically a 1 second or less pause in both the video and the audio which is still quite disruptive for the user.  Much better though, thanks.

src/playback-watcher.js
brandonocasey: The consensus on this was that appended should work fine in this case.

gesinger (approved):

ashiskumarsahu:
Hi,
On which version of videojs, the fix is available?
I am using following:
https://vjs.zencdn.net/7.10.2/video.js
https://vjs.zencdn.net/7.17.1/video.js

**But issue still persist**.
Actually I am generating a playout stream having 1200 segments in playlist.
I am trying to play old chunks based on user plays the content with some timing calculation logic:

player.currentTime(900); //Like this

But if network fluctuates, it playing segments from backward(old chunks).

**How it should behave:**
suppose current loaded segment: 1080/seg-549.ts
when network switch it should load from 720/549.ts or  720/550.ts
Currently its loading 720/200.ts

Thank You

PR #1199, opened 2021-09-13
Fix/llhls query

### LINKED ISSUE #1198: LLHLS - playlist loader, querystring breaks uri
I think below line will break loading of playlist if uri already contain a "?".
https://github.com/videojs/http-streaming/blob/81600194030c22c283892e796d71700820478891/src/playlist-loader.js#L74

For me it ends up trying to load a playlist containing double "?" in uri, example below:
https://manifes.fastly.example.com/rendition.m3u8?cdn=fastly&expires=1653133440&live=1&skid=default&signature=ABC?_HLS_msn=8&_HLS_part=0

brandonocasey:
## Description
Fixes #1198

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -397,6 +397,11 @@ export default class PlaylistLoader extends EventTarget {
     this.customTagMappers = (vhsOptions && vhsOptions.customTagMappers) || [];
     this.experimentalLLHLS = (vhsOptions && vhsOptions.experimentalLLHLS) || false;
 
+    // force experimentalLLHLS for IE 11
+    if (videojs.browser.IE_VERSION) {
+      this.experimentalLLHLS = false;
+    }
+
     // initialize the loader state
     this.state = 'HAVE_NOTHING';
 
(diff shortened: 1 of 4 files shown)


test/loader-common.js
```diff
@@ -849,11 +849,12 @@ export const LoaderCommonFactory = ({
           this.clock.tick(1);
           return Promise.resolve();
         }).then(() => new Promise((resolve, reject) => {
-          loader.on('playlistupdate', () => {
-            this.clock.tick(1);
-            resolve();
-          });
           loader.on('trackinfo', () => {
+            loader.on('playlistupdate', () => {
```
brandonocasey, line 853: Fixes the race condition that currently causes ie 11 to fail

src/playlist-loader.js
```diff
@@ -62,19 +64,29 @@ const addLLHLSQueryDirectives = (uri, media) => {
...
   }
 
-  query.forEach(function(str, i) {
-    const symbol = i === 0 ? '?' : '&';
+  if (Object.keys(parameters).length) {
+    const parsedUri = new window.URL(uri);
 
-    uri += `${symbol}${str}`;
-  });
+    ['_HLS_skip', '_HLS_msn', '_HLS_part'].forEach(function(name) {
```
brandonocasey, line 80: I wanted to make sure that we keep the parameters in an expected order of `skip`, `msn`, and `part` as it makes reading the urls much easier.

gkatsev (approved):
I also tested it and it works.

PR #1201, opened 2021-09-15
Fix/more llhls fixes

brandonocasey:
## Description
Based on #1195 for now, but we can pull it out if we want as only one of the fixes requires it.


src/master-playlist-controller.js
```diff
@@ -1431,7 +1431,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
   onSyncInfoUpdate_() {
     let audioSeekable;
 
-    if (!this.masterPlaylistLoader_) {
+    if (!this.masterPlaylistLoader_ || this.sourceUpdater_.hasCreatedSourceBuffers()) {
```
brandonocasey, line 1434: This was causing us to:
1. As soon as audio/video segment loader fires this we come in here
2. Since only one of the loaders is actually ready at this point, we get seekable from just that one
3. We seek to the end of this invalid seekable range
4. We come back in here again and this time we get an intersection between the two and a valid seekable
5. We seek to the valid seekable end via playback watcher
6. This causes content to be removes and things to be downloaded/appended again.

src/playback-watcher.js
```diff
@@ -350,7 +350,9 @@ export default class PlaybackWatcher {
 
     // verify that at least two segment durations have been
     // appended before checking for a gap.
-    const twoSegmentDurations = (this.media().targetDuration - Ranges.TIME_FUDGE_FACTOR) * 2;
+    const targetDuration = this.media().partTargetDuration || this.media().targetDuration;
```
brandonocasey, line 353: This helps skip gaps faster in llhls, mostly this is helping with our bad segment choice at the start of a playlist. It is likely something we want to do anyway though, as we don't want to wait very long to skip gaps in llhls.

src/playlist-loader.js
```diff
@@ -257,7 +257,8 @@ const getAllSegments = function(media) {
 export const isPlaylistUnchanged = (a, b) => a === b ||
   (a.segments && b.segments && a.segments.length === b.segments.length &&
    a.endList === b.endList &&
-   a.mediaSequence === b.mediaSequence);
+   a.mediaSequence === b.mediaSequence &&
+   (a.preloadSegment && b.preloadSegment && a.preloadSegment === b.preloadSegment));
```
brandonocasey, line 261: Without this line we can trigger `playlistunchaned` for llhls playlists that have changed.

src/playlist.js
```diff
@@ -117,7 +117,19 @@ const backwardDuration = function(playlist, endSequence) {
       return { result: result + segment.end, precise: true };
     }
 
-    result += segment.duration;
+    if (segment.preload) {
```
brandonocasey, line 120: This prevents us from considering the full segment duration for preload segments that do not have all of their parts yet.

src/playlist.js
```diff
@@ -321,19 +333,19 @@ export const playlistEnd = function(playlist, expired, useSafeLiveEnd, liveEdgeP
 
   expired = expired || 0;
 
-  let lastSegmentTime = intervalDuration(
+  let lastSegmentEndTime = intervalDuration(
```
brandonocasey, line 348: A more descriptive name as it isn't the start time, but the end time we are getting here.

src/playlist.js
```diff
@@ -407,7 +419,12 @@ export const getMediaInfoForTime = function({
     break;
   }
 
-  if (time < 0) {
+  // for normal playlists we can compare against zero
+  // but we have to be much more specific for LLHLS.
+  // as even half a second can cause us to guess the wrong segment.
+  const zeroOrPartTarget = playlist.partTargetDuration || 0;
```
brandonocasey, line 437: WIthout this fix we end up picking a part ahead of currentTime quite often, presumably because parts are so small and it's hard to pick one exactly like this.

src/segment-loader.js
```diff
@@ -1984,7 +1984,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.setTimeMapping_(segmentInfo.timeline);
 
     // for tracking overall stats
-    this.updateMediaSecondsLoaded_(segmentInfo.segment);
+    this.updateMediaSecondsLoaded_(segmentInfo.part || segmentInfo.segment);
```
brandonocasey, line 1987: stats fix

src/sync-controller.js
```diff
@@ -46,35 +46,33 @@ export const syncPointStrategies = [
...
-          }
-          const distance = Math.abs(currentTime - start);
-
-          // Once the distance begins to increase, or if distance is 0, we have passed
-          // currentTime and can stop looking for better candidates
-          if (lastDistance !== null && (distance === 0 || lastDistance < distance)) {
-            break;
+        // take part duration into account.
+        if (segment.parts && typeof partAndSegment.partIndex === 'number') {
+          for (let z = 0; z < partAndSegment.partIndex - 1; z++) {
```
brandonocasey, line 58: This function is the same except for the short circuit that I added and this line which only considers durations before the current part index, because we are trying to find what part **starts** at a time.

src/playlist-loader.js
```diff
@@ -662,6 +665,8 @@ export default class PlaylistLoader extends EventTarget {
       this.trigger('mediachanging');
     }
 
+    this.pendingMedia_ = playlist;
```
brandonocasey, line 668: Added this to prevent switching to the same playlist twice at the start of playback.

src/master-playlist-controller.js
```diff
@@ -729,15 +734,17 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
   shouldSwitchToMedia_(nextPlaylist) {
-    const currentPlaylist = this.masterPlaylistLoader_.media();
+    const currentPlaylist = this.masterPlaylistLoader_.media() ||
+      this.masterPlaylistLoader_.pendingMedia_;
     const buffered = this.tech_.buffered();
-    const forwardBuffer = buffered.length ?
-      buffered.end(buffered.length - 1) - this.tech_.currentTime() : 0;
-
+    const currentTime = this.tech_.currentTime();
+    const forwardBuffer = Ranges.timeAheadOf(buffered, currentTime);
```
brandonocasey, line 741: Use `timeAheadOf` so that we consider gaps in our forward buffer and don't try to switch when we have enough buffer but have gaps.

src/master-playlist-controller.js
```diff
@@ -729,15 +734,17 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
     const bufferLowWaterLine = this.bufferLowWaterLine();
     const bufferHighWaterLine = this.bufferHighWaterLine();
+    const isBuffered = Boolean(Ranges.findRange(buffered, currentTime).length);
 
     return shouldSwitchToMedia({
+      isBuffered,
```
brandonocasey, line 747: We don't really want to switch at the start of playback for live streams before we have actually started. It almost doubles are time to first playback.

src/playback-watcher.js
```diff
@@ -347,10 +347,11 @@ export default class PlaybackWatcher {
     const buffered = this.tech_.buffered();
     const audioBuffered = sourceUpdater.audioBuffer ? sourceUpdater.audioBuffered() : null;
     const videoBuffered = sourceUpdater.videoBuffer ? sourceUpdater.videoBuffered() : null;
+    const media = this.media();
 
-    // verify that at least two segment durations have been
+    // verify that at least two segment durations or one part duration have been
     // appended before checking for a gap.
-    const twoSegmentDurations = (this.media().targetDuration - Ranges.TIME_FUDGE_FACTOR) * 2;
+    const durations = media.partTargetDuration ? media.partTargetDuration : media.targetDuration * 2;
```
brandonocasey, line 354: I lowered this to one part target duration. parts are so tiny that its almost impossible not be be off by a few ms.

src/sync-controller.js
```diff
@@ -46,35 +46,33 @@ export const syncPointStrategies = [
         const datetimeMapping =
           syncController.timelineToDatetimeMappings[segment.timeline];
 
-        if (!datetimeMapping) {
+        if (!datetimeMapping || !segment.dateTimeObject) {
```
brandonocasey, line 49: short circuit earlier

package.json
```diff
@@ -62,7 +62,7 @@
     "m3u8-parser": "4.7.0",
     "mpd-parser": "0.19.0",
     "mux.js": "5.13.0",
-    "video.js": "^6 || ^7"
+    "video.js": "^7.15.4"
```
gkatsev, line 65: we should revert this change unless we really do depend on something that requires 7.15.4 at this point.

brandonocasey: Yeah I reverted it in the merge that I had to do. I updated video.js in the `package-lock.json` though.

gkatsev: should've waited for the rebase then haha

src/segment-loader.js
```diff
@@ -160,6 +159,10 @@ const segmentInfoString = (segmentInfo) => {
     selection = 'getSyncSegmentCandidate (isSyncRequest)';
   }
 
+  if (segmentInfo.independent) {
+    selection += ` ${segmentInfo.independent}`;
```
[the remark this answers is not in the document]
brandonocasey: We don't have any tests for it now, but I can write some easily enough I think.

src/segment-loader.js
```diff
@@ -1026,7 +1029,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       if (this.mediaIndex !== null) {
         // we must "resync" the segment loader when we switch renditions and
         // the segment loader is already synced to the previous rendition
-        this.resyncLoader();
+        this.resetLoader();
```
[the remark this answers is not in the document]
gkatsev: Might be nice to do this only for live streams, though, if @gesinger thinks it's fine to always do it, I won't have any objections.

src/playback-watcher.js
```diff
@@ -351,10 +351,15 @@ export default class PlaybackWatcher {
     const buffered = this.tech_.buffered();
     const audioBuffered = sourceUpdater.audioBuffer ? sourceUpdater.audioBuffered() : null;
     const videoBuffered = sourceUpdater.videoBuffer ? sourceUpdater.videoBuffered() : null;
+    const media = this.media();
+
+    // verify that at least two segment durations or one part duration have been
+    // appended before checking for a gap.
+    const durations = media.partTargetDuration ? media.partTargetDuration :
```
gesinger, line 358: Maybe we can rename this to something like `minAppendedDuration`.

src/playlist-loader.js
gesinger: Where does it get used?

src/segment-loader.js
```diff
@@ -1024,9 +1027,15 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     if (!oldPlaylist || oldPlaylist.uri !== newPlaylist.uri) {
       if (this.mediaIndex !== null) {
-        // we must "resync" the segment loader when we switch renditions and
+        // we must reset/resync the segment loader when we switch renditions and
         // the segment loader is already synced to the previous rendition
-        this.resyncLoader();
+
+        // for live streams reset, for vod resync
```
gesinger, line 1033: It may be worth adding why we reset for live vs resync for VOD.

test/loader-common.js
```diff
@@ -835,6 +835,68 @@ export const LoaderCommonFactory = ({
...
+        const newPlaylist = playlistWithDuration(50, {
+          mediaSequence: 0,
+          endList: true
+        });
+
+        newPlaylist.uri = 'playlist2.m3u8';
+
+        loader.playlist(newPlaylist);
+
+        assert.true(resyncCalled, 'resync was called');
```
gesinger, line 894: Since `reset` also calls `resync`, we'll want to check `reset` as well.

test/loader-common.js
```diff
@@ -1275,6 +1337,38 @@ export const LoaderCommonFactory = ({
...
+      loader.load();
+      // in a real scenario this probably won't happen
+      // but it makes which segment/part is chosen easier.
+      loader.mediaIndex = 4;
+      loader.partIndex = 1;
+
+      playlist.segments[4].parts[1].independent = true;
+      const segmentInfo = loader.chooseNextRequest_();
+
+      assert.equal(segmentInfo.partIndex, 1, 'previous part');
```
gesinger, line 1358: Doesn't this match the partIndex above?

test/master-playlist-controller.test.js
```diff
@@ -6085,6 +6085,25 @@ QUnit.test('false without nextPlaylist', function(assert) {
   this.env.log.warn.callCount = 0;
 });
 
+QUnit.test('false if llhls playlist and video not buffered', function(assert) {
+  const mpc = this.masterPlaylistController;
+
+  mpc.masterPlaylistLoader_.media = () => ({id: 'foo', endList: false, partTargetDuration: 5});
+  const nextPlaylist = {id: 'bar', endList: false, partTargetDuration: 5};
+
+  assert.notOk(mpc.shouldSwitchToMedia_(nextPlaylist), 'should switch without currentPlaylist');
```
gesinger, line 6094: Should this be "shouldn't switch when video not buffered"?

test/master-playlist-controller.test.js
```diff
@@ -6085,6 +6085,25 @@ QUnit.test('false without nextPlaylist', function(assert) {
...
+});
+
+QUnit.test('true if llhls playlist and video buffered', function(assert) {
+  const mpc = this.masterPlaylistController;
+
+  mpc.tech_.buffered = () => videojs.createTimeRange([[0, 10]]);
+  mpc.masterPlaylistLoader_.media = () => ({id: 'foo', endList: false, partTargetDuration: 5});
+  const nextPlaylist = {id: 'bar', endList: false, partTargetDuration: 5};
+
+  assert.ok(mpc.shouldSwitchToMedia_(nextPlaylist), 'should switch without currentPlaylist');
```
gesinger, line 6104: There should be a currentPlaylists here I imagine?

src/playlist-loader.js
brandonocasey: `masterPlaylistLoader_.shouldSwitchToMedia_()`

test/loader-common.js
brandonocasey: it does, but we also don't have anything buffered so it tries to choose sepmentIndex 4, partIndex 2 but that isn't independent so we choose the current partIndex. Let me see if I can find something better than this though.

src/master-playlist-controller.js
```diff
@@ -1441,6 +1445,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
   onSyncInfoUpdate_() {
     let audioSeekable;
 
+    // If we have two source buffers and only one is created. The seekable range will be incorrect
+    // we should wait until all source buffers are created.
```
gesinger, line 1449:
```suggestion
    // If we have two source buffers and only one is created then the seekable range will be incorrect.
    // We should wait until all source buffers are created.
```

PR #1204, opened 2021-09-16
Pr for media sequence

jontsnz:
## Description

The browser tab running video.js would show a "Page Unresponsive" dialog in Chrome when the EXT-X-MEDIA-SEQUENCE tag increased by a large number between subsequent m3u8 files. This happened because the saveExpiredSegmentInfo function was iterating over this large number.

This behaviour was raised in [issue 1197](https://github.com/videojs/http-streaming/issues/1197)

## Proposed Solution

The proposed solution is to avoid the browser becoming unresponsive by only iterating over the media sequence difference if it is within a reasonable threshold. 

To determine a reasonable threshold, I looked at Apple's ["Live Playlist (Sliding Window) Construction" page](https://developer.apple.com/documentation/http_live_streaming/example_playlists_for_http_live_streaming/live_playlist_sliding_window_construction) which states:

> The EXT-X-MEDIA-SEQUENCE tag value must be incremented by 1 for every media URI that's removed from the playlist file.

Based on this, I have used a threshold of 1000.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -6,6 +6,10 @@ import {sumDurations, getPartsAndSegments} from './playlist';
 import videojs from 'video.js';
 import logger from './util/logger';
 
+// The maximum gap allowed between two media sequence tags when trying to
+// synchronize expired playlist segments.
+const MAX_MEDIA_SEQUENCE_DIFF_FOR_SYNC = 1000;
+
 export const syncPointStrategies = [
   // Stategy "VOD": Handle the VOD-case where the sync-point is *always*
   //                the equivalence display-time 0 === segment-index 0
@@ -363,6 +367,11 @@ export default class SyncController extends videojs.EventTarget {
    */
   saveExpiredSegmentInfo(oldPlaylist, newPlaylist) {
     const mediaSequenceDiff = newPlaylist.mediaSequence - oldPlaylist.mediaSequence;
+
+    // Ignore large media sequence gaps
+    if (mediaSequenceDiff > MAX_MEDIA_SEQUENCE_DIFF_FOR_SYNC) {
+      return;
+    }
 
     // When a segment expires from the playlist and it has a start time
     // save that information as a possible sync-point reference in future
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -370,6 +370,7 @@ export default class SyncController extends videojs.EventTarget {
 
     // Ignore large media sequence gaps
     if (mediaSequenceDiff > MAX_MEDIA_SEQUENCE_DIFF_FOR_SYNC) {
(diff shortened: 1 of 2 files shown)


src/sync-controller.js
```diff
@@ -364,6 +368,12 @@ export default class SyncController extends videojs.EventTarget {
   saveExpiredSegmentInfo(oldPlaylist, newPlaylist) {
     const mediaSequenceDiff = newPlaylist.mediaSequence - oldPlaylist.mediaSequence;
 
+    // Ignore large media sequence gaps
+    if (mediaSequenceDiff > MAX_MEDIA_SEQUENCE_DIFF_FOR_SYNC) {
+      videojs.log.warn('Ignoring sync for large media sequence gap: ' + mediaSequenceDiff);
```
brandonocasey, line 373:
```suggestion
      videojs.log.warn(`Not saving expired segment info. Media sequence gap ${mediaSequenceDiff} is too large.`);
```

brandonocasey (commented):
Thanks for the pull request! I think 1000 is sufficiently large here. I just have a small comment on the log message. Do you think you might have some time to write a test?

src/sync-controller.js
```diff
@@ -6,6 +6,10 @@ import {sumDurations, getPartsAndSegments} from './playlist';
 import videojs from 'video.js';
 import logger from './util/logger';
 
+// The maximum gap allowed between two media sequence tags when trying to
+// synchronize expired playlist segments.
+const MAX_MEDIA_SEQUENCE_DIFF_FOR_SYNC = 1000;
```
gkatsev, line 11: @brandonocasey do you think it would be worth having this be in config.js or just have it be here and if anyone complains we can move it to config.js?

brandonocasey: I think we can just leave it here. I highly doubt that we will need to have it configured.

jontsnz: Have accepted the suggested message change, and added a unit test. Let me know if there's anything else I can do.

gkatsev (dismissed):

src/sync-controller.js
brandonocasey:
```suggestion
// the max media sequence diff is 48 hours of live stream 
// content with two second segments. Anything larger than that
// will likely be invalid.
const MAX_MEDIA_SEQUENCE_DIFF_FOR_SYNC = 86400;
```

One of our colleagues mentioned that 1000 might not be big enough for long running live streams. We all agreed on a number that we were happy with. Sorry about that.

brandonocasey (changes requested):
One last change brought up internally.

gesinger (approved):

gesinger (approved):

PR #1217, opened 2021-10-25
fix: mp4 sources that use bigint numbers

brandonocasey:
## Description
* https://github.com/videojs/mux.js/pull/383
* https://github.com/videojs/mpd-parser/pull/135

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -424,7 +424,6 @@ export const mediaDuration = (timingInfos) => {
       duration = window.BigInt(end) - window.BigInt(start);
     } else if (typeof start === 'number' && typeof end === 'number') {
       duration = end - start;
-
     }
 
     if (typeof duration !== 'undefined' && duration > maxDuration) {
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -407,19 +407,32 @@ export const shouldWaitForTimelineChange = ({
   return false;
 };
 
-export const mediaDuration = (audioTimingInfo, videoTimingInfo) => {
-  const audioDuration =
-    audioTimingInfo &&
-    typeof audioTimingInfo.start === 'number' &&
-    typeof audioTimingInfo.end === 'number' ?
-      audioTimingInfo.end - audioTimingInfo.start : 0;
-  const videoDuration =
-    videoTimingInfo &&
-    typeof videoTimingInfo.start === 'number' &&
-    typeof videoTimingInfo.end === 'number' ?
-      videoTimingInfo.end - videoTimingInfo.start : 0;
+export const mediaDuration = (timingInfos) => {
+  let maxDuration = 0;
 
-  return Math.max(audioDuration, videoDuration);
+  ['video', 'audio'].forEach(function(type) {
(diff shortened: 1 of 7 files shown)


src/segment-loader.js
```diff
@@ -407,19 +407,32 @@ export const shouldWaitForTimelineChange = ({
...
+    }
+    const {start, end} = typeTimingInfo;
+    let duration;
+
+    // eslint-disable-next-line
+    if (typeof start === 'bigint' || typeof end === 'bigint') {
+      duration = window.BigInt(end) - window.BigInt(start);
+    } else if (typeof start === 'number' && typeof end === 'number') {
+      duration = end - start;
+
```
gesinger, line 427:
```suggestion
```

src/segment-loader.js
```diff
@@ -407,19 +407,32 @@ export const shouldWaitForTimelineChange = ({
...
+    }
+
+    if (typeof duration !== 'undefined' && duration > maxDuration) {
+      maxDuration = duration;
+    }
+  });
+
+  return maxDuration;
```
gesinger, line 435: Do we want to check to see if it's a `bigint` here or smaller than the MAX_SAFE_INTEGER to convert back? I expect most segments shouldn't have such a long duration.

src/segment-loader.js
```diff
@@ -407,19 +407,32 @@ export const shouldWaitForTimelineChange = ({
...
+  ['video', 'audio'].forEach(function(type) {
+    const typeTimingInfo = timingInfos[`${type}TimingInfo`];
+
+    if (!typeTimingInfo) {
+      return;
+    }
+    const {start, end} = typeTimingInfo;
+    let duration;
+
+    // eslint-disable-next-line
```
gkatsev, line 422: is this because our version of eslint doesn't know that `typeof` can return `bigint`?

gkatsev: Do you know if later eslint versions support it? If so, we should make a task for us to update eslint so that we can remove this comment (I see it below as well)

src/segment-loader.js
```diff
@@ -407,19 +407,32 @@ export const shouldWaitForTimelineChange = ({
...
+    if (typeof start === 'bigint' || typeof end === 'bigint') {
+      duration = window.BigInt(end) - window.BigInt(start);
```
gkatsev, line 424: each item needs to be wrapped in `BigInt()` to make sure that JS runs it in BigInt mode and doesn't try to run it in regular int mode?

src/segment-loader.js
brandonocasey: I updated videojs-standard to support it, I think I can update it as a part of this pull request and remove this line.

src/segment-loader.js
brandonocasey: That is correct. JavaScript throws an error if you try to do any math between `BigInt` and a normal `Number`. Once you have one `BigInt` you have to stick with it.

gesinger (dismissed):

gkatsev (approved):

gkatsev (approved):

PR #1218, opened 2021-10-29
Add an option to calculate bandwidth by using Chromium's networkInfo API

evanfarina:
## Description
We've found that the player's bandwidth estimations provide lower-than-expected values when a particular segment is encoded with a low bitrate relative to other segments in its lane. This is a fairly common phenomenon when a live stream is using a variable bitrate encoding strategy and is broadcasting a static scene (e.g a slide in a slideshow). We believe that this issue is further exacerbated by two factors:

1. Segment size - Vhs estimates bandwidth by measuring how many bits were downloaded in a segment and how long, in seconds, it took to download those bits. This estimate factors in network-layer overhead which means that as the number of bits in a segment decreases (either due to bitrate or segment size), the impact that network overhead has on the bandwidth estimate increases
2. Bandwidth Variance config - Vhs uses a `BANDWIDTH_VARIANCE` value as a multiplier for the advertised bandwidth value in each lane. The filter calculation looks like `lane bandwidth * bandwidth variance < calculatedBandwidth`. When `calculatedBandwidth` is low due to aforementioned reasons, the bandwidth variance default makes it more likely that higher-quality lanes will be filtered out by the player

## Specific Changes proposed
Introduce an experimental option that allows the player to leverage Chromium's [NetworkInfo.downlink](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink) API to estimate bandwidth. Note: This API is not available in Firefox or Safari. When enabled, the player will take the max value between its own bandwidth estimation and that provided by NetworkInformation.downlink.

Please see attached screenshots which demonstrate how the player now selects the higher-quality rendition during static scenes. [This demo](http://localhost:9999/?debug=false&autoplay=false&muted=false&fluid=false&minified=false&sync-workers=false&liveui=true&llhls=false&url=https%3A%2F%2Flivectorprodmedia11-usw22.licdn.com%2F0dbd58e9-9111-4f66-a1aa-a4013fc974ef%2FL565f16d44f01c2c000-livemanifest.ism%2Fmanifest(format%3Dm3u8-aapl-v3)&type=application%2Fx-mpegURL&keysystems=&buffer-water=false&exact-manifest-timings=false&pixel-diff-selector=false&network-info=true&override-native=true&preload=auto&mirror-source=true) uses 2-second segments.

Screen Shot 2021-10-29 at 2 34 06 PM

Screen Shot 2021-10-29 at 2 33 27 PM

Screen Shot 2021-10-29 at 2 05 13 PM

Screen Shot 2021-10-29 at 2 32 46 PM
## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Video.js Compatibility: 6.0, 7.0
       - [cacheEncryptionKeys](#cacheencryptionkeys)
       - [handlePartialData](#handlepartialdata)
       - [liveRangeSafeTimeDelta](#liverangesafetimedelta)
+      - [useNetworkInformationApi](#usenetworkinformationapi)
       - [captionServices](#captionservices)
         - [Format](#format)
         - [Example](#example)
@@ -472,6 +473,11 @@ This option defaults to `false`.
 * Type: `number`,
 * Default: [`SAFE_TIME_DELTA`](https://github.com/videojs/http-streaming/blob/e7cb63af010779108336eddb5c8fd138d6390e95/src/ranges.js#L17)
 * Allow to  re-define length (in seconds) of time delta when you compare current time and the end of the buffered range.
+
+##### useNetworkInformationApi
+* Type: `boolean`,
+* Default: `false`
+* Use [window.networkInformation.downlink](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink) to estimate the network's bandwidth
 
 ##### captionServices
 * Type: `object`
--- a/README.md
+++ b/README.md
@@ -477,7 +477,7 @@ This option defaults to `false`.
 ##### useNetworkInformationApi
 * Type: `boolean`,
 * Default: `false`
-* Use [window.networkInformation.downlink](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink) to estimate the network's bandwidth
+* Use [window.networkInformation.downlink](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink) to estimate the network's bandwidth. Per mdn, _The value is never greater than 10 Mbps, as a non-standard anti-fingerprinting measure_. Given this, if bandwidth estimates from both the player and networkInfo are >= 10 Mbps, the player will use the larger of the two values as its bandwidth estimate.
 
 ##### captionServices
 * Type: `object`
--- a/index.html
+++ b/index.html
@@ -168,7 +168,7 @@
 
         <div class="form-check">
           <input id=network-info type="checkbox" class="form-check-input">
-          <label class="form-check-label" for="network-info">[EXPERIMENTAL] Use networkInfo API for bandwidth estimations (reloads player)</label>
+          <label class="form-check-label" for="network-info">Use networkInfo API for bandwidth estimations (reloads player)</label>
         </div>
 
         <div class="form-check">
--- a/index.html
+++ b/index.html
@@ -167,6 +167,11 @@
         </div>
 
         <div class="form-check">
+          <input id=network-info type="checkbox" class="form-check-input">
+          <label class="form-check-label" for="network-info">[EXPERIMENTAL] Use networkInfo API for bandwidth estimations (reloads player)</label>
+        </div>
+
+        <div class="form-check">
           <input id=override-native type="checkbox" class="form-check-input" checked>
           <label class="form-check-label" for="override-native">Override Native (reloads player)</label>
         </div>
--- a/index.html
+++ b/index.html
@@ -147,6 +147,11 @@
         </div>
 
         <div class="form-check">
+          <input id=network-info type="checkbox" class="form-check-input">
+          <label class="form-check-label" for="network-info">Use networkInfo API for bandwidth estimations (reloads player)</label>
+        </div>
+
+        <div class="form-check">
           <input id=llhls type="checkbox" class="form-check-input">
           <label class="form-check-label" for="llhls">[EXPERIMENTAL] Enables support for ll-hls (reloads player)</label>
         </div>
@@ -164,11 +169,6 @@
         <div class="form-check">
           <input id=pixel-diff-selector type="checkbox" class="form-check-input">
           <label class="form-check-label" for="pixel-diff-selector">[EXPERIMENTAL] Use the Pixel difference resolution selector (reloads player)</label>
-        </div>
-
-        <div class="form-check">
-          <input id=network-info type="checkbox" class="form-check-input">
-          <label class="form-check-label" for="network-info">Use networkInfo API for bandwidth estimations (reloads player)</label>
         </div>
 
         <div class="form-check">
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -791,22 +791,27 @@ class VhsHandler extends Component {
       },
       bandwidth: {
         get() {
-          let bandwidthEst = this.masterPlaylistController_.mainSegmentLoader_.bandwidth;
+          let playerBandwidthEst = this.masterPlaylistController_.mainSegmentLoader_.bandwidth;
 
           const networkInformation = window.navigator.connection || window.navigator.mozConnection || window.navigator.webkitConnection;
+          const tenMbpsAsBitsPerSecond = 10e6;
 
           if (this.options_.useNetworkInformationApi && networkInformation) {
             // downlink returns Mbps
             // https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink
-            const effectiveBandwidthBitsPerSec = networkInformation.downlink * 1000 * 1000;
+            const networkInfoBandwidthEstBitsPerSec = networkInformation.downlink * 1000 * 1000;
 
-            // downlink maxes out at 10 Mbps. In the event that the player estimates a bandwidth
-            // greater than 10 Mbps, use the larger value to ensure that high quality streams are not
-            // filtered out.
-            bandwidthEst = Math.max(bandwidthEst, effectiveBandwidthBitsPerSec);
+            // downlink maxes out at 10 Mbps. In the event that both networkInformationApi and the player
+            // estimate a bandwidth greater than 10 Mbps, use the larger of the two estimates to ensure that
+            // high quality streams are not filtered out.
+            if (networkInfoBandwidthEstBitsPerSec >= tenMbpsAsBitsPerSecond && playerBandwidthEst >= tenMbpsAsBitsPerSecond) {
+              playerBandwidthEst = Math.max(playerBandwidthEst, networkInfoBandwidthEstBitsPerSec);
+            } else {
+              playerBandwidthEst = networkInfoBandwidthEstBitsPerSec;
+            }
           }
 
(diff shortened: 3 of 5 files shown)


src/videojs-http-streaming.js
```diff
@@ -563,6 +563,7 @@ class VhsHandler extends Component {
     this.source_ = source;
     this.stats = {};
     this.ignoreNextSeekingEvent_ = false;
+    this.effectiveBandwidthInBits_ = 0;
```
jiangyurong609, line 566: rename as effectiveBandwidthInBitPerSec

src/videojs-http-streaming.js
```diff
@@ -617,6 +618,20 @@ class VhsHandler extends Component {
...
+      if (this.networkInformation) {
+        this.networkInformation.addEventListener('change', () => this.setNetworkInformationStats_());
+
+        this.setNetworkInformationStats_();
+      }
+    }
+  }
+
+  setNetworkInformationStats_() {
+    this.effectiveBandwidthInBits_ = this.networkInformation.downlink * 1000 * 1000;
```
jiangyurong609, line 634: Suggest add a comment linking the API doc and original unit megabit/s

src/videojs-http-streaming.js
```diff
@@ -813,7 +830,16 @@ class VhsHandler extends Component {
...
+          let bandwidthEst = this.bandwidth;
+
+          // NetworkInfo.downlink maxes out at 10Mb/s. In the event that the player estimates a bandwidth
+          // greater than 10Mb, use the larger value to ensure that high quality streams are not
+          // accidentally filtered out
+          if (this.options_.useNetworkInformation) {
+            bandwidthEst = Math.max(bandwidthEst, this.effectiveBandwidthInBits_);
+          }
+
+          const invBandwidth = 1 / (bandwidthEst || 1);
```
jiangyurong609, line 842: the invBandwidth can be 0? why not 1.0 / (bandwidthEst || 1)?

evanfarina: Caught up with @jiangyurong609. The concern here was that the value would floor to zero, but js will maintain a floating point here. Resolving the comment.

index.html
```diff
@@ -166,6 +166,11 @@
           <label class="form-check-label" for="pixel-diff-selector">[EXPERIMENTAL] Use the Pixel difference resolution selector (reloads player)</label>
         </div>
 
+        <div class="form-check">
+          <input id=network-info type="checkbox" class="form-check-input">
+          <label class="form-check-label" for="network-info">[EXPERIMENTAL] Use networkInfo API for bandwidth estimations (reloads player)</label>
```
gesinger, line 171: We probably don't need `EXPERIMENTAL` here, unless we want to prefix the option.

src/videojs-http-streaming.js
```diff
@@ -813,7 +830,16 @@ class VhsHandler extends Component {
        */
       systemBandwidth: {
```
gesinger, line 831: We may also want to consider updating `bandwidth` above.

src/videojs-http-streaming.js
```diff
@@ -682,6 +698,7 @@ class VhsHandler extends Component {
       'experimentalBufferBasedABR',
       'liveRangeSafeTimeDelta',
       'experimentalLLHLS',
+      'useNetworkInformation',
```
gesinger, line 701: I think `useNetworkInformationApi`, although a bit more verbose, directly points to the API we're referencing.

We should also add an entry to the README.

test/videojs-http-streaming.test.js
```diff
@@ -1160,6 +1160,99 @@ QUnit.test(
...
+      src: 'manifest/master.m3u8',
+      type: 'application/vnd.apple.mpegurl'
+    });
+
+    this.clock.tick(1);
+
+    assert.strictEqual(
+      this.player.tech_.vhs.effectiveBandwidthInBits_,
+      0,
+      'effectiveBandwidthInBits_ defaults to 0'
```
gesinger, line 1222: Would it be worth having a default value? Or reading the initial value?

src/videojs-http-streaming.js
```diff
@@ -617,6 +618,20 @@ class VhsHandler extends Component {
     });
 
     this.on(this.tech_, 'play', this.play);
+
+    if (this.options_.useNetworkInformation) {
+      this.networkInformation = window.navigator.connection || window.navigator.mozConnection || window.navigator.webkitConnection;
+
+      if (this.networkInformation) {
+        this.networkInformation.addEventListener('change', () => this.setNetworkInformationStats_());
```
gesinger, line 626: Since `downlink` is a property on `navigator.connection`, I think we may be better off accessing the property instead of adding a new property and updating it on each `change`.

gesinger (commented):
Thank you for the PR @evanfarina ! A few comments for consideration.

src/videojs-http-streaming.js
evanfarina: Agree. Was originally planning on storing more properties to support future use-cases with this API, but no need to get ahead of ourselves.

src/videojs-http-streaming.js
evanfarina: Added

src/videojs-http-streaming.js
evanfarina: SGTM

test/videojs-http-streaming.test.js
evanfarina: downlink will provide a default value. From the docs: "In the absence of recent bandwidth measurement data, the attribute value is determined by the properties of the underlying connection technology." Given this, I don't think there is a need for us to test for the default. Also, we'll always choose the larger value between Vhs' bandwidth estimate and the estimate provided by downlink, so we should be safe even if downlink estimates a low initial value for some reason

src/videojs-http-streaming.js
```diff
@@ -795,7 +795,19 @@ class VhsHandler extends Component {
...
+
+          // NetworkInfo.downlink maxes out at 10 Mbps. In the event that the player estimates a bandwidth
+          // greater than 10 Mbps, use the larger value to ensure that high quality streams are not
+          // accidentally filtered out
+          if (this.options_.useNetworkInformationApi && this.networkInformation) {
+            // Downlink property returns Mbps https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink
+            const effectiveBandwidthBitsPerSec = this.networkInformation.downlink * 1000 * 1000;
+
+            bandwidthEst = Math.max(bandwidthEst, effectiveBandwidthBitsPerSec);
+          }
```
gesinger, line 808: I think we can remove the `this.networkInformation` property and create a variable for it here to avoid creating another private property and keep everything in one spot.

```suggestion
          const  networkInformation = window.navigator.connection || window.navigator.mozConnection || window.navigator.webkitConnection;

          if (this.options_.useNetworkInformationApi && networkInformation) {
            // downlink returns Mbps
            // https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink
            const effectiveBandwidthBitsPerSec = this.networkInformation.downlink * 1000 * 1000;

            // downlink maxes out at 10 Mbps. In the event that the player estimates a bandwidth
            // greater than 10 Mbps, use the larger value to ensure that high quality streams are not
            // filtered out.
            bandwidthEst = Math.max(bandwidthEst, effectiveBandwidthBitsPerSec);
          }
```

index.html
```diff
@@ -166,6 +166,11 @@
           <label class="form-check-label" for="pixel-diff-selector">[EXPERIMENTAL] Use the Pixel difference resolution selector (reloads player)</label>
         </div>
 
+        <div class="form-check">
```
gkatsev, line 169: can this be moved above the experimental items?

src/videojs-http-streaming.js
```diff
@@ -789,7 +791,22 @@ class VhsHandler extends Component {
...
+
+          if (this.options_.useNetworkInformationApi && networkInformation) {
+            // downlink returns Mbps
+            // https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink
+            const effectiveBandwidthBitsPerSec = networkInformation.downlink * 1000 * 1000;
+
+            // downlink maxes out at 10 Mbps. In the event that the player estimates a bandwidth
+            // greater than 10 Mbps, use the larger value to ensure that high quality streams are not
+            // filtered out.
+            bandwidthEst = Math.max(bandwidthEst, effectiveBandwidthBitsPerSec);
```
gkatsev, line 806: you mentioned that you noticed that `downlink` was also unreliable when throttling down? Not sure if it's worth mentioning that or not.

README.md
```diff
@@ -473,6 +474,11 @@ This option defaults to `false`.
 * Default: [`SAFE_TIME_DELTA`](https://github.com/videojs/http-streaming/blob/e7cb63af010779108336eddb5c8fd138d6390e95/src/ranges.js#L17)
 * Allow to  re-define length (in seconds) of time delta when you compare current time and the end of the buffered range.
 
+##### useNetworkInformationApi
```
gkatsev, line 477: thanks for adding docs!

gkatsev (commented):
It's nice that it's pretty simple and self-contained!

src/videojs-http-streaming.js
evanfarina: The opposite:) I found that it’s more reliable than the player estimate even at low network bandwidths

index.html
evanfarina: Sure thing 👍

README.md
```diff
@@ -473,6 +474,11 @@ This option defaults to `false`.
...
+* Type: `boolean`,
+* Default: `false`
+* Use [window.networkInformation.downlink](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink) to estimate the network's bandwidth
```
brandonocasey, line 480: Would be nice to add the caveat here that we will only use it if it's supported and that:

`The value is never greater than 10 Mbps, as a non-standard anti-fingerprinting measure.`

src/videojs-http-streaming.js
```diff
@@ -789,7 +791,22 @@ class VhsHandler extends Component {
...
+          const networkInformation = window.navigator.connection || window.navigator.mozConnection || window.navigator.webkitConnection;
+
+          if (this.options_.useNetworkInformationApi && networkInformation) {
+            // downlink returns Mbps
+            // https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink
+            const effectiveBandwidthBitsPerSec = networkInformation.downlink * 1000 * 1000;
+
+            // downlink maxes out at 10 Mbps. In the event that the player estimates a bandwidth
+            // greater than 10 Mbps, use the larger value to ensure that high quality streams are not
+            // filtered out.
```
brandonocasey, line 805: We might want to explicitly check if the network API reported bandwidth and the player reported are both over 10mbps and only then use the player/vhs bandwidth estimate. The way that the code is written now, if the player/vhs estimates higher bandwidth we will always use that. I think that the network API should be prioritized instead.

gesinger (approved):

PR #1221, opened 2021-11-08
Don't run networkInfo tests against ie11

evanfarina:
## Description
The stubbing of networkInfo doesn't play nicely with ie11 which complains with the following message: `Assignment to read-only properties is not allowed in strict mode`. After speaking with Gary, we've decided to skip these tests in IE.

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev (approved):
Confirmed locally that these tests are now skipped on IE.

Issue #973, opened 2020-10-15
cacheEncryptionKeys has no effect with subtitle streams


phloxic:
## Description
Reduced test case can be found [here](https://phloxic.productions/test/videojs/cachevttkeys.html).

## Steps to reproduce

1. AES encrypted HLS source with subtitle stream
2. set `cacheEncryptionKeys: true`
3. play and monitor GETting of key files

### Results
Keys for video renditions are fetched once, key(s) for subtitles are re-fetched and not cached.
### Expected
Key files for subtitle streams should be cached too.

blacktrash:
Still happening with 7.11.1:
<img width="454" alt="Screen Shot 2020-12-31 at 15 22 54" src="https://user-images.githubusercontent.com/168338/103416066-c5b60380-4b7c-11eb-9895-fda63308c064.png">

blacktrash:
Could the "outdated" label be removed? imho the issue is not outdated. Thank you.

blacktrash:
oh, thank you bot

gkatsev:
Looks like a long time ago I wrote a response to this, and then this got lost in the churn. Thankfully, Firefox saved my response, which is posted below. A PR would be helpful, as it's clearly something we're unlikely to get to any time soon.

----

Looks like our [VTT Segment Loader](https://github.com/videojs/http-streaming/blob/main/src/vtt-segment-loader.js) isn't aware of that option, but our [regular segment loader is](https://github.com/videojs/http-streaming/blob/main/src/segment-loader.js). It should be relatively straight forward to add.

phloxic:
Wanted to look into this anyway, thanks for the nudge. Will have a go.

For the time being I had brushed this under the carpet somewhat by shamelessly increasing targetduration for vtt.

phloxic:
unstale

venomone:
Any update onto this? Would also interest me to have this fixed ...

phloxic:
@venomone - this is fixed I believe.

phloxic:
Fixed via 66a5b17de192eee5d23fadba6a21f6a54f0eb34c

PR #1228, opened 2021-12-01
fix: cache aes keys for text tracks (#973)

phloxic:
## Description
So far aes keys for vtt are not cached even when `cacheEncryptionKeys: true`.

## Specific Changes proposed
Call `segmentKey()` if the requested vtt segment includes a key before appending, so key can be cached.

## Requirements Checklist
  - [ ] Docs/guides updated


phloxic:
Here's [an example](https://phloxic.productions/test/videojs/fix973.html).
In network panel filter URLs with `.key`.
Example [before fix](https://phloxic.productions/test/videojs/cachevttkeys.html).

phloxic:
@gkatsev the fix is easy and works, but right now chances are low that I will time to dive into the testing rabbit hole anytime soon, sorry.

phloxic:
@gkatsev - is the lack of a test blocking this?

gkatsev:
A test would be greatly appreciated!

phloxic:
@gkatsev - I feared as much ;-)
Unfortunately I have no clue how the tests work, especially the media related ones. For instance `test/manifests/subtitles.m3u8` loads `./[0-2].webvtt` which I can't find, probably because they are generated somehow, similarly to the manifests referenced by `test/manifests/master-subtitles.m3u8` ... and at that point I give up because I am getting too embarrased ;-) but mainly because, as much as I'd like to, I simply do not have the time atm to really dig into it. Sorry.

misteroneill (approved):

gkatsev (approved):

PR #1229, opened 2021-12-03
fix: remove deprecation hls options, properties, and events

gkatsev:
BREAKING CHANGE: remove deprecation hls options, properties, and events


src/videojs-http-streaming.js
```diff
@@ -519,42 +515,9 @@ class VhsHandler extends Component {
 
     this.logger_ = logger('VhsHandler');
 
-    // tech.player() is deprecated but setup a reference to HLS for
-    // backwards-compatibility
+    // tech.player() but we need access to the player in a few cases
```
gesinger, line 518: I think this comment needs to be updated.

gesinger (approved):

PR #1235, opened 2021-12-10
test(playack): make live dash test take 5 seconds

gkatsev:


gkatsev:
Increasing the `playFor` timer seems to make the test more reliable, particularly, on Firefox. I've run the tests twice.
Since this is a test-only change, and it only involves changing the `playFor` from 2 to 5, I'm merging it without review.

PR #1268, opened 2022-04-07
Remove deprecated smooth quality change

harisha-swaminathan:
## Description

Removes deprecated `smoothQualityChange` option


misteroneill (changes requested):
Something's up here. The `next` branch may need to be updated.

gesinger (commented):
Looks good! Only a couple lingering references in `test/segment-loader.test.js`.

When we add a migration guide we should also mention the removal (though it probably shouldn't be in use at this point).

gkatsev (approved):

PR #1274, opened 2022-04-22
refactor: rename 'blacklist' to 'exclude'

harisha-swaminathan:


alex-barstow (approved):

gesinger (commented):
While we're modifying the public pieces, it may be worth renaming all instance of `blacklist` throughout the repo as well as the properties (e.g., `ABORT_EARLY_BLACKLIST_SECONDS` and `blacklistReasons` in `MasterPlaylistController`, various comments talking about blacklisting, and test messages).

We also should start a migration guide for updating to the next major version.

gesinger (commented):
I think there are only a few more references.

https://github.com/videojs/http-streaming/blob/8ca62cc6827556b5c92d38ce6546bcdf404383a4/docs/a-walk-through-vhs.md?plain=1#L250

https://github.com/videojs/http-streaming/blob/8ca62cc6827556b5c92d38ce6546bcdf404383a4/src/media-segment-request.js#L428

 and then we trigger `blacklistplaylist` in src/master-playlist-controller.js: https://github.com/videojs/http-streaming/blob/8ca62cc6827556b5c92d38ce6546bcdf404383a4/src/master-playlist-controller.js#L1233

(and test for it in test/videojs-http-streaming.text.js):
https://github.com/videojs/http-streaming/blob/8ca62cc6827556b5c92d38ce6546bcdf404383a4/test/videojs-http-streaming.test.js#L2056

gesinger (approved):

PR #1301, opened 2022-07-28
feat: remove experimental options

kchang-brightcove:
## Description
Removed some experimental options and make them default.

### DIFF
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -513,7 +513,6 @@
     });
 
     [
-      'buffer-water',
       'llhls',
     ].forEach(function(name) {
       stateEls[name].checked = true
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -520,7 +520,8 @@
 
     [
       'exact-manifest-timings',
-      'pixel-diff-selector'
+      'pixel-diff-selector',
+      'bufferBasedABR'
     ].forEach(function(name) {
       stateEls[name].checked = false;
     });
@@ -579,7 +580,7 @@
           html5: {
             vhs: {
               overrideNative: getInputValue(stateEls['override-native']),
-              experimentalBufferBasedABR: getInputValue(stateEls['buffer-water']),
+              bufferBasedABR: getInputValue(stateEls['buffer-water']),
               llhls: getInputValue(stateEls.llhls),
               exactManifestTimings: getInputValue(stateEls['exact-manifest-timings']),
               leastPixelDiffSelector: getInputValue(stateEls['pixel-diff-selector']),
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -521,7 +521,7 @@
     [
       'exact-manifest-timings',
       'pixel-diff-selector',
-      'bufferBasedABR'
+      'buffer-water'
     ].forEach(function(name) {
       stateEls[name].checked = false;
     });
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -513,16 +513,16 @@
     });
 
     [
-      'llhls',
+      'llhls'
     ].forEach(function(name) {
-      stateEls[name].checked = true
+      stateEls[name].checked = true;
     });
 
     [
       'exact-manifest-timings',
       'pixel-diff-selector'
     ].forEach(function(name) {
-      stateEls[name].checked = false
+      stateEls[name].checked = false;
     });
 
     stateEls.debug.addEventListener('change', function(event) {
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -512,6 +512,20 @@
       });
     });
 
+    [
+      'buffer-water',
+      'llhls',
+    ].forEach(function(name) {
+      stateEls[name].checked = true
+    });
+
+    [
+      'exact-manifest-timings',
+      'pixel-diff-selector'
+    ].forEach(function(name) {
+      stateEls[name].checked = false
+    });
+
     stateEls.debug.addEventListener('change', function(event) {
       saveState();
       window.videojs.log.level(event.target.checked ? 'debug' : 'info');
@@ -567,9 +581,9 @@
             vhs: {
               overrideNative: getInputValue(stateEls['override-native']),
               experimentalBufferBasedABR: getInputValue(stateEls['buffer-water']),
-              experimentalLLHLS: getInputValue(stateEls.llhls),
-              experimentalExactManifestTimings: getInputValue(stateEls['exact-manifest-timings']),
-              experimentalLeastPixelDiffSelector: getInputValue(stateEls['pixel-diff-selector']),
+              llhls: getInputValue(stateEls.llhls),
+              exactManifestTimings: getInputValue(stateEls['exact-manifest-timings']),
+              leastPixelDiffSelector: getInputValue(stateEls['pixel-diff-selector']),
               useNetworkInformationApi: getInputValue(stateEls['network-info']),
               useDtsForTimestampOffset: getInputValue(stateEls['dts-offset'])
             }
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -34,7 +34,7 @@ export const parseManifest = ({
   manifestString,
   customTagParsers = [],
   customTagMappers = [],
-  llhls
+  llhls = true
 }) => {
   const parser = new M3u8Parser();
 
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -34,7 +34,7 @@ export const parseManifest = ({
   manifestString,
   customTagParsers = [],
   customTagMappers = [],
-  llhls = true
+  llhls
 }) => {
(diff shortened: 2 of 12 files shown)


gkatsev:
This probably should go against the `next` branch? Also, should we enable llhls by default here as well? @misteroneill?

src/manifest.js
```diff
@@ -23,7 +23,7 @@ export const createPlaylistID = (index, uri) => {
  *        An array of custom tag parsers for the m3u8-parser instance
  * @param {Object[]} [customTagMappers]
  *        An array of custom tag mappers for the m3u8-parser instance
- * @param {boolean} [experimentalLLHLS=false]
+ * @param {boolean} [llhls=false]
```
misteroneill, line 26: This should default to `true` now.

misteroneill (changes requested):
Yeah, we should, I think. The others are blocked by lack of enough real world data.

src/manifest.js
```diff
@@ -34,7 +34,7 @@ export const parseManifest = ({
   manifestString,
   customTagParsers = [],
   customTagMappers = [],
-  experimentalLLHLS
+  llhls
```
misteroneill, line 37: The change to the JSDoc isn't enough to default it to `true`. It would need to be changed here as well
```suggestion
  llhls = true
```

misteroneill (changes requested):

misteroneill (approved):
LGTM. @gesinger would you mind having a quick look?

src/playlist-loader.js
```diff
@@ -395,11 +395,11 @@ export default class PlaylistLoader extends EventTarget {
...
     this.customTagMappers = (vhsOptions && vhsOptions.customTagMappers) || [];
-    this.experimentalLLHLS = (vhsOptions && vhsOptions.experimentalLLHLS) || false;
+    this.llhls = (vhsOptions && vhsOptions.llhls) || false;
 
-    // force experimentalLLHLS for IE 11
+    // force llhls for IE 11
     if (videojs.browser.IE_VERSION) {
-      this.experimentalLLHLS = false;
+      this.llhls = false;
     }
```
gesinger, line 403: I think this whole block can be removed.

src/manifest.js
```diff
@@ -34,7 +34,7 @@ export const parseManifest = ({
   manifestString,
   customTagParsers = [],
   customTagMappers = [],
-  experimentalLLHLS
+  llhls = true
```
gesinger, line 37: I think this default should be at a higher level (probably in src/videojs-http-streaming.js), as, as it stands, it looks like playlist-loader will set it to false if left unspecified.

src/playlist-loader.js
```diff
@@ -395,12 +395,7 @@ export default class PlaylistLoader extends EventTarget {
 
     this.customTagParsers = (vhsOptions && vhsOptions.customTagParsers) || [];
     this.customTagMappers = (vhsOptions && vhsOptions.customTagMappers) || [];
-    this.llhls = (vhsOptions && vhsOptions.llhls) || false;
-
-    // force llhls for IE 11
-    if (videojs.browser.IE_VERSION) {
-      this.llhls = false;
-    }
+    this.llhls = (vhsOptions && vhsOptions.llhls) || true;
```
gesinger, line 398: This line will always evaluate to true. But I think the option should to be set at a higher level (in src/videojs-http-streaming.js, probably in `setOptions_` https://github.com/videojs/http-streaming/blob/main/src/videojs-http-streaming.js#L622 ).

src/manifest.js
```diff
@@ -23,7 +23,7 @@ export const createPlaylistID = (index, uri) => {
  *        An array of custom tag parsers for the m3u8-parser instance
  * @param {Object[]} [customTagMappers]
  *        An array of custom tag mappers for the m3u8-parser instance
- * @param {boolean} [experimentalLLHLS=false]
+ * @param {boolean} [llhls=true]
```
gesinger, line 26:
```suggestion
 * @param {boolean} [llhls]
```

src/playlist-loader.js
```diff
@@ -395,7 +395,7 @@ export default class PlaylistLoader extends EventTarget {
 
     this.customTagParsers = (vhsOptions && vhsOptions.customTagParsers) || [];
     this.customTagMappers = (vhsOptions && vhsOptions.customTagMappers) || [];
-    this.llhls = (vhsOptions && vhsOptions.llhls) || true;
+    this.llhls = (vhsOptions && vhsOptions.llhls) || false;
```
gesinger, line 398: Since the option is set at a higher level, I think we can just use:

```suggestion
    this.llhls = vhsOptions && vhsOptions.llhls;
```

src/videojs-http-streaming.js
```diff
@@ -597,6 +597,7 @@ class VhsHandler extends Component {
     this.options_.customTagParsers = this.options_.customTagParsers || [];
     this.options_.customTagMappers = this.options_.customTagMappers || [];
     this.options_.cacheEncryptionKeys = this.options_.cacheEncryptionKeys || false;
+    this.options_.llhls = this.options_.llhls === false ? false : true;
```
gesinger, line 600: I know we use this style above, but I think it's clearer as:

```suggestion
    this.options_.llhls = Boolean(this.options_.llhls);
```

src/playlist-loader.js
```diff
@@ -395,7 +395,7 @@ export default class PlaylistLoader extends EventTarget {
 
     this.customTagParsers = (vhsOptions && vhsOptions.customTagParsers) || [];
     this.customTagMappers = (vhsOptions && vhsOptions.customTagMappers) || [];
-    this.llhls = (vhsOptions && vhsOptions.llhls) || false;
+    this.llhls = (vhsOptions && vhsOptions.llhls);
```
gesinger, line 398: Minor, but the parens can be removed.

gesinger (commented):
I think the build is failing due to some linting issues in `scripts/index.js`

scripts/old-index.js
```diff
@@ -398,9 +398,9 @@
             vhs: {
```
gkatsev, line 398: I removed this entire file in #1308

src/videojs-http-streaming.js
```diff
@@ -597,6 +597,7 @@ class VhsHandler extends Component {
     this.options_.customTagParsers = this.options_.customTagParsers || [];
     this.options_.customTagMappers = this.options_.customTagMappers || [];
     this.options_.cacheEncryptionKeys = this.options_.cacheEncryptionKeys || false;
+    this.options_.llhls = Boolean(this.options_.llhls);
```
misteroneill, line 600: I could be wrong, but this would still default to `false`, right?

src/videojs-http-streaming.js
```diff
@@ -641,11 +642,11 @@ class VhsHandler extends Component {
       'initialPlaylistSelector',
       'experimentalBufferBasedABR',
```
misteroneill, line 643: Also, we do want to rename this to `bufferBasedABR` - but leave it defaulting to `false` for the time being.

misteroneill (changes requested):

src/videojs-http-streaming.js
kchang-brightcove: Yes

scripts/index.js
```diff
@@ -512,6 +512,20 @@
...
+    [
+      'llhls'
+    ].forEach(function(name) {
+      stateEls[name].checked = true;
+    });
+
+    [
+      'exact-manifest-timings',
+      'pixel-diff-selector',
+      'bufferBasedABR'
```
misteroneill, line 524: This is throwing an error on [the Netlify preview build](https://deploy-preview-1301--videojs-http-streaming.netlify.app/):
```suggestion
      'buffer-water'
```

src/videojs-http-streaming.js
misteroneill: I think we want it to default to `true`, though. Maybe something like this?
```suggestion
    this.options_.llhls = this.options_.hasOwnProperty('llhls') ? this.options_.llhls : true;
```

misteroneill (changes requested):
The [netlify preview](https://deploy-preview-1301--videojs-http-streaming.netlify.app/) is broken. Also, I noticed that fields in the "Options" tabs still say "[EXPERIMENTAL]".

src/videojs-http-streaming.js
misteroneill: Coming back to this, Kevin has restored the previous state:
```js
this.options_.llhls === false ? false : true
```
This is because simply doing `Boolean(this.options_.llhls)` was not defaulting to `true`, which is what was wanted.

misteroneill (approved):

gesinger (approved):

PR #1305, opened 2022-08-09
fix: output-restricted event handling for unplayable streams

alex-barstow:
## Description
This change will improve handling of EME `'keystatuschange'` events of the type `'output-restricted'`. The existing implementation uses `blacklistCurrentPlaylist()` to exclude a single playlist from ABR selection when such an event is encountered, but does not handle cases where there are multiple unplayable streams.

## Specific Changes proposed
I am proposing we do the following:
- When encountering an `'output-restricted'`, we should assume that this is due to HDCP non-compliance and that **all** HD playlists will be unplayable.
- Therefore, exclude all HD playlists from ABR selection.
- Unplayable segments may have already been buffered, so after excluding the HD playlists, perform a `fastQualityChange_()` which will clear the buffer and switch to a playable stream.

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -1062,11 +1062,34 @@ class VhsHandler extends Component {
 
     this.player_.tech_.on('keystatuschange', (e) => {
       if (e.status === 'output-restricted') {
-        this.masterPlaylistController_.blacklistCurrentPlaylist({
-          playlist: this.masterPlaylistController_.media(),
-          message: `DRM keystatus changed to ${e.status}. Playlist will fail to play. Check for HDCP content.`,
-          blacklistDuration: Infinity
-        });
+        const masterPlaylist = this.masterPlaylistController_.media();
+        const excludedHDPlaylists = [];
+
+        if (masterPlaylist && masterPlaylist.playlists) {
+          // Assume all HD streams are unplayable and exclude them from ABR selection, then perform
+          // a fastQualityChange to clear the buffer since it may already contain unplayable segments
+          masterPlaylist.playlists.forEach(playlist => {
+            if (playlist && playlist.attributes && playlist.attributes.RESOLUTION &&
+                playlist.attributes.RESOLUTION.height >= 720) {
+              if (!playlist.excludeUntil || playlist.excludeUntil < Infinity) {
+
+                playlist.excludeUntil = Infinity;
+                excludedHDPlaylists.push(playlist);
+              }
+            }
+          });
+
+          if (excludedHDPlaylists.length) {
+            videojs.log.warn(
+              'DRM keystatus changed to "output-restricted." Removing the following HD playlists ' +
+              'that will most likely fail to play and clearing already buffered HD segments. ' +
+              'Check for HDCP content.',
+              ...excludedHDPlaylists
+            );
+
+            this.masterPlaylistController_.fastQualityChange_();
+          }
+        }
       }
     });
 
--- a/test/videojs-http-streaming.test.js
+++ b/test/videojs-http-streaming.test.js
@@ -4617,12 +4617,21 @@ QUnit.test('configures eme for HLS on source buffer creation', function(assert) 
   }, 'set source eme options');
 });
 
-QUnit.test('eme handles keystatuschange where status is output-restricted', function(assert) {
+QUnit.only('eme handles keystatuschange where status is output-restricted', function(assert) {
+  const originalWarn = videojs.log.warn;
+  let warning = '';
+  let qualitySwitches = 0;
+
+  videojs.log.warn = (...text) => {
+    warning += [...text].join('');
+  };
+
   this.player.eme = {
     options: {
       previousSetting: 1
     }
   };
+
   this.player.src({
     src: 'manifest/master.m3u8',
     type: 'application/x-mpegURL',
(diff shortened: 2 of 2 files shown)


src/videojs-http-streaming.js
```diff
@@ -1062,11 +1062,34 @@ class VhsHandler extends Component {
 
     this.player_.tech_.on('keystatuschange', (e) => {
       if (e.status === 'output-restricted') {
```
gesinger, line 1064: If we change this to:

```
      if (e.status !== 'output-restricted') {
        return;
      }
```

we can remove a level of indentation.

src/videojs-http-streaming.js
```diff
@@ -1062,11 +1062,34 @@ class VhsHandler extends Component {
...
-        this.masterPlaylistController_.blacklistCurrentPlaylist({
-          playlist: this.masterPlaylistController_.media(),
-          message: `DRM keystatus changed to ${e.status}. Playlist will fail to play. Check for HDCP content.`,
-          blacklistDuration: Infinity
-        });
+        const masterPlaylist = this.masterPlaylistController_.media();
+        const excludedHDPlaylists = [];
+
+        if (masterPlaylist && masterPlaylist.playlists) {
```
gesinger, line 1068: Same for this line too, we can return early if these aren't available.

src/videojs-http-streaming.js
```diff
@@ -1062,11 +1062,34 @@ class VhsHandler extends Component {
 
     this.player_.tech_.on('keystatuschange', (e) => {
       if (e.status === 'output-restricted') {
-        this.masterPlaylistController_.blacklistCurrentPlaylist({
-          playlist: this.masterPlaylistController_.media(),
-          message: `DRM keystatus changed to ${e.status}. Playlist will fail to play. Check for HDCP content.`,
-          blacklistDuration: Infinity
-        });
+        const masterPlaylist = this.masterPlaylistController_.media();
```
gesinger, line 1065: `this.masterPlaylistController_.media()` actually returns the currently active media playlist. You want https://github.com/videojs/http-streaming/blob/97e02fb73e227d9f82c0594f42f0575a8f89a29c/src/master-playlist-controller.js#L1665

```suggestion
        const masterPlaylist = this.masterPlaylistController_.master();
```

src/videojs-http-streaming.js
```diff
@@ -1062,11 +1062,34 @@ class VhsHandler extends Component {
...
+        const excludedHDPlaylists = [];
+
+        if (masterPlaylist && masterPlaylist.playlists) {
+          // Assume all HD streams are unplayable and exclude them from ABR selection, then perform
+          // a fastQualityChange to clear the buffer since it may already contain unplayable segments
```
gesinger, line 1070: Might be better to move the second part of the comment ("...clear the buffer since it may already contain unplayable segments") above the call to `fastQualityChange_`

src/videojs-http-streaming.js
```diff
@@ -1062,11 +1062,34 @@ class VhsHandler extends Component {
...
+          masterPlaylist.playlists.forEach(playlist => {
+            if (playlist && playlist.attributes && playlist.attributes.RESOLUTION &&
+                playlist.attributes.RESOLUTION.height >= 720) {
+              if (!playlist.excludeUntil || playlist.excludeUntil < Infinity) {
+
+                playlist.excludeUntil = Infinity;
```
gesinger, line 1076: I'm not sure if we need to check the current `excludeUntil` and can instead probably just set `playlist.excludeUntil = Infinity;`.

src/videojs-http-streaming.js
```diff
@@ -1062,11 +1062,34 @@ class VhsHandler extends Component {
...
+                excludedHDPlaylists.push(playlist);
+              }
+            }
+          });
+
+          if (excludedHDPlaylists.length) {
+            videojs.log.warn(
+              'DRM keystatus changed to "output-restricted." Removing the following HD playlists ' +
+              'that will most likely fail to play and clearing already buffered HD segments. ' +
```
gesinger, line 1085: Technically, we're clearing a buffer which may contain more than just HD segments.

```suggestion
              'that will most likely fail to play and clearing the buffer. ' +
```

src/videojs-http-streaming.js
```diff
@@ -1062,11 +1062,34 @@ class VhsHandler extends Component {
...
+              'Check for HDCP content.',
```
gesinger, line 1086: This may not always be an issue for the stream provider, but can instead involve, for instance, the client not having the right hardware to play back HDCP content. We can add something like:

`This may be due to HDCP restrictions on the stream and the capabilities of the current device."

test/videojs-http-streaming.test.js
```diff
@@ -4635,36 +4644,63 @@ QUnit.test('eme handles keystatuschange where status is output-restricted', func
...
 
   this.player.tech_.vhs.playlists = {
-    master: { playlists: [media] },
-    media: () => media
+    master: { playlists },
+    media: () => playlists[0]
   };
 
-  const excludes = [];
+  this.player.tech_.vhs.masterPlaylistController_.media = () => {
```
gesinger, line 4679:
```suggestion
  this.player.tech_.vhs.masterPlaylistController_.master = () => {
```

src/videojs-http-streaming.js
alex-barstow: Yup, should have done that.

src/videojs-http-streaming.js
alex-barstow: I added this to make sure we only exclude/clear buffer/switch playlists once in the event we get multiple `output-restricted` events, which I have encountered in testing.

gesinger (approved):

misteroneill (approved):

PR #1329, opened 2022-09-22
feat: Add an option to support forced subtitles

amtins:
# Description

Proposal to introduce a new `useForcedSubtitles` option to access forced subtitles when available.

forced-subtitles

## Current State

Currently, VHS ignores forced subtitle tracks. This behavior makes it impossible to access such subtitles, when available, from `player.textTracks`.

As a result, content with occasional foreign language dialogue cannot be translated, forcing the user to enable subtitles for the entire playback time.

### Safari

Referring to the question asked in the [Apple Developer Forums](https://developer.apple.com/forums/thread/43424). Forced subtitles, if available, should be displayed automatically either according to the system language or the selected audio language. This seems to be confirmed by the [Advice about subtitles](https://developer.apple.com/library/archive/releasenotes/AudioVideo/RN-AVFoundation/index.html#//apple_ref/doc/uid/TP40010717-CH1-DontLinkElementID_3) section.

However, it seems that Safari does not implement this behavior, as confirmed by an [Apple Media Engineer](https://developer.apple.com/forums/thread/43424?answerId=212406022#212406022) in point 2 of the response.

Finally, although Safari does not appear to implement this behavior correctly, forced subtitles are available via `document.querySelector('video').textTracks`, allowing them to be enabled programmatically.

### Http Streaming

The [`media-groups.js`](https://github.com/videojs/http-streaming/blob/main/src/media-groups.js#L549-L556) file mentions the implementation choice with reference to point 5.8 of the [Apple's HLS Authoring Specification](https://developer.apple.com/documentation/http_live_streaming/http_live_streaming_hls_authoring_specification_for_apple_devices#2969509). However, this choice affects point 5.9 of the specification, especially when the content contains multiple audio and occasional translations are not burned into the video.

## Specific Changes proposed

- Updated `videojs-http-streaming.js` by adding the `useForcedSubtitles` option _which is false by default_, which can be used as a **initialization** or **source** option.
- Updated the `media-groups.js` file to support the `useForcedSubtitles` option.
- Update the `README.md` file documenting the new option.
- Updated the demo to add a check box for the `useForcedSubtitles` option.
- Add a unit test case in the `master-playlist-controller.test.js` file to ensure that forced subtitles are loaded.
- Update the settings of the `media-groups.test.js` file so that it does not have an undefined value.

### Use of the new option
- At initialization 
  ```JavaScript
  const player = videojs(videoEl, {
    html5 : {
      vhs : { useForcedSubtitles : true }
    }
  });
  
  // Or
  const player = videojs(videoEl, {
    html5 : {
      vhs : { useForcedSubtitles : !videojs.browser.IS_SAFARI }
    }
  });
  ```
- Through the `src`
  ```JavaScript
  const player = videojs(videoEl);
  
  player.src({
    src :'https://example.com/playlist.m3u8', 
    useForcedSubtitles : true 
  });
  
  // Or
  player.src({
    src :'https://example.com/playlist.m3u8', 
    useForcedSubtitles : !videojs.browser.IS_SAFARI
  });
  ```

## Requirements Checklist
- [X] If necessary, more likely in a feature request than a bug fix
  - [X] Docs/guides updated

### DIFF
--- a/src/media-groups.js
+++ b/src/media-groups.js
@@ -546,7 +546,7 @@ export const initialize = {
       }
 
       for (const variantLabel in mediaGroups[type][groupId]) {
-        if (mediaGroups[type][groupId][variantLabel].forced) {
+        if (!vhs.options_.useForcedSubtitles && mediaGroups[type][groupId][variantLabel].forced) {
           // Subtitle playlists with the forced attribute are not selectable in Safari.
           // According to Apple's HLS Authoring Specification:
           //   If content has forced subtitles and regular subtitles in a given language,
--- a/test/media-groups.test.js
+++ b/test/media-groups.test.js
@@ -1076,7 +1076,9 @@ QUnit.module('MediaGroups', function() {
       this.settings = {
         mode: 'html5',
         mainPlaylistLoader: {main: this.main},
-        vhs: {},
+        vhs: {
+          options_: {}
+        },
         tech: {
           options_: {},
           addRemoteTextTrack(track) {
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Video.js Compatibility: 7.x, 8.x
       - [liveRangeSafeTimeDelta](#liverangesafetimedelta)
       - [useNetworkInformationApi](#usenetworkinformationapi)
       - [useDtsForTimestampOffset](#usedtsfortimestampoffset)
+      - [useForcedSubtitles](#useforcedsubtitles)
       - [captionServices](#captionservices)
         - [Format](#format)
         - [Example](#example)
@@ -460,6 +461,14 @@ This option defaults to `false`.
 * Type: `boolean`,
 * Default: `false`
 * Use [Decode Timestamp](https://www.w3.org/TR/media-source/#decode-timestamp) instead of [Presentation Timestamp](https://www.w3.org/TR/media-source/#presentation-timestamp) for [timestampOffset](https://www.w3.org/TR/media-source/#dom-sourcebuffer-timestampoffset) calculation. This option was introduced to align with DTS-based browsers. This option affects only transmuxed data (eg: transport stream). For more info please check the following [issue](https://github.com/videojs/http-streaming/issues/1247).  
+
+##### useForcedSubtitles
+* Type: `boolean`
+* Default: `false`
+* can be used as a source option
+* can be used as an initialization option
+
+If true, this option allows the player to display forced subtitles. When available, forced subtitles allow to translate foreign language dialogues or images containing foreign language characters.
 
 ##### captionServices
 * Type: `object`
--- a/index.html
+++ b/index.html
@@ -171,6 +171,11 @@
           <label class="form-check-label" for="mirror-source">Mirror sources from player.src (reloads player, uses EXPERIMENTAL sourceset option)</label>
         </div>
 
+        <div class="form-check">
+          <input id="forced-subtitles" type="checkbox" class="form-check-input">
+          <label class="form-check-label" for="forced-subtitles">Use Forced Subtitles (reloads player)</label>
+        </div>
+
         <div class="input-group">
           <span class="input-group-text"><label for=preload>Preload (reloads player)</label></span>
           <select id=preload class="form-select">
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -451,7 +451,8 @@
     'dts-offset',
     'override-native',
     'preload',
-    'mirror-source'
+    'mirror-source',
+    'forced-subtitles'
   ].forEach(function(name) {
     stateEls[name] = document.getElementById(name);
   });
@@ -503,7 +504,8 @@
       'pixel-diff-selector',
       'network-info',
       'dts-offset',
-      'exact-manifest-timings'
+      'exact-manifest-timings',
+      'forced-subtitles'
     ].forEach(function(name) {
       stateEls[name].addEventListener('change', function(event) {
         saveState();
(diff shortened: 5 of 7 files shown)


gkatsev:
Thanks for the PR. This seems reasonable to me. Thoughts @videojs/collaborators?

privaloops:
lovely, please merge !

mister-ben:
Would we want the forced tracks to show up in the menu and be user-selectable, or just programatically selectable as Safari?

Safari uses a `kind` of `"forced"` for these tracks. This [is not one of the standard kinds](https://html.spec.whatwg.org/multipage/media.html#text-track-kind) but it's worth discussing whether or not we'd want to follow Safari's approach.

amtins:
@mister-ben, @privaloops, thanks for the feedback. 

My initial thought was to opt for the simplest approach based on what different players in the market are doing. See:
- [hls.js](https://hls-js.netlify.app/demo/?src=https%3A%2F%2Fd2zihajmogu5jn.cloudfront.net%2Fbipbop-advanced%2Fbipbop_16x9_variant.m3u8&demoConfig=eyJlbmFibGVTdHJlYW1pbmciOnRydWUsImF1dG9SZWNvdmVyRXJyb3IiOnRydWUsInN0b3BPblN0YWxsIjpmYWxzZSwiZHVtcGZNUDQiOmZhbHNlLCJsZXZlbENhcHBpbmciOi0xLCJsaW1pdE1ldHJpY3MiOi0xfQ==)
- [jwplayer](https://developer-tools.jwplayer.com/stream-tester)
- [kaltura player](http://player.kaltura.com/modules/KalturaSupport/tests/HLSPlaybackTester.html#%7B%22engine%22%3Atrue%2C%22stvod%22%3Atrue%2C%22ncweb%22%3Atrue%2C%22url%22%3A%22https%3A%2F%2Fd2zihajmogu5jn.cloudfront.net%2Fbipbop-advanced%2Fbipbop_16x9_variant.m3u8%22%2C%22thumb%22%3A%22%22%2C%22hlsLiveSegmentBuffer%22%3A-1%2C%22hlsInitialBufferTime%22%3A-1%2C%22hlsExpandedBufferTime%22%3A-1%2C%22hlsMaxBufferTime%22%3A-1%2C%22hlsLogs%22%3Atrue%2C%22dvr%22%3Afalse%7D)
- [theoplayer](https://www.theoplayer.com/test-your-stream-hls-dash-hesp)
- [visualon](https://www.visualon.com/index.php/html5demo/?demo=custom)

The idea of using a `kind` of `"forced"` is very interesting, even if it is not standard. It seems that there was a discussion on this subject within the [whatwg(issue 4472)](https://github.com/whatwg/html/issues/4472#issuecomment-478514727) which tends to the same conclusion. This discussion is based on another [W3C discussion](https://www.w3.org/Bugs/Public/show_bug.cgi?id=23507).

The use of `"forced"` for the `kind` implies a modification of `video.js` to add this value in the [track-enum](https://github.com/videojs/video.js/blob/4238f5c1d88890547153e7e1de7bd0d1d8e0b236/src/js/tracks/track-enums.js#L44), otherwise all forced subtitles will continue to have a `kind` of `"subtitles"` even if this PR is updated.

Finally, I can also make a PR in the `video.js` repo to add the `"forced"` value. The result will be that the forced subtitles will not be visible in [the subtitles menu](https://github.com/videojs/video.js/blob/e49286a81cf30969f5dd8d739c6740bc07e3bc00/src/js/control-bar/text-track-controls/subs-caps-button.js#L70) but will remain available via the API `player.textTracks()` allowing developers to make their own UX.

amtins:
Hi @mister-ben, @gkatsev, is there anything I can do to help this PR be merged?

misteroneill (approved):
Sorry for the delay on your PR @amtins! It looks reasonable to me, but I'm not a playback expert.

Maybe I could get @adrums86 or @dzianis-dashkevich to have a look?

src/media-groups.js
```diff
@@ -546,7 +546,7 @@ export const initialize = {
       }
 
       for (const variantLabel in mediaGroups[type][groupId]) {
-        if (mediaGroups[type][groupId][variantLabel].forced) {
+        if (!vhs.options_.useForcedSubtitles && mediaGroups[type][groupId][variantLabel].forced) {
```
dzianis-dashkevich, line 549:
```suggestion
        if (!vhs.options().useForcedSubtitles && mediaGroups[type][groupId][variantLabel].forced) {
```

test/media-groups.test.js
```diff
@@ -1076,7 +1076,9 @@ QUnit.module('MediaGroups', function() {
       this.settings = {
         mode: 'html5',
         mainPlaylistLoader: {main: this.main},
-        vhs: {},
+        vhs: {
+          options_: {}
```
dzianis-dashkevich, line 1080:
```suggestion
          options: () => ({})
```

src/media-groups.js
amtins: @dzianis-dashkevich  thank you for taking the time to do the code review and the change proposal. 👍🏽 

I agree with the proposed change which seems more logical, that was my first approach. However, after looking at the code base I saw that instead of the method `options()` it was the property `options_` that was used. So I ended up using the property, to keep some consistency. Would you like me to make the change anyway?

Please see:
https://github.com/videojs/http-streaming/blob/93a2bfdb197ba552122784db430e816bec1f2130/src/media-groups.js#L650
https://github.com/videojs/http-streaming/blob/93a2bfdb197ba552122784db430e816bec1f2130/src/playlist-loader.js#L395

dzianis-dashkevich: Oh never mind then!
I guess it will be a separate effort to refactor all places in the future.
Thank you!

adrums86 (approved):
My apologies for coming in late with this review. I had previously looked over this PR and simply forgot the final step, my approval 😁 ! Nice work, your contribution is greatly appreciated.

adrums86:
Updating the branch as it would be great to get this merged before the next release.

Issue #1287, opened 2022-05-19
HLS stream not switching rendition when network condition changes


andreasgan:
## Description
HLS stream not switching rendition when network condition changes.
Screen recording: https://user-images.githubusercontent.com/727125/169267092-86277870-9d3a-4179-98a0-679aded8fbaf.mp4

## Sources
Can reproduce with https://d2zihajmogu5jn.cloudfront.net/bipbop-advanced/bipbop_16x9_variant.m3u8, see codesandbox below.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.

1. Open https://codesandbox.io/s/gallant-grass-o97t5c?file=/src/index.js
2. Open devtools network tab
3. Start the stream with no network throttling, playing one of the high bandwidth renditions.
4. Switch to "Slow 3G". The video request will likely time out.

### Expected
Quickly switch to a rendition requiring lower bandwidth.

### Actual
Completely stops requesting new segments.

### videojs-http-streaming version
what version of videojs-http-streaming does this occur with?
videojs-http-streaming 2.13.1

### videojs version
what version of videojs does this occur with?
video.js 7.18.1

### Browsers
Microsoft Edge (Chromium-based) Version 101.0.1210.47 (Official build) (arm64)

### Platforms
MacOS

andreasgan:
Updated. Forgot to mention that this only happens with experimentalBufferBasedABR set to true.

gesinger:
This is a great catch @andreasgan ! Thank you for submitting, and for the reproducible test case!

It looks like what's happening is the request times out, leading to the segment loader pausing: https://github.com/videojs/http-streaming/blob/5223427eac208e50af6285562767111c865d05ac/src/segment-loader.js#L2681-L2689

The `handleTimeout_` function triggers `bandwidthupdate`: https://github.com/videojs/http-streaming/blob/5223427eac208e50af6285562767111c865d05ac/src/segment-loader.js#L2636

A `bandwidthupdate` should trigger a new quality selection, but `experimentalBufferBasedABR` disables that quality selection, as it's running on a timer instead: https://github.com/videojs/http-streaming/blob/5223427eac208e50af6285562767111c865d05ac/src/master-playlist-controller.js#L773-L779

This pause shouldn't happen for `experimentalBufferBasedABR`, and we'll look into fixing it.

Thank you again!

gesinger:
[...]

PR #1333, opened 2022-10-13
fix: resume loading on segment timeout for `experimentalBufferBasedABR`

gesinger:
Previously, if there was a segment timeout and `experimentalBufferBasedABR` was set to `true`, then the main segment loader would stay in a `READY` state and never resume loading segments. This is because the buffer based ABR path doesn't follow the same flow as non buffer based ABR, where a new playlist would be loaded immediately and it would trigger a load. Buffer based ABR may determine that no rendition change should be made, despite the timeout, leading to nothing happening. This change makes the call to load explicit, but only for buffer based ABR on timeouts.

This addresses https://github.com/videojs/http-streaming/issues/1287

# Test Changes

### Changed
* Updated test to show that `experimentalBufferBasedABR` does make a playlist selection on `bandwidthupdate`
* Updated test to remove warning logged on a `bandwidthupdate` when there's only one playlist. This warning doesn't always make sense, as a timeout that causes a `bandwidthupdate` would be a connection issue, rather than a stream issue.

### Added

* Added test that segment-loader fires `timeout` event on timeout
* Added test that master-playlist-controller reloads segment loader on `timeout` if experimentalBufferBasedABR

# Testing

1. Open the [demo page](https://videojs-http-streaming.netlify.app/?debug=false&autoplay=false&muted=false&fluid=false&minified=false&sync-workers=false&liveui=true&llhls=false&url=https%3A%2F%2Fd2zihajmogu5jn.cloudfront.net%2Fbipbop-advanced%2Fbipbop_16x9_variant.m3u8&type=application%2Fx-mpegurl&keysystems=&buffer-water=true&exact-manifest-timings=false&pixel-diff-selector=false&network-info=false&dts-offset=false&override-native=true&preload=none&mirror-source=true). Enable `[EXPERIMENTAL] Use Buffer Level for ABR (reloads player)` and set `preload` to `none`.
2. In Chrome Devtools use a very slow connection (I manually created a 10kbps profile).
3. Press play
4. Without the change, after the segment request times out (16.5 seconds for the first bipbop stream, no more requests are made and the buffering spinner remains. The video never starts, even if changing the network connectivity back to something reasonable.
5. With the change, a new segment is requested and playback resumes (if changing network connectivity to something reasonable).

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -365,14 +365,14 @@ export class MasterPlaylistController extends videojs.EventTarget {
   /**
    * Run selectPlaylist and switch to the new playlist if we should
    *
+   * @param {string} [reason=abr] a reason for why the ABR check is made
    * @private
-   *
    */
-  checkABR_() {
+  checkABR_(reason = 'abr') {
     const nextPlaylist = this.selectPlaylist();
 
     if (nextPlaylist && this.shouldSwitchToMedia_(nextPlaylist)) {
-      this.switchMedia_(nextPlaylist, 'abr');
+      this.switchMedia_(nextPlaylist, reason);
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/master-playlist-controller.js
```diff
@@ -770,17 +770,26 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      // useful to check to see if a rendition switch should be made.
+      this.checkABR_('bandwidthupdate');
+      this.tech_.trigger('bandwidthupdate');
+    });
 
-        this.tech_.trigger('bandwidthupdate');
-      });
+    this.mainSegmentLoader_.on('timeout', () => {
+      if (this.experimentalBufferBasedABR) {
+        // If a rendition change is needed, then it would've be done on `bandwidthupdate`.
```
evanfarina, line 782: "wouldn't _been_ done"

evanfarina (approved):
Looks great

gkatsev (approved):

misteroneill (approved):
As discussed on collaborators call, we should ship this since it makes things better.

PR #1334, opened 2022-10-13
Account for MPEGTS rollover when adjusting vtt cues

raytiley:
## Description
This fixes an issue with a long running stream where the cue start times are larger than the maximum mpegts timestamp of ~27 hours.

## Specific Changes proposed
VTT cue start and end times are adjusted to ensure that they are within the range of mpeg ts PTS which has a max size of 2^33.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/vtt-segment-loader.js
+++ b/src/vtt-segment-loader.js
@@ -476,11 +476,13 @@ export default class VTTSegmentLoader extends SegmentLoader {
 
     const timestampmap = segmentInfo.timestampmap;
     const diff = (timestampmap.MPEGTS / ONE_SECOND_IN_TS) - timestampmap.LOCAL + mappingObj.mapping;
+    const mpegTsRollover = Math.floor((2 ** 33) / ONE_SECOND_IN_TS);
 
     segmentInfo.cues.forEach((cue) => {
       // First convert cue time to TS time using the timestamp-map provided within the vtt
-      cue.startTime += diff;
-      cue.endTime += diff;
+      // Also account for mpegts rollover
+      cue.startTime = (cue.startTime % mpegTsRollover) + diff;
+      cue.endTime = (cue.endTime % mpegTsRollover) + diff;
     });
 
     if (!playlist.syncInfo) {


src/vtt-segment-loader.js
```diff
@@ -476,11 +476,12 @@ export default class VTTSegmentLoader extends SegmentLoader {
 
     const timestampmap = segmentInfo.timestampmap;
     const diff = (timestampmap.MPEGTS / ONE_SECOND_IN_TS) - timestampmap.LOCAL + mappingObj.mapping;
-
+    const mpegTsRollover = Math.floor((2**33) / ONE_SECOND_IN_TS);
```
gkatsev, line 479: nit: to fix linter errors
```suggestion
    const mpegTsRollover = Math.floor((2 ** 33) / ONE_SECOND_IN_TS);
    
```

gkatsev, line 479: I wonder if there's a way to know whether TS segments are used here and only apply this rollover in those cases. @gesinger thoughts?

raytiley:
@gkatsev / @gesinger - Any thoughts on what I can do to get this in? Thanks!

misteroneill (commented):
@raytiley Sorry for the delay on this. I've changed the base branch to 2.x - can you verify that the diff looks correct to you?

adrums86 (changes requested):
It would be great to see a test for this.

src/vtt-segment-loader.js
```diff
@@ -476,11 +476,13 @@ export default class VTTSegmentLoader extends SegmentLoader {
 
     const timestampmap = segmentInfo.timestampmap;
     const diff = (timestampmap.MPEGTS / ONE_SECOND_IN_TS) - timestampmap.LOCAL + mappingObj.mapping;
+    const mpegTsRollover = Math.floor((2 ** 33) / ONE_SECOND_IN_TS);
```
adrums86, line 479: Since this code isn't isolated to TS segments, we may want to use the actual media timescale rather than `ONE_SECOND_IN_TS` (90000).

dzianis-dashkevich:
I assume that this PR should handle this case as well:
https://github.com/videojs/http-streaming/pull/1472

PS: it is not related to TS segments only. There might be cases when we have CMAF for video/audio, and VTT might still have an MPEGTS timestamp map.

adrums86:
I think this was handled in this PR https://github.com/videojs/http-streaming/pull/1472. @raytiley please let me know if this doesn't fix your rollover issue. Closing for now.

Issue #1351, opened 2022-12-02
Linear DASH SSAI Label Issue


adrums86:
## Description
We have an issue with a [DASH stream](https://cs1.brightcodes.net/doleskevicius/cases/case-692718/6.66.6.html) with SSAI. Occasionally this stream will stall when we reach an ad break. This is because of our support for the `<Label>` element in an `AdaptationSet`. In this specific case some of the `AdaptationSet`s in the ad break don’t have a `<Label>` element and some do. Because we use the contents of the `<Label>` to set the `mediaGroup.type.group.label` from the [mpd-parser](https://github.com/videojs/mpd-parser/blob/3fc0486cbc06114635055d540606f850477c8c5d/src/toM3u8.js#L178), when an ad period enters the manifest with a different label we don’t add and update the associated playlists because we rely on the label from the new/updated manifest to access the mediaGroups _and_  playlists from the old or `update` manifest. Examples [here](https://github.com/videojs/http-streaming/blob/afb1ff700e490d131df4aea66d746cf72794caae/src/playlist-loader.js#L278) in the `playlist-loader` and [here](https://github.com/videojs/http-streaming/blob/afb1ff700e490d131df4aea66d746cf72794caae/src/dash-playlist-loader.js#L174) in the `dash-playlist-loader`

This is not reproducible with the same stream in _both_ [DASH.js](https://github.com/Dash-Industry-Forum/dash.js/) and [Shaka](https://github.com/shaka-project/shaka-player)

## Sources
Load the [example page](https://cs1.brightcodes.net/doleskevicius/cases/case-692718/6.66.6.html) and copy the `.mpd` URL from dev tools.

## Steps to reproduce
_**Note: This stream is geo-restricted and requires an IP from within Australia, VPN may be necessary!**_
1. Load the `.mpd` URL from the example above into a VHS demo page. `npm run start` locally or the [live example](https://jsbin.com/gejugat/edit?html,output)
2. Wait for content to hit an ad break.

## Results
Playback stalls often without recovering, or concurrent stalls eventually cause the player to fall out of the live window.
### Expected
Playback continues through the ad break without issue.

### Error output
No specific error is logged until well after the stream has already stalled and failed to recover. The player will `"blacklist"` the playlist after exhausting attempts to increase the buffer with the old playlist. Look for `bad segment` in the debug logging.

### videojs-http-streaming version
videojs-http-streaming 3.0.0

### videojs version
video.js 7.15.4

### Browsers
[...]

PR #1352, opened 2022-12-03
fix: Linear DASH multiperiod label issue

adrums86:
## Description
Fixes [Linear DASH SSAI Label Issue](https://github.com/videojs/http-streaming/issues/1351)
Linear multi-period DASH with `<Label>` elements that roll in and out of continuous `AdaptationSet`s cause the stream to stall because we use the value of the `<Label>` as the key for the `mediaGroups.type.group.<labelKey>` object _and_ the associated mediaGroup group/playlist ID. When the manifest updates and reaches the period either with a `<Label>` or omitting the `<Label>` we can no longer access that media group or playlists to update the playlist as we rely on the labelKey set from the previous manifest/period to look for new segments in the update.

## Specific Changes proposed

- Change the way we create groupIDs for mediaGroup playlists to allow for DASH playlists to use the representation ID rather than the label from the parser. Per [DASH IF](https://dashif.org/docs/DASH-IF-IOP-v4.2-clean.htm), continuous media in separate periods share the same representation IDs across period boundaries, so this change can easily utilize that feature to ensure our playlistIDs are consistent between continuous periods.
- Add the new `labelKey` to the `mediaGroups.type.group` object and add the playlist reference.
- Remove any old `labelKey`s from the `mediaGroups.type.group` object that no longer exist in the new manifest update.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -86,8 +86,12 @@ const dashPlaylistUnchanged = function(a, b) {
   return true;
 };
 
-// DASH playlist playlistIDs should be the same accross period regardless of the mediaGroup label
+/**
+ * Use the representation ids from the mpd object to create groupIDs, this allows for continuity across periods
+ * regardless of the current label.
+ */
 const dashGroupId = (type, group, label, playlist) => {
+  // NAME is set to the id in the parser.
   const playlistId = playlist.attributes && playlist.attributes.NAME ? playlist.attributes.NAME : label;
 
   return `placeholder-uri-${type}-${group}-${playlistId}`;
@@ -126,6 +130,22 @@ export const parseMainXml = ({
   addPropertiesToMain(manifest, srcUrl, dashGroupId);
 
   return manifest;
+};
+
+/**
+ * Removes any mediaGroup labels that no longer exist in the newMain
+ *
+ * @param {Object} update
+ *         The previous mpd object being updated
+ * @param {Object} newMain
+ *         The new mpd object
+ */
+const removeOldMediaGroupLabels = (update, newMain) => {
+  forEachMediaGroup(update, (_, type, group, label) => {
+    if (!(label in newMain.mediaGroups[type][group])) {
+      delete update.mediaGroups[type][group][label];
+    }
+  });
 };
 
 /**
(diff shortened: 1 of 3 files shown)


src/dash-playlist-loader.js
```diff
@@ -116,11 +129,27 @@ export const parseMainXml = ({
...
+/**
+ * Removes any mediaGroup labels that no longer exist in the newMain
+ *
+ * @param {Object} update
+ *         The previous mpd object being updated
+ * @param {Object} newMain
+ *         The new mpd object
+ */
+const removeOldMediaGroupLabels = (update, newMain) => {
+  forEachMediaGroup(update, (_, type, group, label) => {
```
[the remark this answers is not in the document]
adrums86: Good call, I have an old lint-driven habit of using `_` for any unused parameters. `Consistency > arbitrary lint rules` IMO.

src/dash-playlist-loader.js
```diff
@@ -86,6 +86,19 @@ const dashPlaylistUnchanged = function(a, b) {
...
 
+/**
+ * Use the representation IDs from the mpd object to create groupIDs, this allows for continuous playout across periods
+ * with the same representation IDs (continuous periods as defined in DASH-IF 3.2.12). This is assumed in the mpd-parser
+ * as well. If we want to support periods without continuous playback this function may need modification as well as the
+ * parser.
+ */
+const dashGroupId = (type, group, label, playlist) => {
+  // NAME is set to the id in the parser.
+  const playlistId = playlist.attributes && playlist.attributes.NAME ? playlist.attributes.NAME : label;
```
wseymour15, line 97: nit: You can save a bit of code by using the [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator here, and then using an OR statement:
`const playlistId = playlist.attributes?.NAME || label;`

wseymour15 (approved):
Looks good! Just have that one optional change.

src/dash-playlist-loader.js
gkatsev: I think the question is whether the babel config transpiles optional chaining, because its output is a lot larger comparably.

wseymour15: Good point, if this is going to cause transpiling issues then it's definitely not worth the change. However, I would expect it to be able to transpile optional chaining.

adrums86: @wseymour15 I love that syntax change, however it does unfortunately transpile to a significantly longer switch statement in this case.
` const playlistId = ((_playlist$attributes = playlist.attributes) == null ? void 0 : _playlist$attributes.NAME) || label;`

adrums86: That said, I think I can actually get away with `const playlistId = playlist.attributes.NAME || label;` per the NAME being the representation `id` from the DASH parser which is a required, unique property. My old typescript habits leaking through with that extra check that `playlist.attributes` exists

alex-barstow (approved):

PR #1360, opened 2023-01-18
Fix: In-manifest VTT iOS MSE issue

dzianis-dashkevich:
## Description
We have a bug with default options setup for HLS in-manifest VTT in chrome iOS (and I assume other non-safari Webkit-based iOS browsers). The problem is that we do not load `vtt.js`, but we use `VHS` to handle playback. `vtt-segment-loader` implicitly depends on `window.WebVTT.Parser`. Since `vtt.js` is not loaded - `window.WebVTT` is `undefined`. It means that we are not able to parse received VTT cues.

## Specific Changes proposed
- Do not override native for all iOS/iPadOS browsers by default.
- Add fallback logic to load `vtt.js` in `vtt-segment-loader` if we do not have it loaded. (eg: force `overrideNative` option case).
- Add exception guard with a specific error message in case we call parse but still do not have vtt.js loaded for any reason.
- Fix/Add tests to cover all listed changes.

> **Note**
> Tested Chrome on iPadOS 16.2:
> default run: (native playback and apple's native text tracks are shown)
> set override Native to true (VHS playback and apple's native text tracks are shown)
> set override Native to true and set native text tracks to false (VHS playback and styled text tracks are shown)

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/vtt-segment-loader.js
+++ b/src/vtt-segment-loader.js
@@ -41,7 +41,7 @@ export default class VTTSegmentLoader extends SegmentLoader {
 
     this.featuresNativeTextTracks_ = settings.featuresNativeTextTracks;
 
-    this.loadVttJs_ = settings.loadVttJs;
+    this.loadVttJs = settings.loadVttJs;
 
     // The VTT segment will have its own time mappings. Saving VTT segment timing info in
     // the sync controller leads to improper behavior.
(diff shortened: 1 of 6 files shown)


src/vtt-segment-loader.js
```diff
@@ -298,29 +306,16 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
-      this.subtitlesTrack_.tech_.one('vttjsloaded', loadHandler);
-      this.subtitlesTrack_.tech_.one('vttjserror', errorHandler);
-
+      // should be fine to call multiple times
+      // script will be loaded once but multiple listeners will be added to the queue, which is expected.
+      this.loadVttJs()
+        .then(
+          () => this.segmentRequestFinished_(error, simpleSegment, result),
+          () => this.stopForError({ message: 'Error loading vtt.js' })
+        );
```
Essk, line 318: Nice tidy-up job here 🙌

Essk (approved):
This all looks good to me

alex-barstow (approved):

adrums86 (approved):
:shipit: !

PR #1370, opened 2023-02-08
feat: add fmp4 emsg ID3 support

adrums86:
## Description
Currently VHS does not support passing ID3 frames to the text track from fmp4 segment emsg boxes. With some [additional parsing from mux.js](https://github.com/videojs/mux.js/pull/426) and this PR, this enables passing parsed ID3 frame data to the player `textTracks_` for use as timed metadata.

## Specific Changes proposed
Adding a new function to the `transmuxer-worker` that calls `getEmsgId3` on the transmuxer and returns parsed ID3 frames. Calling the worker function and passing the ID3 frames to the `id3Fn` passed in from the mediaSegmentRequest.

> ~Note: This is blocked by the work in the transmuxer being released and integrated. See: https://github.com/videojs/mux.js/pull/426~
Changes are integrated into VHS see PR: https://github.com/videojs/http-streaming/pull/1372

References: 
https://aomediacodec.github.io/id3-emsg/
https://dashif-documents.azurewebsites.net/Events/master/event.html#event-metadata-timing
https://github.com/id3/ID3v2.4/blob/master/id3v2.40-structure.txt

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -462,7 +462,7 @@ const handleSegmentBytes = ({
     // Note that the start time returned by the probe reflects the baseMediaDecodeTime, as
     // that is the true start of the segment (where the playback engine should begin
     // decoding).
-    const finishLoading = (captions) => {
+    const finishLoading = (captions, id3Frames) => {
       // if the track still has audio at this point it is only possible
       // for it to be audio only. See `tracks.video && tracks.audio` if statement
       // above.
@@ -471,6 +471,9 @@ const handleSegmentBytes = ({
         data: bytesAsUint8Array,
         type: trackInfo.hasAudio && !trackInfo.isMuxed ? 'audio' : 'video'
       });
+      if (id3Frames && id3Frames.length) {
+        id3Fn(segment, id3Frames);
+      }
       if (captions && captions.length) {
         captionsFn(segment, captions);
       }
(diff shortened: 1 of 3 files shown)


adrums86:
Setting as draft until transmuxer changes are integrated, as the tests will fail until we have access to the `getEmsgId3` function.

src/media-segment-request.js
```diff
@@ -516,7 +519,19 @@ const handleSegmentBytes = ({
...
+              data: bytesAsUint8Array,
+              transmuxer: segment.transmuxer,
+              offset: startTime,
+              callback: ({emsgData, id3Frames}) => {
+                // transfer bytes back to us
+                bytes = emsgData.buffer;
+                segment.bytes = bytesAsUint8Array = emsgData;
+                finishLoading(message.captions, id3Frames);
+              }
+            });
```
dzianis-dashkevich, line 534: Curious: Currently, we have three `workerCallbacks` here chaining each other. It seems like they can be run absolutely separately. Should we separate them?

dzianis-dashkevich (approved):

src/media-segment-request.js
adrums86: I couldn't find a way to separate them completely without breaking the `startTime` and `captions` probe for some reason. After playing around with it a bit I settled on nesting the id3 probing similar to the way the captions extraction is nested. I'll play around with it a bit more and see if there's a better way to pull them all apart, I think that we should separate them if possible.

adrums86: I tried a few other permutations of splitting these `workerCallback`s and am still running into issues sending messages to the service worker. Going to leave it as is for now, and make a note and possibly an issue to look into splitting up those calls backs.

PR #1374, opened 2023-02-24
Fix: CMAF HLS. Source buffer change type is called with wrong codecs sometimes when append segment without init data because of a race condition.

dzianis-dashkevich:
## Description
We have a race condition issue that breaks playback during quality switch:
CleanShot 2023-02-23 at 22 52 47@2x
CleanShot 2023-02-23 at 22 55 32@2x

Here is a visual representation of the race condition:
CleanShot 2023-02-25 at 23 25 44@2x

As you can see from the visual representation, we have a pending segment request which is from the previous playlist, because the `seeking` event kick-starts the segment loader before we have a new playlist loaded. But we receive `trackInfo` after the new playlist is loaded. That means that during codec calculation it will get codecs from the new playlist, while the segment loader is about to append a segment from the previous one. Because initialization data is not added and codecs do not match with the source buffer's current codecs setup - the source buffer fails to append this segment.
The problem is reproducible mainly with CMAF HLS since the transport stream is usually self-initialized.

## Specific Changes proposed
I was thinking about possible ways to solve this problem. The first idea was the following:
Do not load segments while we are waiting for a playlist. 
Possible solutions could be: Ignore this seeking event after setting the current time after a segment loader resets everything. Visually it would look like this:
CleanShot 2023-02-25 at 23 22 27@2x

It works fine and fixes the issue, but!
Users will always see a loading spinner during the quality switch because the player removes everything from the source buffer and does not append anything. It will append only after the playlist + first segment are loaded.
So, users will lose a seamless quality switching experience.
This problem led me to the second idea: Keep segments loading but ensure that the player always adds initialization data after the quality switch, and the player always uses the segment's playlist for source buffer codecs calculation for consistency.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1725,9 +1725,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
       audio: this.audioSegmentLoader_.getCurrentMediaInfo_() || {}
     };
 
+    const playlist = this.mainSegmentLoader_.getPendingSegmentPlaylist() || this.media();
+
     // set "main" media equal to video
     media.video = media.main;
-    const playlistCodecs = codecsForPlaylist(this.master(), this.media());
+    const playlistCodecs = codecsForPlaylist(this.master(), playlist);
     const codecs = {};
     const usingAudioLoader = !!this.mediaTypes_.AUDIO.activePlaylistLoader;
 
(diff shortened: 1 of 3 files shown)


wseymour15 (approved):
LGTM

adrums86 (approved):
Nice work! I'm really surprised we didn't see this more often when seeking.

PR #1377, opened 2023-02-28
chore: npm publish for release workflow

dzianis-dashkevich:
add npm publish step for release workflow
update release documentation

NOTE: `NPM_TOKEN` should be added to the repo secrets.

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -37,7 +37,7 @@ jobs:
 
       # publish runs build for us via a prepublishOnly script
       - name: npm release
-        run: npm publish --tag next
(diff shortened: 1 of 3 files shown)


gkatsev (commented):
can `github-release.yml` be renamed just `release.yml`? (The job name and the workflow name will need to be updated in the yml file too.

Also, while we're at it, can you add "discussion_category_name" to the github release action? See https://github.com/videojs/video.js/blob/main/.github/workflows/release.yml#L66 for example

Thanks!

dzianis-dashkevich:
@gkatsev updated

gkatsev (approved):

PR #1381, opened 2023-03-17
refactor: remove nested loop from removeDuplicateCuesFromTrack function

alex-barstow:
## Description
The `removeDuplicateCuesFromTrack()` function was poorly optimized. The nested loop would run a staggering number of times in cases of very large VTT files. This change accomplishes duplicate cue removal with a single loop.


dzianis-dashkevich (approved):

adrums86 (approved):
Good find!

PR #1386, opened 2023-03-29
fix: use audio offset for id3 with audio-only

### LINKED ISSUE #130: cuechange event not being triggered on audio only HLS streams
*[nomayonnaise](https://github.com/nomayonnaise) commented on Sep 12, 2017, 10:04 AM UTC:*

## Description

The cuechange eventListener is **not** being triggered for audio only HLS streams when using this plug-in. The audio only stream contains 3 10 second .ts segments with AAC audio 64kb/s. This happens with a real live HLS stream as well.

Here are 2 examples with 3 ts segments

With audio only - cuechange event **is not** triggered
[http://jsfiddle.net/8f05vyb2/](http://jsfiddle.net/8f05vyb2/)

With video and audio - cuechange event **is** triggered
[http://jsfiddle.net/p228339u/4/](http://jsfiddle.net/p228339u/4/)

## Sources

With audio only
[http://li720-23.members.linode.com/hls/audio.m3u8](http://li720-23.members.linode.com/hls/audio.m3u8)
With video and audio
[http://li720-23.members.linode.com/hls/video.m3u8](http://li720-23.members.linode.com/hls/video.m3u8)

## Steps to reproduce

1.  Open audio only link in Chrome
2.  Play video (should autoplay)

### Expected

cuechange triggered, input value is updated

### Error output

There is no error output, cuechange event not triggered.

## Additional Information

This only happens with audio only .ts segments. .ts segments with video and audio work correctly. This example only has 3 10 seconds .ts segments so the player will stop after 30 seconds.

### videojs-contrib-hls version

what version of videojs-contrib-hls does this occur with?
videojs-contrib-hls 5.10.1

### videojs version

what version of videojs does this occur with?
video.js 6.2.8

### Browsers

what browsers are affected? please include browser and version for each

Google Chrome 60.0.3112.113

### Platforms

what platforms are affected? please include operating system and version or device and version for each

Windows 7,8.1,10, OSX/MacOs

### Other Plugins

None

### Other JavaScript

None

*This issue was moved by [forbesjo](https://github.com/forbesjo) from [videojs/videojs-contrib-hls#1255](https://github.com/videojs/videojs-contrib-hls/issues/1255).*

mister-ben:
## Description
id3 cues are not added correctly to audio-only streams, as the test for an audio-only stream is broken. `videoTimestampOffset()` will never be `null` as it defaults to `0`, so the audio offset is not used.

## Specific Changes proposed
Uses the `videoTimestampOffset()` when `segmentInfo.trackInfo.hasVideo` is `false` rather than when `videoTimestampOffset()` is null

Fixes #130 

## Requirements Checklist
  - [x] Example created (below)

## Example
- Open the [deploy preview](https://deploy-preview-1386--videojs-http-streaming.netlify.app/) and https://videojs-http-streaming.netlify.app to compare.
- In each console, add
  ```
  const printCueData = (cues) => {
    Array.from(cues).forEach((cue) => {
      if ("value" in cue && cue?.value?.data) {
        const value = new TextDecoder()
          .decode(cue.value.data)
          .replace("\x03", "")
          .replace("\x00", "");
        console.log({
          value,
          start: cue.startTime,
          end: cue.endTime,
        });
      }
    });
  };
  
  const onAddTrack = ({ track }) => {
    if (
      !(track?.kind === "metadata" && track.label === "Timed Metadata")
    ) {
      return;
    }
  
    const onCueChange = () => {
      if (!track.activeCues) {
        return;
      }
  
      const cues = printCueData(track.activeCues);
    };
  
    track.mode = "hidden";
    track.addEventListener("cuechange", onCueChange);
  };
  
  player.textTracks().addEventListener("addtrack", onAddTrack);
  ```
- In each load https://cue-points-test.vercel.app/audio/master.m3u8 and play
- Observe cues logged to the console in deploy preview (like `{value: '{"webcastId":"635fc...`)

Thanks to @berrutti for the test streams from [#130](https://github.com/videojs/http-streaming/issues/130#issuecomment-1301796493)

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1873,9 +1873,9 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
-    const timestampOffset = this.sourceUpdater_.videoTimestampOffset() === null ?
-      this.sourceUpdater_.audioTimestampOffset() :
-      this.sourceUpdater_.videoTimestampOffset();
+    const timestampOffset = segmentInfo.trackInfo.hasVideo ?
+      this.sourceUpdater_.videoTimestampOffset() :
+      this.sourceUpdater_.audioTimestampOffset();
 
     // There's potentially an issue where we could double add metadata if there's a muxed
     // audio/video source with a metadata track, and an alt audio with a metadata track.
--- a/test/segment-loader.test.js
+++ b/test/segment-loader.test.js
@@ -2491,6 +2491,9 @@ QUnit.module('SegmentLoader', function(hooks) {
           // Simulate a caption event happening that will call handleCaptions_
           const dispatchType = 0x10;
 
+          // Ensure video and audio offset are different, to know which offset is used
+          loader.sourceUpdater_.videoTimestampOffset(loader.sourceUpdater_.audioTimestampOffset() - 10);
+
           loader.handleId3_(loader.pendingSegment_, metadata, dispatchType);
         });
 
@@ -2508,6 +2511,8 @@ QUnit.module('SegmentLoader', function(hooks) {
           const cue = addCueSpy.getCall(0).args[0];
 
           assert.strictEqual(cue.value.data, 'This is a priv tag', 'included the text');
+
+          assert.strictEqual(cue.startTime, metadata[0].cueTime + loader.sourceUpdater_.audioTimestampOffset(), 'cue.startTime offset from audioTimestampOffset');
           done();
         });
 


adrums86:
My apologies @mister-ben, moving the common metadata functionality up to accommodate EventStream messed up your fix. The timestamp offset ternary now lives in https://github.com/videojs/http-streaming/blob/f6b94987fea5fd75b4c54018ff93c2c23ad2f0a1/src/playlist-controller.js#L2015

I think it could be done without the segmentData like this: `const timestampOffset = this.sourceUpdater_.videoBuffer ?`

mister-ben:
> I think it could be done without the segmentData like this: const timestampOffset = this.sourceUpdater_.videoBuffer ?

That works with the test stream, but there's a `videoBuffer` present in the tests...

adrums86 (approved):
:shipit:

PR #1393, opened 2023-04-27
feat(xhr): add request and response hook API

adrums86:
## Description
This PR would add xhr request and response hooks via an `onRequest(callback)` and `onResponse(callback)` API. The user can add as many unique hook callbacks as they want, and they will be called in the order they were added before the request and after the response is received respectively. The user can un-register these callbacks via the `offRequest(callback)` and `offResponse(callback)` API. The callback functions can take the following parameters:
- `onRequest` callbacks will pass the xhr `request` Object to the callback.
- `onResponse` callbacks will pass the xhr `request, error, and response` Objects to the callback in that order.

In addition, the user can use both **global** `videojs.Vhs` _and_ **player** `player.tech().xhr` scoped API functions to add these hooks, where player hooks will _override_ any global hooks.

Note: `onRequest` hooks are called **_after_** the `beforeRequest` function, so be aware that these hooks may further modify request options passed through that function.

## Specific Changes proposed
Adding API functions to both `Vhs` and the `VhsHandler` for adding and removing unique xhr request and response hooks, passing those hooks to the xhr module, then calling them in the order they were added before an xhr request and in the response callback respectively.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -644,9 +644,9 @@ callback function as a parameter as well as `offRequest` and `offResponse`
 functions which will remove a callback function from the `onRequest` or 
 `onResponse` set if it exists. 
 
-The `onRequest(callback)` function takes a `callback` function that will pass the xhr `request` 
-Object to that callback. These callbacks are called in the order registered and act as pre-request 
-hooks for modifying the xhr `request` Object prior to making a request.
+The `onRequest(callback)` function takes a `callback` function that will pass the xhr `request`
+Object to that callback. These callbacks are called synchronously, in the order registered
+and act as pre-request hooks for modifying the xhr `request` Object prior to making a request.
 
 Example:
 ```javascript
--- a/README.md
+++ b/README.md
@@ -701,7 +701,7 @@ Example
 const globalRequestHook = (request) => {
   const requestUrl = new URL(request.uri);
   requestUrl.searchParams.set('foo', 'bar');
-  request.uri = decodeURIComponent(requestUrl.href);
+  request.uri = requestUrl.href;
 };
 videojs.Vhs.onRequest(globalRequestHook);
 
@@ -709,7 +709,7 @@ videojs.Vhs.onRequest(globalRequestHook);
 const playerRequestHook = (request) => {
   const requestUrl = new URL(request.uri);
   requestUrl.searchParams.set('foo', 'bar');
-  request.uri = decodeURIComponent(requestUrl.href);
+  request.uri = requestUrl.href;
 };
 player.tech().vhs.onRequest(playerRequestHook);
 ```
--- a/README.md
+++ b/README.md
@@ -69,10 +69,6 @@ Video.js Compatibility: 7.x, 8.x
     - [vhs.selectPlaylist](#vhsselectplaylist)
     - [vhs.representations](#vhsrepresentations)
     - [vhs.xhr](#vhsxhr)
-    - [vhs.onRequest](#vhsonrequest)
-    - [vhs.onResponse](#vhsonresponse)
-    - [vhs.offRequest](#vhsoffrequest)
-    - [vhs.offResponse](#vhsoffresponse)
     - [vhs.stats](#vhsstats)
   - [Events](#events)
     - [loadedmetadata](#loadedmetadata)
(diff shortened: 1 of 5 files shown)


src/xhr.js
```diff
@@ -52,7 +52,10 @@ const callbackWrapper = function(request, error, response, callback) {
     error = new Error('XHR Failed with a response of: ' +
                       (request && (reqResponse || request.responseText)));
   }
-
+  // If we have a response hook, call it now.
+  if (beforeResponse && typeof beforeResponse === 'function') {
+    beforeResponse(response);
```
misteroneill, line 57: I think this function should receive the arguments `error, response` for [parity with the XHR module](https://github.com/videojs/xhr/blob/main/README.md#xhrcallback).

src/xhr.js
```diff
@@ -80,7 +84,7 @@ const xhrFactory = function() {
     const xhrMethod = videojs.Vhs.xhr.original === true ? videojsXHR : videojs.Vhs.xhr;
 
     const request = xhrMethod(options, function(error, response) {
-      return callbackWrapper(request, error, response, callback);
+      return callbackWrapper(request, error, response, callback, beforeResponse);
```
misteroneill, line 87: I believe the `callbackWrapper` gets invoked directly in other places in VHS. That means we likely need a different solution because the user-provided response callback needs to be called on all requests.

src/videojs-http-streaming.js
```diff
@@ -1197,6 +1229,22 @@ class VhsHandler extends Component {
       callback
     });
   }
+
+  onRequest(callback) {
+    this.xhr.beforeRequest = callback;
```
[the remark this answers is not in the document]
adrums86: It's actually getting set in the `VhsSourceHandler` object below, in `handleSource` right after `tech.vhs` is set with a `new VhsHandler`. https://github.com/videojs/http-streaming/blob/bdd842a0546b752d165f47fd708c53168a15c0a8/src/videojs-http-streaming.js#L1221

README.md
```diff
@@ -30,7 +30,6 @@ Video.js Compatibility: 7.x, 8.x
 - [Compatibility](#compatibility)
   - [Browsers which support MSE](#browsers-which-support-mse)
   - [Native only](#native-only)
-  - [Flash Support](#flash-support)
```
adrums86, line 33: This removal happened due to adding new API docs for some reason. 🤷‍♂️

README.md
```diff
@@ -677,10 +680,96 @@ player.ready(function() {
...
+the global or player level, where the player level hooks will override global.
+
+
+Example
+```javascript
+// Global
+const globalRequestHook = (request) => {
+  const requestUrl = new URL(request.uri);
+  requestUrl.searchParams.set('foo', 'bar');
+  request.uri = decodeURIComponent(requestUrl.href);
```
misteroneill, line 704: Does this need to be passed to `decodeURIComponent`? 🤔

adrums86: Good catch, it does not, just tested it. Got a little copy paste happy this morning for the docs. 😅

adrums86: Removed

wseymour15 (approved):

misteroneill (approved):

README.md
gkatsev: we probably forgot to run `npm run docs` (or `npm run docs:toc`) previously.

src/videojs-http-streaming.js
```diff
@@ -427,6 +427,64 @@ const expandDataUri = (dataUri) => {
...
 };
 
+/**
+ * Adds a request hook to an xhr object
+ *
+ * @param {Object} xhr object to add the onRequest hook to
+ * @param {function} callback hook function for an xhr request
+ */
+const addOnRequestHook = (xhr, callback) => {
+  if (!xhr.onRequest) {
```
gkatsev, line 437: these probably should be pseudo-private properties. Maybe something like `xhr._requestCallbacks`

src/videojs-http-streaming.js
```diff
@@ -427,6 +427,64 @@ const expandDataUri = (dataUri) => {
...
+};
+
+/**
+ * Adds a response hook to an xhr object
+ *
+ * @param {Object} xhr object to add the onResponse hook to
+ * @param {function} callback hook function for an xhr response
+ */
+const addOnResponseHook = (xhr, callback) => {
+  if (!xhr.onResponse) {
```
gkatsev, line 450: these probably should be pseudo-private properties. Maybe something like `xhr._responseCallbacks`

gkatsev (commented):
A couple minor points.

I think my main question is whether these new methods should be namespaced on the xhr object itself. The main reason is that I think it'll reduce confusion in the future when VHS gets `fetch` support. So, people won't expect these to work when `fetch` is enabled.

dzianis-dashkevich (approved):

dzianis-dashkevich:
I thought about this kind of simplification

```js
xhr._requestHooksSet = new Set();
xhr._responseHooksSet = new Set();

const addOnRequestHook = (callback) => xhr._requestHooksSet.add(callback);
const addOnResponseHook = (callback) => xhr._responseHooksSet.add(callback);

const removeOnRequestHook = (callback) => xhr._requestHooksSet.delete(callback);
const removeOnResponseHook = (callback) => xhr._responseHooksSet.delete(callback);
```

but then I realized that those 2 XHRs are not the same pointers...

LGTM!

adrums86:
@gkatsev I totally agree regarding namespacing these to the xhr object to separate the hooks from fetch. Update incoming.

adrums86:
@gkatsev and @misteroneill the namespace changes are pushed. Another quick look would be appreciated before merging. Thanks!

misteroneill (approved):
:shipit:

README.md
```diff
@@ -637,12 +636,62 @@ player.tech().vhs.representations().forEach(function(rep) {
...
-the ability to specify a `beforeRequest` function that will be called
-with an object containing the options that will be used to create the
-xhr request.
+the function with your own implementation. Instead, `xhr` provides
+the ability to specify `onRequest` and `onResponse` hooks which take a 
+callback function as a parameter as well as `offRequest` and `offResponse` 
+functions which will remove a callback function from the `onRequest` or 
+`onResponse` set if it exists. 
+
+The `onRequest(callback)` function takes a `callback` function that will pass the xhr `request`
```
gkatsev, line 647: should this mention that these callbacks are called synchronously?

gkatsev (approved):

PR #1396, opened 2023-05-05
fix: onRequest hooks called too late

adrums86:
## Description
The xhr wrapper calls `open` and `send` when a request object is instantiated, this is causing `onRequest` hooks to modify the xhr object too late, as they were already sent when the object was returned to the VHS xhr module. Additionally, setting xhr hooks in time for the initial manifest request when `preload` is set to `auto` or `metadata` is not possible with existing events or `ready` callbacks.

### Moving the request hooks callback and changing the callback parameter
By calling all the `onRequest` hooks **before** we create the `request` object and passing the callback the `options` object instead, we can modify the request before `open` is called in the xhr wrapper. This `options` object allows for modifying the url/headers/methods, as well as setting a `beforeSend` callback, which gives the `onRequest` callback direct access to the `xhr` object via a nested callback. `beforeSend` will be called immediately before `xhr.send` is called, which allows for full control of the xhr object prior to sending the request.

For example, setting a request header.
```js
const playerXhrRequestHook = (options) => {
  options.beforeSend = (xhr) => {
    xhr.setRequestHeader('foo', 'bar');
  };
};
player.tech().vhs.xhr.onRequest(playerXhrRequestHook);
```
See all supported options: https://github.com/videojs/xhr#options

### Adding an xhr-hooks-ready event
This fixes a `preload` option issue in VHS, where if metadata is preloaded setting `onRequest` hooks is not possible on `player.ready` because `player.tech().vhs` doesn't exist yet as it's created in `handleSources`. Also if the user tries to set `onRequest` hooks on the `loadstart` event, it's too late and the callback will not be set before the manifest is requested. Thus, a new event is needed to be fired immediately when the hooks are available, which allows the user to apply these hooks to all requests, including the main manifest.

### Deprecating xhr.beforeRequest
Since adding an `onRequest` hooks is functionally the same as setting a `xhr.beforeRequest` callback,  a deprecation warning will log when setting a `beforeRequest` function. `beforeRequest` will also now alias to an `onRequest` callback.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -646,8 +646,7 @@ functions which will remove a callback function from the `onRequest` or
 
 The `onRequest(callback)` function takes a `callback` function that will pass the xhr `options`
 Object to that callback. These callbacks are called synchronously, in the order registered 
-and act as pre-request hooks for modifying the xhr `options` Object prior to making a request. 
-https://github.com/videojs/xhr#options
+and act as pre-request hooks for modifying the xhr `options` Object prior to making a request.
 
 Example:
 ```javascript
--- a/README.md
+++ b/README.md
@@ -701,25 +701,9 @@ Example:
 ```javascript
 player.tech().vhs.xhr.offResponse(playerResponseHook);
 ```
-Additionally a `beforeRequest` function can be defined, 
-that will be called with an object containing the options that will be used 
-to create the xhr request.
 
-Note: any registered `onRequest` hooks, are called _after_ the `beforeRequest` function, so xhr
-options modified by this function may be further modified by these hooks.
-
-Example:
-```javascript
-player.tech().vhs.xhr.beforeRequest = function(options) {
-  options.uri = options.uri.replace('example.com', 'foo.com');
-
-  return options;
-};
-```
-
-The global `videojs.Vhs` also exposes an `xhr` property. Adding
-`onRequest`, `onResponse` hooks and/or specifying a `beforeRequest` 
-function that will allow you to intercept the request options, xhr 
+The global `videojs.Vhs` also exposes an `xhr` property. Adding `onRequest` 
+and/or `onResponse` hooks will allow you to intercept the request options, xhr 
 Object, error and response data for *all* requests in every player on a page. 
 For consistency across browsers the video source should be set at runtime 
 once the video player is ready.
@@ -762,24 +746,6 @@ videojs.Vhs.xhr.offRequest(globalRequestHook);
 ```javascript
 // Remove a global onResponse callback.
 videojs.Vhs.xhr.offResponse(globalResponseHook);
-```
-
-```javascript
-videojs.Vhs.xhr.beforeRequest = function(options) {
-  /*
-   * Modifications to requests that will affect every player.
-   */
-
-  return options;
-};
-
-var player = videojs('video-player-id');
-player.ready(function() {
-  this.src({
-    src: 'https://d2zihajmogu5jn.cloudfront.net/bipbop-advanced/bipbop_16x9_variant.m3u8',
-    type: 'application/x-mpegURL',
-  });
-});
 ```
 
 For information on the type of options that you can modify see the
--- a/README.md
+++ b/README.md
@@ -644,18 +644,33 @@ callback function as a parameter as well as `offRequest` and `offResponse`
 functions which will remove a callback function from the `onRequest` or 
 `onResponse` set if it exists. 
 
-The `onRequest(callback)` function takes a `callback` function that will pass the xhr `request`
-Object to that callback. These callbacks are called synchronously, in the order registered
-and act as pre-request hooks for modifying the xhr `request` Object prior to making a request.
+The `onRequest(callback)` function takes a `callback` function that will pass the xhr `options`
+Object to that callback. These callbacks are called synchronously, in the order registered 
+and act as pre-request hooks for modifying the xhr `options` Object prior to making a request. 
+https://github.com/videojs/xhr#options
 
 Example:
 ```javascript
-const playerRequestHook = (request) => {
-  const requestUrl = new URL(request.uri);
+const playerRequestHook = (options) => {
+  const requestUrl = new URL(options.uri);
   requestUrl.searchParams.set('foo', 'bar');
-  request.uri = requestUrl.href;
+  options.uri = requestUrl.href;
 };
 player.tech().vhs.xhr.onRequest(playerRequestHook);
+```
+
+If access to the `xhr` Object is required prior to the `xhr.send` call, an `options.beforeSend` 
+callback can be set within an `onRequest` callback function that will pass the `xhr` Object 
+as a parameter and will be called immediately prior to `xhr.send`.
+
+Example:
+```javascript
+const playerXhrRequestHook = (options) => {
+  options.beforeSend = (xhr) => {
+    xhr.open('GET', 'https://new.uri');
+  };
+};
+player.tech().vhs.xhr.onRequest(playerXhrRequestHook);
 ```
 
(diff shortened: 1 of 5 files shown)


misteroneill (commented):
We've talked about this offline, but it's just now occurring to me... is there functionally any difference between the `onRequest` hook and the `beforeRequest` thing that already existed?

adrums86:
@misteroneill It's definitely a bit of hair splitting at this point, but there seems to be no real functional difference. Especially with the revelation that the `options` object can pass a `beforeSend` callback which provides access to the xhr object. We could properly differentiate these by modifying the XHR library, but that is also another can-o-worms. I think the biggest fix here is the event. `beforeRequest` has the same timing issue as the hooks API.

gkatsev:
if `beforeRequest` and `onRequest` are basically functionality equivalent, we could deprecate `beforeRequest` to have consistent naming with `onRequest` and `onResponse`. Also, completely missed that `beforeSend` was a thing that could be provided to the options object. We probably could've saved a lot of headache if we realized that ahead of time, haha.

README.md
```diff
@@ -704,22 +719,32 @@ player.tech().vhs.xhr.beforeRequest = function(options) {
...
+  options.uri = requestUrl.href;
 };
 videojs.Vhs.xhr.onRequest(globalRequestHook);
 ```
 
+```javascript
+// Global request callback defining beforeSend function, will affect every player.
+const globalXhrRequestHook = (options) => {
+  options.beforeSend = (xhr) => {
+    xhr.open('GET', 'https://new.uri');
```
gkatsev, line 742: is this recommended? Since I'd assume `xhr` would try and call `open` after itself?

gkatsev (commented):
should `beforeRequest` alias `onRequest`?

README.md
misteroneill: Good point. Should probably do... something else. Maybe an equivalent example of setting a header?

adrums86: Yeah I wouldn't say I'd recommend it 😅 , it does work however. According to MDN it re-initializes the existing request with every subsequent `open` call. Probably not ideal if applied to _every_ request. I'll think up a better example for the docs.

adrums86:
Yeah I kind of figured this was headed toward the depreciation and/or aliasing of `beforeRequest`, for some reason I didn't think to explicitly call that out until now. I'll add a warning to `beforeRequest` and create an alias to `onRequest`.

README.md
adrums86: Fixed. Used `setRequestHeader` instead.

adrums86:
Added a deprecation warning to `beforeRequest` and aliased the function to an `onRequest` callback. Also removed `beforeRequest` from the docs.

src/xhr.js
```diff
@@ -82,23 +82,29 @@ const xhrFactory = function() {
...
+      videojs.log.warn('beforeRequest is deprecated, use onRequest instead.');
+      _requestCallbackSet.add(beforeRequest);
     }
 
     // Use the standard videojs.xhr() method unless `videojs.Vhs.xhr` has been overriden
     // TODO: switch back to videojs.Vhs.xhr.name === 'XhrFunction' when we drop IE11
     const xhrMethod = videojs.Vhs.xhr.original === true ? videojsXHR : videojs.Vhs.xhr;
 
+    // call all registered onRequest hooks
+    callAllHooks(_requestCallbackSet, options);
```
misteroneill, line 102: It looks like `options` is passed here but no return value is used, which would be a change from the above removed behavior where `beforeRequest` returned a new options object. It seems that this new expectation is that each request callback would reference the same options object - each potentially modifying the same object in sequence.

I don't think this is a bad implementation, but it could be a breaking change to `beforeRequest`. Consider this would likely fail with this refactor:

```js
videojs.Vhs.xhr.beforeRequest = (options) => {
  // Ignores `options` and returns a totally new object.
  return {...};
};
```
We probably don't want that to break, so perhaps the `callAllHooks` should have special handling for the `beforeRequest`/`onRequest` hooks where it passes the return value of the previous hook through a chain and returns the last result?

misteroneill (changes requested):

src/xhr.js
adrums86: Great catch @misteroneill ! I added some logic to support both cases and a test for the use case you described. Let me know what you think.

src/xhr.js
```diff
@@ -60,17 +60,23 @@ const callbackWrapper = function(request, error, response, callback) {
...
-  if (!hooks) {
+  if (!hooks || !hooks.size) {
     return;
   }
+  let hookOptions;
+
   hooks.forEach((hookCallback) => {
-    hookCallback(request, error, response);
+    // Pass the returned options back to the next hook callback to support a callback returning a new options object.
+    hookOptions = hookCallback(hookOptions || request, error, response);
```
adrums86, line 77: Open question here and below: Do we want to `merge` options here instead?

adrums86:
Couple of open question I would appreciate some feedback on per the latest changes.

1. Is it better to `merge` or `||` (or) the options returned from `beforeRequest` and/or hooks?
2. Should we handle `onRequest` and `onResponse` hooks separately, or still use `callAllHooks` for both?

README.md
```diff
@@ -644,20 +645,34 @@ callback function as a parameter as well as `offRequest` and `offResponse`
...
-and act as pre-request hooks for modifying the xhr `request` Object prior to making a request.
+The `onRequest(callback)` function takes a `callback` function that will pass the xhr `options`
+Object to that callback. These callbacks are called synchronously, in the order registered 
+and act as pre-request hooks for modifying the xhr `options` Object prior to making a request.
 
 Example:
 ```javascript
-const playerRequestHook = (request) => {
-  const requestUrl = new URL(request.uri);
+const playerRequestHook = (options) => {
```
misteroneill, line 654: One thought on these samples, should they be wrapped in an `xhr-hooks-ready` event callback to show the recommended approach?

```js
player.on('xhr-hooks-ready', () => {
  player.tech().vhs.xhr.onRequest((options) => options);
});
```

README.md
```diff
@@ -644,20 +645,34 @@ callback function as a parameter as well as `offRequest` and `offResponse`
...
+  const requestUrl = new URL(options.uri);
   requestUrl.searchParams.set('foo', 'bar');
-  request.uri = requestUrl.href;
+  options.uri = requestUrl.href;
```
misteroneill, line 657: Will need to update these samples to return the `options` object (or a new object) in the same pattern as `beforeRequest`.

It would make sense to document the behavior here - each `onRequest` hook should return an options object and each will receive the previous hook's options object.

Also, we should document that the `onResponse` hooks do not behave this way - the `request` object is passed to each by reference.

README.md
adrums86: Good call, will add examples with the event.

README.md
adrums86: Yeah I wasn't 100% sure that we wanted to document it that way as it wasn't clear whether we would remove the `options` return/chaining after `beforeRequest` is removed. I agree it makes sense if we keep it that way for `onRequest` hooks.

adrums86: Added `options` return to the `onRequest` examples. Documented `onResponse` passing parameters by reference as well.

README.md
adrums86: Added the `'xhr-hooks-ready'` event listener to all the per-player examples.

adrums86:
Ended up splitting `callAllHooks` into 2 functions to better accommodate the requirement for `onRequest` callbacks to return an `options` object.

misteroneill (approved):

PR #1402, opened 2023-05-25
feat: add daterange support

harisha-swaminathan:
## Description
Adds support for [#EXT-X-DATERANGE](https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.7) tags. This PR exposes parsed daterange data from [this m3u8 parser PR](https://github.com/videojs/m3u8-parser/pull/163) as a metadata text track (as Safari does).

## Specific Changes proposed
The PR adds two new functions 

- `addDaterangeToTextTrack` (similar to `addMetadataToTextTrack`) to create the `daterange-metadata` text track
-  `addDaterangeMetadata` (similar to `addMetadata`) to create cues, calculate startTime and endTime of the cue and add cue to the track

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -26,7 +26,7 @@ import { codecsForPlaylist, unwrapCodecList, codecCount } from './util/codecs.js
 import { createMediaTypes, setupMediaGroups } from './media-groups';
 import logger from './util/logger';
 import {merge, createTimeRanges} from './util/vjs-compat';
-import { addMetadata, createMetadataTrackIfNotExists, addDaterangeMetadata } from './util/text-tracks';
+import { addMetadata, createMetadataTrackIfNotExists, addDateRangeMetadata } from './util/text-tracks';
 
 const ABORT_EARLY_EXCLUSION_SECONDS = 60 * 2;
 
@@ -265,7 +265,7 @@ export class PlaylistController extends videojs.EventTarget {
     // PlaylistLoader should be used.
     this.mainPlaylistLoader_ = this.sourceType_ === 'dash' ?
       new DashPlaylistLoader(src, this.vhs_, merge(this.requestOptions_, { addMetadataToTextTrack: this.addMetadataToTextTrack.bind(this) })) :
-      new PlaylistLoader(src, this.vhs_, merge(this.requestOptions_, { addDaterangeToTextTrack: this.addDaterangeToTextTrack.bind(this) }));
+      new PlaylistLoader(src, this.vhs_, merge(this.requestOptions_, { addDateRangesToTextTrack: this.addDateRangesToTextTrack_.bind(this) }));
     this.setupMainPlaylistLoaderListeners_();
 
     // setup segment loaders
@@ -2024,12 +2024,11 @@ export class PlaylistController extends videojs.EventTarget {
     return Config.BUFFER_HIGH_WATER_LINE;
   }
 
-  addDaterangeToTextTrack(mediaPlaylist) {
-
+  addDateRangesToTextTrack_(dateRanges) {
     createMetadataTrackIfNotExists(this.inbandTextTracks_, 'com.apple.streaming', this.tech_);
-    addDaterangeMetadata({
+    addDateRangeMetadata({
       inbandTextTracks: this.inbandTextTracks_,
-      mediaPlaylist
+      dateRanges
     });
   }
 
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -256,8 +256,7 @@ export class PlaylistController extends videojs.EventTarget {
       sourceUpdater: this.sourceUpdater_,
       timelineChangeController: this.timelineChangeController_,
       exactManifestTimings: options.exactManifestTimings,
-      addMetadataToTextTrack: this.addMetadataToTextTrack.bind(this),
-      addDaterangeToTextTrack: this.addDaterangeToTextTrack.bind(this)
+      addMetadataToTextTrack: this.addMetadataToTextTrack.bind(this)
     };
 
     // The source type check not only determines whether a special DASH playlist loader
@@ -2025,7 +2024,7 @@ export class PlaylistController extends videojs.EventTarget {
     return Config.BUFFER_HIGH_WATER_LINE;
   }
 
-  addDaterangeToTextTrack(metadataArray) {
+  addDaterangeToTextTrack(mediaPlaylist) {
     const timestampOffset = this.sourceUpdater_.videoBuffer ?
       this.sourceUpdater_.videoTimestampOffset() : this.sourceUpdater_.audioTimestampOffset();
 
@@ -2035,7 +2034,7 @@ export class PlaylistController extends videojs.EventTarget {
     }, false).track;
     addDaterangeMetadata({
       inbandTextTracks: this.inbandTextTracks_,
-      metadataArray,
+      mediaPlaylist,
       timestampOffset
     });
   }
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -26,7 +26,7 @@ import { codecsForPlaylist, unwrapCodecList, codecCount } from './util/codecs.js
 import { createMediaTypes, setupMediaGroups } from './media-groups';
 import logger from './util/logger';
 import {merge, createTimeRanges} from './util/vjs-compat';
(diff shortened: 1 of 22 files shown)


src/util/text-tracks.js
```diff
@@ -223,6 +223,59 @@ export const addMetadata = ({
...
+    scte35In: 'SCTE35-IN'
+  };
+
+  if (!metadataTrack) {
+    return;
+  }
+
+  metadataArray.daterange.forEach((metadata) => {
+    const startTime = timestampOffset + (new Date(metadata.startDate).getTime() - new Date(metadataArray.dateTimeObject).getTime()) / 1000;
+    const endTime = startTime + parseFloat(metadata.plannedDuration || metadata.duration) || 0;
```
adrums86, line 263: I think we may want to apply the `dateTimeObject` offset to the end time as well and maybe not use the `plannedDuration` for the endTime calculation. HLS.js is doing this: https://github.com/video-dev/hls.js/blob/78b50130bbacb268f3400a8211f6fe9580afb0fa/src/controller/id3-track-controller.ts#L322

src/playlist-loader.js
```diff
@@ -404,6 +405,11 @@ export default class PlaylistLoader extends EventTarget {
     // live playlist staleness timeout
     this.handleMediaupdatetimeout_ = this.handleMediaupdatetimeout_.bind(this);
     this.on('mediaupdatetimeout', this.handleMediaupdatetimeout_);
+    this.one('loadedplaylist', ()=>{
```
adrums86, line 408: nit: spacing.

adrums86 (commented):
This looks great so far. I think we want to play with the `endTime` calculation and see how it lines up against Safari with native HLS playback.

src/util/text-tracks.js
harisha-swaminathan: Updated endTime value in the latest commit. 
What should the endTime value be when there's no endDate attribute?

adrums86: Good question, I think 0 is probably fine, but I think we will also want to account for `END-ON-NEXT` tags.

adrums86 (commented):
I think once we account for `END-ON-NEXT` + testing for that logic and this will be good to go. Nice work!

src/util/text-tracks.js
```diff
@@ -223,6 +223,60 @@ export const addMetadata = ({
...
+    scte35In: 'SCTE35-IN'
+  };
+
+  if (!metadataTrack) {
+    return;
+  }
+
+  metadataArray.daterange.forEach((metadata) => {
+    const startTime = timestampOffset + (new Date(metadata.startDate).getTime() - new Date(metadataArray.dateTimeObject).getTime()) / 1000;
+    const endTime = metadata.endDate ? (new Date(metadata.endDate).getTime() - new Date(metadataArray.dateTimeObject).getTime()) / 1000 : 0;
```
dzianis-dashkevich, line 263: IMHO, here are a few naming ideas:
- Instead of`metadataArray`, the better name is `mediaPlaylist`, because it is parsed media playlist and not an array of metadata.
- Instead of `daterange`, the better name is `dateRanges` because it is literally `Array<DateRange>` and it is semantically right to say `dateRanges.map((dateRange) => ....)`

Regarding `endTime` I would assume something like this (we can use backward loop to optimize performance, but I keep 2 for loops for simplicity):

```javascript
    const localDateRanges = [...mediaPlaylist.dateRanges];
    const dateRangeClasses = {};

    for (let dateRange of localDateRanges) {
        const startDateTimeStamp = new Date(metadata.startDate).getTime();
        const startTime = timestampOffset + (startDateTimeStamp - dateTimeTimeStamp) / 1000;

        dateRange.startTime = startTime;

        if (dateRange.class) {
            if (dateRangeClasses[dateRange.class]) {
                dateRangeClasses[dateRange.class].push(dateRange);
            } else {
                dateRangeClasses[dateRange.class] = [dateRange];
            }
        }
    }

    for (let dateRange of localDateRanges) {
        let endTime;
        const classList = dateRangeClasses[dateRange.class];
        const classListIndex = classList.findIndex(dateRange);

        if (dateRange.endDate) {
            // Use endDate directly
            const endDateTimeStamp = new Date(metadata.startDate).getTime();
            endTime = timestampOffset + (endDateTimeStamp - dateTimeTimeStamp) / 1000;
        } else if (dateRange.endOnNext && classList[classListIndex + 1]) {
            // Fallback to the startTime of the next dateRange with the same class
            const { startTime } = classList[classListIndex + 1];
            endTime = startTime;
        } else if (dateRange.duration) {
            // Fallback to startTime + duration
            endTime = dateRange.startTime + dateRange.duration;
        } else if (dateRange.plannedDuration) {
            // Fallback to startTime + plannedDuration
            endTime = dateRange.startTime + dateRange.plannedDuration;
        } else {
            // Fallback to 0
            endTime = 0;
        }

        // generate and add cue
    }
```

src/util/text-tracks.js
```diff
@@ -223,6 +223,60 @@ export const addMetadata = ({
...
+
+    const cue = new Cue(startTime, endTime, '');
+
+    cue.id = metadata.id;
+    Object.keys(metadata).forEach((key) => {
+      if (!['id', 'class', 'startDate', 'duration', 'endDate', 'endOnNext'].includes(key)) {
+        cue.value = {keys: daterangeAttr[key], data: metadata[key]};
```
dzianis-dashkevich, line 270: Should we map all hex values (eg: `SCTE-35`) to `ArrayBuffer` to mimic native Safari?

src/util/text-tracks.js
```diff
@@ -223,6 +223,60 @@ export const addMetadata = ({
...
+    id: 'ID',
+    class: 'CLASS',
+    startDate: 'START-DATE',
+    duration: 'DURATION',
+    endDate: 'END-DATE',
+    endOnNext: 'END-ON-NEXT',
+    plannedDuration: 'PLANNED-DURATION',
+    scte35Out: 'SCTE35-OUT',
+    scte35In: 'SCTE35-IN'
+  };
```
dzianis-dashkevich, line 255: I guess we can move this object outside of the function cause it is static

src/playlist-controller.js
```diff
@@ -2024,6 +2025,21 @@ export class PlaylistController extends videojs.EventTarget {
     return Config.BUFFER_HIGH_WATER_LINE;
   }
 
+  addDaterangeToTextTrack(metadataArray) {
+    const timestampOffset = this.sourceUpdater_.videoBuffer ?
+      this.sourceUpdater_.videoTimestampOffset() : this.sourceUpdater_.audioTimestampOffset();
```
dzianis-dashkevich, line 2030: Is it possible to have `dateRanges` from multiple discontinuities?

src/util/text-tracks.js
harisha-swaminathan: Thank you! Did the renaming and updated endTime in the latest commit.

src/util/text-tracks.js
harisha-swaminathan: Updated endTime calculation and added tests in the latest commit

src/playlist-loader.js
```diff
@@ -404,6 +405,11 @@ export default class PlaylistLoader extends EventTarget {
     // live playlist staleness timeout
     this.handleMediaupdatetimeout_ = this.handleMediaupdatetimeout_.bind(this);
     this.on('mediaupdatetimeout', this.handleMediaupdatetimeout_);
+    this.one('loadedplaylist', () => {
```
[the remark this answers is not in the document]
dzianis-dashkevich: More context: I think we should use `on` to handle live scenarios as well and manage cues using `addCue` and `removeCue`.

src/playlist-controller.js
dzianis-dashkevich: More context: I see we are using `timestampOffset` discontinuities will have different `timestampOffset`s.

My thoughts: I am not sure we should use timestampOffset for dateRanges

adrums86 (commented):
@harisha-swaminathan This is awesome! Very excited to get this merged. Can you add a couple tests for `scte35In` signals? I see a few things in the spec that are probably worthy of at least adding a test. Specifically in/out pairs and maybe duration.

adrums86 (approved):
:shipit:

PR #1408, opened 2023-07-21
feat: enable caption positioning

wseymour15:
## Description
This consumes the work completed in mux.js here: https://github.com/videojs/mux.js/pull/434

The goal of this work is to ensure that we can position CAE 608 captions. The bulk of the work was done in mux.js, but this just ensures that we take that data and turn them into valid `VTTCue`s . This will ensure that the captions are positioned correctly based on that CAE 608 caption data.

## To Test
Pull down this branch and run `npm i` and `npm start`, and then navigate to http://localhost:9999. In the closed caption menu, choose the CC1 option and see that the captions are positioned in different places on the screen.

## Specific Changes proposed
- Package updates to mux.js and video.js
- Ensuring we add the correct properties to VTTCues per caption sent to VHS.
- Tests regarding the above

## Requirements Checklist
  - [ ] Docs/guides updated


src/util/text-tracks.js
```diff
@@ -82,11 +82,31 @@ export const addCaptionData = function({
...
+        const cue = new Cue(
+          caption.startTime + timestampOffset,
+          caption.endTime + timestampOffset,
+          value.text
+        );
+
+        cue.line = value.line;
+        cue.align = 'left';
+        cue.position = value.position;
+        cue.positionAlign = 'line-left';
```
misteroneill, line 98: Forgive my ignorance of captions and tracks, but is `line-left` here going to make us unable to support RTL languages OOTB?

misteroneill (commented):
Just a question!

src/util/text-tracks.js
wseymour15: This is a good question! My best attempt to answer this will come from [the spec for CEA 608 captions](https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#positioning-in-cea-608), where it states: "CEA-608 does not distinguish between different alignment, so all captions are left aligned and require a WebVTT align cue setting to override the default middle alignment".

My interpretation of this is that there RTL languages _should_ still start at the left by default. If the desire for those tracks was to be right aligned, it is always possible to include that positioning data in the VTT files so it will align right.

misteroneill: Let the spec guide us!

misteroneill (approved):

src/util/text-tracks.js
mister-ben: CEA-608 supports only a limited Latin character set so doesn't have any consideration of RTL languages. 708 however supports Unicode.

PR #1407, opened 2023-07-10
chore: update m3u8-parser version and fix tests

harisha-swaminathan:
Updates m3u8-parser version to 7.0, replaces `dateTimeObject` and `dateTimeString` with `programDateTime` and  fixes tests

PR #1409, opened 2023-07-24
refactor: add dateTimeObject and dateTimeString to Cues for backwards compatibility

harisha-swaminathan:
Adds `dateTimeObject` and `dateTimeString` back to segment metadata cues for backwards compatibility

### DIFF
--- a/src/util/time.js
+++ b/src/util/time.js
@@ -74,7 +74,7 @@ export const findSegmentForProgramTime = (programTime, playlist) => {
 
   let segment = playlist.segments[0];
 
-  if (dateTimeObject < segment.programDateTime) {
+  if (dateTimeObject < new Date(segment.programDateTime)) {
     // Requested time is before stream start.
     return null;
   }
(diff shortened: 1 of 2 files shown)


src/util/time.js
```diff
@@ -74,15 +74,15 @@ export const findSegmentForProgramTime = (programTime, playlist) => {
 
   let segment = playlist.segments[0];
 
-  if (dateTimeObject < segment.programDateTime) {
+  if (dateTimeObject < new Date(segment.programDateTime)) {
```
wseymour15, line 77: I'm assuming this is unrelated to adding back the old properties. Just for my understanding, what is the purpose for these changes? I am sure there is a good reason :)

harisha-swaminathan: So this is related to the changes I made in the original PR that replaced dateTimeObject/String with `programDateTime` here https://github.com/videojs/http-streaming/pull/1407/files#diff-ba19ddab36415030a61221adf95d07ad0ec76ce6453cdd837b69285eb4533974L77

With the changes I made in that PR, we were comparing `dateTimeObject` (a date object)  and `programDateTime` (a timestamp represented in ms) and it works fine but just wanted them both to be of the same type just to be safe.

wseymour15: Makes sense, thanks!

wseymour15 (approved):

PR #1411, opened 2023-07-25
fix(live): cue end time not finite

adrums86:
## Description
When creating metadata cues, the last cue uses the videoDuration as the end time, in live playback this is `Infinity` and throws an error.

## Specific Changes proposed
Add an `isFinite` check to the duration, if it's not finite we use 0 instead.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/util/text-tracks.js
+++ b/src/util/text-tracks.js
@@ -234,7 +234,8 @@ export const addMetadata = ({
   // Map each cue group's endTime to the next group's startTime
   sortedStartTimes.forEach((startTime, idx) => {
     const cueGroup = cuesGroupedByStartTime[startTime];
-    const nextTime = Number(sortedStartTimes[idx + 1]) || videoDuration;
+    const finiteDuration = isFinite(videoDuration) ? videoDuration : 0;
+    const nextTime = Number(sortedStartTimes[idx + 1]) || finiteDuration;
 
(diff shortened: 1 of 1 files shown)


misteroneill (approved):

wseymour15 (approved):
Looks good!

src/util/text-tracks.js
```diff
@@ -234,7 +234,8 @@ export const addMetadata = ({
   // Map each cue group's endTime to the next group's startTime
   sortedStartTimes.forEach((startTime, idx) => {
     const cueGroup = cuesGroupedByStartTime[startTime];
-    const nextTime = Number(sortedStartTimes[idx + 1]) || videoDuration;
```
dzianis-dashkevich, line 237: Sorry for commenting on merged PR, but maybe we should consider `endTime`= `startTime` instead of 0 in this case?

adrums86: Good catch. Just saw a link to this from another issue. Will put up a quick PR for that fix, I agree start time is more appropriate.

robert-schurig-mi: Sry to comment this, but when the endTime is set to the starttime or "0".... the cue doesn't get active anymore.

PR #1412, opened 2023-08-04
fix: restore dateTimeObject and dateTimeString usage

harisha-swaminathan:
This PR shouldn't be merged before 

- [x] the corresponding [m3u8-parser PR](https://github.com/videojs/m3u8-parser/pull/174) is merged and released
- [ ] m3u8-parser is updated to the latest version containing the above PR
- [ ] tests pass 

## Description
Restores dateTimeObject and dateTimeString segment properties that were previously removed and refactored in these two PRs

- https://github.com/videojs/http-streaming/pull/1407
- https://github.com/videojs/http-streaming/pull/1409

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -421,7 +421,7 @@ export default class PlaylistLoader extends EventTarget {
     this.dateRangesStorage_.setPendingDateRanges(mediaPlaylist.dateRanges);
     const availableDateRanges = this.dateRangesStorage_.getDateRangesToProcess();
 
-    if (!availableDateRanges.length) {
+    if (!availableDateRanges.length || !this.addDateRangesToTextTrack_) {
       return;
     }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -6299,9 +6299,9 @@
       }
     },
     "m3u8-parser": {
-      "version": "7.0.0",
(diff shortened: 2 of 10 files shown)


adrums86 (commented):
LGTM. Obviously lets get the parser update in here then +1 from me.

src/playlist-loader.js
```diff
@@ -421,7 +421,7 @@ export default class PlaylistLoader extends EventTarget {
     this.dateRangesStorage_.setPendingDateRanges(mediaPlaylist.dateRanges);
     const availableDateRanges = this.dateRangesStorage_.getDateRangesToProcess();
 
-    if (!availableDateRanges.length) {
+    if (!availableDateRanges.length || !this.addDateRangesToTextTrack_) {
```
harisha-swaminathan, line 424: Noticed that there were several `!this.addDateRangesToTextTrack_ is not a function` errors when I tested dateRanges with a specific stream. And noticed that the function altered between being undefined and being available on each `loadedplaylist` event (line 410). Not sure why this is happening, so added a quick fix before debugging in detail.

adrums86: This seems to be because the test stream has subtitles, which in this case we create a `PlaylistLoader` for, this does not have the `addDateRangesToTextTrack_` function defined, as it is only needed for the `mainPlaylistLoader` for HLS.

adrums86 (approved):
:shipit:

PR #1413, opened 2023-08-07
fix: reduce playlist exclusion defaults

adrums86:
## Description
The default playlist exclusion duration seemed arbitrarily long, especially for live playback. Lets reduce the default and the `earlyAbortWhenNeeded_` exclusion to seconds instead of minutes to allow for more live resiliency.

## Specific Changes proposed
Reduce the default `playlistExclusionDuration` value from 5 minutes to 10 seconds. Also reduce the `ABORT_EARLY_EXCLUSION_SECONDS` from 2 minutes to 10 seconds.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -28,7 +28,7 @@ import logger from './util/logger';
 import {merge, createTimeRanges} from './util/vjs-compat';
 import { addMetadata, createMetadataTrackIfNotExists, addDateRangeMetadata } from './util/text-tracks';
 
-const ABORT_EARLY_EXCLUSION_SECONDS = 60 * 2;
+const ABORT_EARLY_EXCLUSION_SECONDS = 10;
 
 let Vhs;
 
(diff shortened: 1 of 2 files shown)


harisha-swaminathan (dismissed):

dzianis-dashkevich:
Should we mark it as a draft if it requires more investigation? @adrums86 @harisha-swaminathan

adrums86:
@dzianis-dashkevich yeah good question. The [original PR](https://github.com/videojs/videojs-contrib-hls/pull/1220) cites preventing a cache loop, but the time still seems arbitrary. I haven't seen any adverse effects lowering this value while testing. I'm starting to think this is likely a safe change that will net more stability for the average user than any risk assumed by lowering the values significantly.

adrums86:
@gesinger I know the original fix was an eternity ago, but I figure it's worth pinging you to see if you have any thoughts on reducing these values?

dzianis-dashkevich:
Maybe we should consider some hybrid approach: like introducing `min=10` and `max=Infinity` values.
On each failed load - increase exclude by some factor `excludeTime = previousExcludeTime * factor`
and on every success load - reset `previousExcludeTime` back to `min`

dzianis-dashkevich:
sure, lets merge it

dzianis-dashkevich (approved):

harisha-swaminathan (approved):

PR #1414, opened 2023-08-08
fix: fastQualityChange refactor

adrums86:
## Description
`fastQualityChange_` was causing unnecessary rebuffering and requesting cached segments. This was resulting in slow and very buggy playlist changes. This was because it was calling `resetEverything` on the mainSegmentLoader, then calling `setCurrentTime` which calls `resetEverything` again on ALL the segment loaders, which removes all content from the buffer and aborts all requests.

Edit: Had to add a property `replaceSegmentsUntil`, which represents the end of the buffer at the moment `fastQualityChange_` is called. This tells the segment loader to leave `fetchAtBuffer_` as false until we've successfully replaced all the buffered segments between `currentTime` and `bufferedEnd` at the moment `fastQualityChange_` is called.

## Specific Changes proposed
Remove the `resetEverything` and `setCurrentTime` call from `fastQualityChange_` and instead reset the next segment position to `currentTime` with a `resetLoader` call, keeping the current buffer and overwriting it.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -949,7 +949,7 @@ export class PlaylistController extends videojs.EventTarget {
     // Set the replace segments flag to the buffered end, this forces fetchAtBuffer
     // on the main loader to remain, false after the resetLoader call, until we have
     // replaced all content buffered ahead of the currentTime.
-    this.mainSegmentLoader_.replaceSegmentsUntil_ = bufferedEnd;
+    this.mainSegmentLoader_.replaceSegmentsUntil = bufferedEnd;
     this.mainSegmentLoader_.resetLoaderProperties();
     this.mainSegmentLoader_.resetLoader();
   }
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -933,7 +933,10 @@ export class PlaylistController extends videojs.EventTarget {
 
     this.switchMedia_(media, 'fast-quality');
 
-    // Reset main segment loader properties and next segment position information
+    // Reset main segment loader properties and next segment position information.
+    // Don't need to reset audio as it is reset when media changes.
+    // We resetLoaderProperties separately here as we want to fetch init segments if
+    // necessary and ensure we're not in an ended state when we switch playlists.
     this.mainSegmentLoader_.resetLoaderProperties();
     this.mainSegmentLoader_.resetLoader();
   }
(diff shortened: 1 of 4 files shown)


src/playlist-controller.js
```diff
@@ -933,17 +933,9 @@ export class PlaylistController extends videojs.EventTarget {
...
-    // Edge/IE bug: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14600375/
-    // Chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=651904
-    this.mainSegmentLoader_.resetEverything(() => {
-      this.tech_.setCurrentTime(this.tech_.currentTime());
-    });
-
-    // don't need to reset audio as it is reset when media changes
+    // Reset main segment loader properties and next segment position information
+    this.mainSegmentLoader_.resetLoaderProperties();
+    this.mainSegmentLoader_.resetLoader();
```
dzianis-dashkevich, line 938: I see that `resetLoaderProperties` resets `initSegment` things + `ended`, should we make it a part of the `resetLoader`?

adrums86: That was my original inclination, however `resetLoader` is called in a couple other spots so I decided against it for now. Definitely open to moving it in there if it makes sense, the risk seems rather low given the other uses.

dzianis-dashkevich: Make sense, 
maybe we can leave a comment about this?

dzianis-dashkevich (approved):

src/playlist-controller.js
adrums86: Good idea. Also, looking at this again I think `// don't need to reset audio as it is reset when media changes` is still useful. Going to add that back in.

wseymour15 (approved):

adrums86:
I found a bug where, if there is > 1 segment buffered, the player will switch back to that segment (at reduced quality) when switching. Have a fix almost complete that I will push ASAP and tag all reviewers for another look.

adrums86:
@dzianis-dashkevich @wseymour15 This could use another look after the latest commit [cd180ff](https://github.com/videojs/http-streaming/pull/1414/commits/cd180ff25807e602596cf616e3488faca17bd9db)

gsimko:
@adrums86 is it a known bug that the new code might fail if nothing is buffered?
Specifically, I get Failed to execute 'end' on 'TimeRanges': The index provided (4294967295) is greater than the maximum bound (0) due to the line `const bufferedEnd = buffered.end(buffered.length - 1);`.

I guess this happens because nothing is buffered. Could we early return from that function if that's the case?

PR #1419, opened 2023-08-16
feat: request Content Steering manifest

adrums86:
## Description
This is the first step toward full content steering support in VHS. This pull request will add the `ContentSteeringController` class which will allow VHS to parse content steering tags from both the m3u8 and mpd parsers, request and parse a content steering manifest, and listen to events from the content-steering-controller that will notify VHS of changes in content steering data.

## Specific Changes proposed
Adding a new `ContentSteeringController` class that will likely live in the playback-controller, parsing content steering tags as they appear in the main manifest. Also requesting the steering manifest when required, and reacting to changes in the steering manifest and playback state.

## Requirements Checklist
  - [ ] Docs/guides updated


src/content-steering-controller.js
```diff
@@ -0,0 +1,317 @@
...
+    }
+  }
+
+  /**
+   * Requests the content steering manifest and parse the response.
+   */
+  requestSteeringManifest() {
+    // add parameters to the steering uri
+    const reloadUri = this.steeringManifest.reloadUri;
+    const uri = this.proxyServerUrl_ ? this.setProxyServerUrl_(reloadUri) : this.setSteeringParams_(reloadUri);
```
wseymour15, line 134: This could be another TODO as I don't believe it is make or break, but are we accounting for params that are tied to the original MPD url for DASH? See Section 8.1 in https://dashif.org/docs/DASH-IF-CTS-00XX-Content-Steering-Community-Review.pdf

test/content-steering-controller.test.js
```diff
@@ -0,0 +1,340 @@
...
+
+  this.mockSegmentLoader.throughput.rate = 99;
+  this.assignAndRequest(steeringTag);
+  assert.equal(this.requests[0].uri, expectedProxyUrl, 'returns expected proxy server URL');
+});
+
+// Common steering manifest tests
+QUnit.test('Can handle content steering manifest with VERSION', function(assert) {
+  const steeringTag = {
+    serverUri: '/content/steering'
```
wseymour15, line 190: SHould this be serverURL?

src/content-steering-controller.js
adrums86: So yeah I actually coded up support for this then deleted it after I realized it's tied to a specific MPD tag. 😅 
`<up:ExtUrlQueryInfo queryTemplate="token=$query:steeringToken$" useMPDUrlQuery="true"
includeInRequests="steering"/> </up:ExtUrlQueryInfo>`
I agree though, lets add at least a comment citing lack of support.

wseymour15 (approved):

Issue #1049, opened 2021-01-21
Issue when seeking across PTS rollovers.


iamstickfigure:
## Description
Seeking in this sample HLS playlist does not work. There is a PTS rollover happening in the file, and that probably is contributing to the issue.

## Sources
I went ahead and setup an hls playlist on S3 from the problematic sample file.
https://snapstream-dev-test-public.s3.us-east-1.amazonaws.com/videojs-pts-rollover-issue/playlist.m3u8

## Steps to reproduce
Load the sample file in video.js
Try to seek far into the file

Here's a test player that demonstrates the issue.
https://videojs-http-streaming.netlify.app/?debug=false&autoplay=false&muted=false&minified=false&liveui=false&partial=false&url=https%3A%2F%2Fsnapstream-dev-test-public.s3.us-east-1.amazonaws.com%2Fvideojs-pts-rollover-issue%2Fplaylist.m3u8&type=&keysystems=&buffer-water=false&override-native=true

### Actual
The seek fails, and it starts playing from the beginning.

### Expected
Should seek properly.

Here is an older test player where seeking works properly. You can just load in the s3 link on this page.
https://videojs.github.io/http-streaming/

### videojs version
video.js v7.11.2

### Workaround
In case it helps, we currently have a workaround hack for this issue. In TimestampRolloverStream, we've commented out the `discontinuity()` in the `reset` function.
```js
this.reset = function () {
  // this.discontinuity();
  this.trigger('reset');
};
```

gkatsev:
@video-archivist-bot save KEPKAp

asoulliereHT:
via gesinger,
https://hockeytech-lh.akamaihd.net/i/game939202audio1_1@742830/master.m3u8
from debug logs:
["VIDEOJS:", "DEBUG:", "VHS:", "PlaybackWatcher >", "Trying to seek outside of seekable at time 0 with …84144.29168888889. Seeking to -84144.29168888889."]

it looks like a rollover happens in the stream around segment https://hockeytech-lh.akamaihd.net/i/game939202audio1_1@742830/segment161013552_1_av-p.ts?sd=10&rebase=on
I think the issue is that we should be able to play fine when going sequentially through segments, but if we seek, we don’t use the reference for the last time when determining if a rollover happens or not. This is something we should fix, somewhere around https://github.com/videojs/http-streaming/blob/main/src/segment-loader.js#L2282-L2300
[...]

PR #1423, opened 2023-08-31
Fix ddos created by requesting infinitely for the same chunks

302Gerben:
## Description
As stated in the following discussion, videojs player requests in certain situations indefinitely for the same chunks which results in a ddos being fired at the server. This issue keeps getting ignored/closed automatically.

https://github.com/videojs/http-streaming/issues/1000#issuecomment-899934566

This situation happens when there is a network congestion or just really slow internet speeds. (Chrome dev tools, throttling slow 3G for example) the player starts buffering after a few seconds. When the network recovers to normal speed, the player keeps requesting the same chunk over and over. (Imagine you have a hundreds of visitors..)

## Specific Changes proposed
Check if there are seekable points.

## Requirements Checklist
- [ ] Bug fixed

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -319,7 +319,7 @@ export default class PlaybackWatcher {
       seekTo = seekableEnd;
     }
 
-    if (this.beforeSeekableWindow_(seekable, currentTime)) {
+    if (seekable.length && seekable.length > 1 && this.beforeSeekableWindow_(seekable, currentTime)) {
       const seekableStart = seekable.start(0);
 
       // sync to the beginning of the live window


adrums86:
Thanks for submitting this PR and linking the issue! I read through everything and think I understand what the issue is, this fix may be preventing a problematic seek, but I'm not sure it's doing it in the way that's expected in the original issue. The seekable range is a TimeRanges object that will return a `start` and `end`, not correlating directly to the number of segments that are currently buffered. If the seekable ranges length is > 1 it essentially means there are 2 ranges that are seekable, for example if there's a buffer gap. 
That said, I do think this needs to be addressed ASAP. I'm wondering if we can just remove the entire block of code that causes this issue? Per the `TODO` at the top of the function the scenario where a seek is before the seekable range or into a gap during live playback seems to be handled with the buffered checks below.

amtins:
_I may be totally off topic, as different comments in different issues seem to mention different problems._ 

@adrums86 I went through the different issues and it reminded me of https://github.com/videojs/mux.js/pull/430  because of https://github.com/videojs/http-streaming/issues/1049#issuecomment-780649483.

I tried to make a fix at the mux level that is based on the same logic but in m2ts ElementaryStream https://github.com/amtins/mux.js/pull/1.

It seems to fix the issue (but which one? 😅), the stream used to test is https://snapstream-dev-test-public.s3.us-east-1.amazonaws.com/videojs-pts-rollover-issue/playlist.m3u8 

**How to test**

- Open the page and start playback
-  Seek near the end (around 00:50:00)

**Before**
https://videojs-http-streaming.netlify.app/?debug=false&autoplay=false&muted=false&fluid=false&minified=false&sync-workers=false&liveui=false&llhls=true&url=https%3A%2F%2Fsnapstream-dev-test-public.s3.us-east-1.amazonaws.com%2Fvideojs-pts-rollover-issue%2Fplaylist.m3u8&type=application%2Fx-mpegURL&keysystems=&buffer-water=false&exact-manifest-timings=false&pixel-diff-selector=false&network-info=false&dts-offset=false&override-native=true&preload=auto&mirror-source=true&forced-subtitles=false

**After**
https://amtins.github.io/http-streaming-mux/

However, if we take the [stream](https://hs-video-testing.s3.amazonaws.com/master_abc_60.m3u8), which is unavailable, the player continually tries each playlist. This seems to be expected behavior, as the player waits for a child playlist to become available again.

adrums86:
Great catch @amtins ! Lets get that PR merged this week and update VHS with a new mux and see if this alleviates the issue for everyone.

amtins:
@adrums86 I'll try to do a proper PR tomorrow, but maybe I could use a little help with the unit tests. However, it seems odd that I should have done this, as the various `piplines` call `TimestampRolloverStream`

adrums86:
This might be fixed in #1433 can anyone confirm?

GeppieNL:
> This might be fixed in #1433 can anyone confirm?

We can confirm it is fixed by 1433! :tada:

adrums86:
@GeppieNL Thanks for the confirmation. Closing this as it appears this is fixed in #1433.

PR #1427, opened 2023-09-13
feat: content steering switching

wseymour15:
## Description
The changes include the logic for switching pathways regarding content steering. All specifications for both DASH and HLS can be found in the resources below. The function of these changes are to ensure that we are communicating with the content steering server, handling the responses, switching playlists based on this data, and ensuring that we follow the requirements of the spec when deciding when/where to switch.

## Specific Changes proposed
- Logic to determine which pathway/serviceLocation should be selected based on priority.
- Logic to switch media in cases where we want to switch playlists.
- Logic to exclude playlists when necessary according to the spec.
- Logic to set timers to retry certain content steering requests.
- Tests to cover specification

## Example Manifests to Test Locally
DASH: https://fastly.content-steering.com/bbb/playlist_steering_fastly_https_cdn-a_cdn-c_cdn-b.mpd
HLS: https://fastly.content-steering.com/bbb_hls/master_steering_fastly_https.m3u8

## Specs
https://dashif.org/docs/DASH-IF-CTS-00XX-Content-Steering-Community-Review.pdf
https://developer.apple.com/streaming/HLSContentSteeringSpecification.pdf

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/content-steering-controller.js
+++ b/src/content-steering-controller.js
@@ -77,7 +77,6 @@ export default class ContentSteeringController extends videojs.EventTarget {
     this.defaultPathway = null;
     this.queryBeforeStart = null;
     this.availablePathways_ = new Set();
-    // TODO: Implement exclusion.
     this.excludedPathways_ = new Set();
     this.steeringManifest = new SteeringManifest();
     this.proxyServerUrl_ = null;
@@ -187,7 +186,7 @@ export default class ContentSteeringController extends videojs.EventTarget {
 
           this.logger_(`manifest request 429 ${error}.`);
           this.logger_(`content steering will retry in ${retrySeconds} seconds.`);
-          this.startTTLTimeout_(retrySeconds);
+          this.startTTLTimeout_(parseInt(retrySeconds, 10));
           return;
         }
         // If the Steering Manifest cannot be loaded and parsed correctly, the
@@ -276,13 +275,13 @@ export default class ContentSteeringController extends videojs.EventTarget {
     this.steeringManifest.priority = steeringJson['PATHWAY-PRIORITY'] || steeringJson['SERVICE-LOCATION-PRIORITY'];
     // TODO: HLS handle PATHWAY-CLONES. See section 7.2 https://datatracker.ietf.org/doc/draft-pantos-hls-rfc8216bis/
 
-    // TODO: fully implement priority logic.
     // 1. apply first pathway from the array.
-    // 2. if first first pathway doesn't exist in manifest, try next pathway.
+    // 2. if first pathway doesn't exist in manifest, try next pathway.
     //    a. if all pathways are exhausted, ignore the steering manifest priority.
     // 3. if segments fail from an established pathway, try all variants/renditions, then exclude the failed pathway.
     //    a. exclude a pathway for a minimum of the last TTL duration. Meaning, from the next steering response,
     //       the excluded pathway will be ignored.
+    //       See excludePathway usage in excludePlaylist().
 
     // If there are no available pathways, we need to stop content steering.
     if (!this.availablePathways_.size) {
@@ -355,7 +354,7 @@ export default class ContentSteeringController extends videojs.EventTarget {
   /**
    * Start the timeout for re-requesting the steering manifest at the TTL interval.
    *
-   * @param {string} ttl time in seconds of the timeout. Defaults to the
+   * @param {number} ttl time in seconds of the timeout. Defaults to the
    *        ttl interval in the steering manifest
    */
   startTTLTimeout_(ttl = this.steeringManifest.ttl) {
--- a/src/content-steering-controller.js
+++ b/src/content-steering-controller.js
@@ -66,11 +66,11 @@ class SteeringManifest {
  * https://datatracker.ietf.org/doc/draft-pantos-hls-rfc8216bis/ section 4.4.6.6.
  * DASH: https://dashif.org/docs/DASH-IF-CTS-00XX-Content-Steering-Community-Review.pdf
  *
- * @param {Object} segmentLoader a reference to the mainSegmentLoader
+ * @param {function} xhr for making a network request from the browser.
+ * @param {function} bandwidth for fetching the current bandwidth from the main segment loader.
  */
 export default class ContentSteeringController extends videojs.EventTarget {
-  // pass a segment loader reference for throughput rate and xhr
-  constructor(segmentLoader) {
+  constructor(xhr, bandwidth) {
     super();
 
     this.currentPathway = null;
@@ -84,8 +84,9 @@ export default class ContentSteeringController extends videojs.EventTarget {
     this.ttlTimeout_ = null;
     this.request_ = null;
     this.excludedSteeringManifestURLs = new Set();
-    this.mainSegmentLoader_ = segmentLoader;
     this.logger_ = logger('Content Steering');
+    this.xhr_ = xhr;
+    this.getBandwidth_ = bandwidth;
   }
 
   /**
@@ -159,7 +160,7 @@ export default class ContentSteeringController extends videojs.EventTarget {
       return;
     }
 
-    this.request_ = this.mainSegmentLoader_.vhs_.xhr({
+    this.request_ = this.xhr_({
       uri
     }, (error, errorInfo) => {
       if (error) {
@@ -236,6 +237,7 @@ export default class ContentSteeringController extends videojs.EventTarget {
   setSteeringParams_(url) {
     const urlObject = new window.URL(url);
     const path = this.getPathway();
+    const networkThroughput = this.getBandwidth_();
 
     if (path) {
       const pathwayKey = `_${this.manifestType_}_pathway`;
@@ -243,10 +245,10 @@ export default class ContentSteeringController extends videojs.EventTarget {
       urlObject.searchParams.set(pathwayKey, path);
     }
 
-    if (this.mainSegmentLoader_.throughput.rate) {
+    if (networkThroughput) {
       const throughputKey = `_${this.manifestType_}_throughput`;
 
-      urlObject.searchParams.set(throughputKey, this.mainSegmentLoader_.bandwidth);
+      urlObject.searchParams.set(throughputKey, networkThroughput);
     }
     return urlObject.toString();
   }
--- a/src/content-steering-controller.js
+++ b/src/content-steering-controller.js
@@ -126,7 +126,7 @@ export default class ContentSteeringController extends videojs.EventTarget {
     }
 
     if (this.queryBeforeStart) {
-      this.requestSteeringManifest(this.queryBeforeStart);
+      this.requestSteeringManifest(this.steeringManifest.reloadUri);
     }
   }
 
@@ -134,15 +134,15 @@ export default class ContentSteeringController extends videojs.EventTarget {
    * Requests the content steering manifest and parse the response. This should only be called after
    * assignTagProperties was called with a content steering tag.
    *
-   * @param {boolean} withQueryBeforeStart If true, the request should be made with the initial serverUri.
+   * @param {string} initialUri The optional uri to make the request with.
+   *    If set, the request should be made with exactly what is passed in this variable.
    *    This scenario is specific to DASH when the queryBeforeStart parameter is true.
    *    This scenario should only happen once on initalization.
    */
-  requestSteeringManifest(withQueryBeforeStart = false) {
-    // add parameters to the steering uri
+  requestSteeringManifest(initialUri) {
     const reloadUri = this.steeringManifest.reloadUri;
 
-    if (!reloadUri) {
+    if (!initialUri && !reloadUri) {
       return;
     }
 
@@ -150,7 +150,7 @@ export default class ContentSteeringController extends videojs.EventTarget {
     // ExtUrlQueryInfo tag support. See the DASH content steering spec section 8.1.
 
     // This request URI accounts for manifest URIs that have been excluded.
-    const uri = withQueryBeforeStart ? reloadUri : this.getRequestURI(reloadUri);
+    const uri = initialUri || this.getRequestURI(reloadUri);
 
     // If there are no valid manifest URIs, we should stop content steering.
     if (!uri) {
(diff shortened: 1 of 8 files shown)


src/media-groups.js
```diff
@@ -250,13 +250,10 @@ export const onError = {
    */
   AUDIO: (type, settings) => () => {
     const {
-      segmentLoaders: { [type]: segmentLoader},
       mediaTypes: { [type]: mediaType },
       excludePlaylist
     } = settings;
 
-    stopLoaders(segmentLoader, mediaType);
```
dzianis-dashkevich, line 258: trying to understand why this has been removed?

wseymour15: I believe we removed this here because this was a redundant call of `stopLoaders`. Apparently, stop loaders will always be called [here](https://github.com/videojs/http-streaming/blob/04451d4d44d2c9ac803b91898497225f2a56ea99/src/media-groups.js#L170) in this case, and then that function will be returned before calling `startLoaders` again..

dzianis-dashkevich: Does it mean that it should be removed for all other media-group types?

wseymour15: I think you're right! I just removed the same piece of code from `SUBTITLES` here: https://github.com/videojs/http-streaming/pull/1427/commits/65f1016df5036783a457dda4ef068da7d0fa0bf8

Before resolving, we'll make sure that @adrums86 takes a look at this change and agrees.

src/content-steering-controller.js
```diff
@@ -109,7 +110,8 @@ export default class ContentSteeringController extends videojs.EventTarget {
       this.decodeDataUriManifest_(steeringUri.substring(steeringUri.indexOf(',') + 1));
       return;
     }
-    this.steeringManifest.reloadUri = resolveUrl(baseUrl, steeringUri);
+    // With DASH queryBeforeStart, we want to use the steeringUri as soon as possible for the request.
+    this.steeringManifest.reloadUri = this.queryBeforeStart ? steeringUri : resolveUrl(baseUrl, steeringUri);
     // pathwayId is HLS defaultServiceLocation is DASH
     this.defaultPathway = steeringTag.pathwayId || steeringTag.defaultServiceLocation;
     // currently only DASH supports the following properties on <ContentSteering> tags.
```
dzianis-dashkevich, line 117: Curious, do we need this `if (this.manifestType_ === 'DASH')` can we just read tag values and set defaults if values are not presented?

src/content-steering-controller.js
```diff
@@ -120,46 +122,85 @@ export default class ContentSteeringController extends videojs.EventTarget {
...
   /**
    * Requests the content steering manifest and parse the response. This should only be called after
    * assignTagProperties was called with a content steering tag.
+   *
+   * @param {boolean} withQueryBeforeStart If true, the request should be made with the initial serverUri.
+   *    This scenario is specific to DASH when the queryBeforeStart parameter is true.
+   *    This scenario should only happen once on initalization.
    */
-  requestSteeringManifest() {
+  requestSteeringManifest(withQueryBeforeStart = false) {
```
dzianis-dashkevich, line 143: Do we need this argument? I see the only usage with `this.queryBeforeStart`, but since this info is available in `this`, we can use it directly in the method

dzianis-dashkevich:
Curious about the usage of `throughput` from segmentLoader:

Are we sure we want to use this value?

`throughput` is the time since all segment-related downloads are done till the time the segment is appended to the buffer (so basically, it is all about decrypt, transmux, and browser append time)

It has nothing to do with actual network time.

src/playlist-controller.js
```diff
@@ -2052,4 +2099,128 @@ export class PlaylistController extends videojs.EventTarget {
...
+      for (const playlist of main.playlists) {
+        this.contentSteeringController_.addAvailablePathway(this.pathwayAttribute_(playlist));
+      }
+
+      this.contentSteeringController_.assignTagProperties(main.uri, main.contentSteering);
+    };
+
+    updateSteeringValues(initialMain);
+
+    this.contentSteeringController_.on('content-steering', this.excludeThenChangePathway_.bind(this));
```
dzianis-dashkevich, line 2126: Do we clear this event listener?

src/playlist-controller.js
```diff
@@ -2052,4 +2099,128 @@ export class PlaylistController extends videojs.EventTarget {
...
+    // We need to ensure we update the content steering values when a new
+    // manifest is loaded in live DASH with content steering.
+    if (this.sourceType_ === 'dash') {
+      this.mainPlaylistLoader_.on('mediaupdatetimeout', () => {
+        this.mainPlaylistLoader_.refreshMedia_(this.mainPlaylistLoader_.media().id);
+
+        // clear past values
+        this.contentSteeringController_.abort();
+        this.contentSteeringController_.clearTTLTimeout_();
+        this.contentSteeringController_.availablePathways_.clear();
```
dzianis-dashkevich, line 2137: we should provide a public interface to work with private stuff

src/content-steering-controller.js
wseymour15: The reason we use this param instead of just using `queryBeforeStart` is because we only want to use that specific reload uri when we call `requestSteeringManifest` during the initial parsing of the manifest file. We will call `requestSteeringManifest` many times later while `queryBeforeStart` is true, but we only want it to use the unparameterized reload uri before the first request to the content steering server.

dzianis-dashkevich: Given the following code:

```js
    if (this.queryBeforeStart) {
      this.requestSteeringManifest(this.queryBeforeStart);
    }
```    

And the fact that `this.queryBeforeStart` is a boolean, I believe it is more semantically right to pass URI instead of this boolean again:

```js
 if (this.queryBeforeStart) {
     const reloadUri = this.getRequestURI(this.steeringManifest.reloadUri)
      this.requestSteeringManifest(reloadUri);
    }
``` 

And for the method declaration part:
`requestSteeringManifest(reloadUri = this.steeringManifest.reloadUri)`

src/playlist-controller.js
wseymour15: I added an `off` call in the dispose function here: https://github.com/videojs/http-streaming/pull/1427/commits/e882de56db6a14be2e74b41ccd16ceba0e67cccf#diff-a167ea6828d7c01e63609400b5cb67b841a64432c629f7913fc04bc528fa3f3cR389

src/playlist-controller.js
wseymour15: Added this public function here, let me know if that is what you were thinking: https://github.com/videojs/http-streaming/pull/1427/commits/e882de56db6a14be2e74b41ccd16ceba0e67cccf#diff-a167ea6828d7c01e63609400b5cb67b841a64432c629f7913fc04bc528fa3f3cR420

src/content-steering-controller.js
```diff
@@ -109,57 +110,95 @@ export default class ContentSteeringController extends videojs.EventTarget {
...
-      // it MUST NOT issue another request for that URI for the remainder of the
-      // playback session. It MAY continue to use the most-recently obtained set
-      // of Pathways.
-      // If the client receives HTTP 429 Too Many Requests with a Retry-After
-      // header in response to a manifest request, it SHOULD wait until the time
-      // specified by the Retry-After header to reissue the request.
+    }, (error, errorInfo) => {
       if (error) {
-        // TODO: HLS RETRY CASE:
+        this.requestError_ = error;
```
dzianis-dashkevich, line 167: Do we use it somewhere?

src/playlist-controller.js
dzianis-dashkevich: Just want to clarify: Does it mean that when `dispose` is called - "contentSteeringController" is no longer usable and we don't need to re-init this listener?

src/content-steering-controller.js
```diff
@@ -109,57 +110,95 @@ export default class ContentSteeringController extends videojs.EventTarget {
...
-    const uri = this.proxyServerUrl_ ? this.setProxyServerUrl_(reloadUri) : this.setSteeringParams_(reloadUri);
+
+    // This request URI accounts for manifest URIs that have been excluded.
+    const uri = initialUri || this.getRequestURI(reloadUri);
+
+    // If there are no valid manifest URIs, we should stop content steering.
+    if (!uri) {
+      this.logger_('No valid content steering manifest URIs. Stopping content steering.');
+      this.dispose();
+      this.trigger('error');
```
dzianis-dashkevich, line 159: We should probably call `this.trigger('error')` before `this.dispose()`.

We should also probably clear all `error` listeners during dispose.

src/content-steering-controller.js
```diff
@@ -234,44 +273,91 @@ export default class ContentSteeringController extends videojs.EventTarget {
...
+
+    const chooseNextPathway = (pathwaysByPriority) => {
+      for (const path of pathwaysByPriority) {
         if (this.availablePathways_.has(path)) {
           return path;
         }
       }
+
+      // If no pathway matches, ignore the manifest and choose the first available.
+      return [...this.availablePathways_][0];
```
dzianis-dashkevich, line 299: Should we handle here the case when `this.availablePathways_` is empty?

src/content-steering-controller.js
```diff
@@ -77,13 +77,14 @@ export default class ContentSteeringController extends videojs.EventTarget {
...
     this.availablePathways_ = new Set();
-    // TODO: Implement exclusion.
     this.excludedPathways_ = new Set();
     this.steeringManifest = new SteeringManifest();
     this.proxyServerUrl_ = null;
     this.manifestType_ = null;
     this.ttlTimeout_ = null;
     this.request_ = null;
+    this.requestError_ = null;
+    this.excludedSteeringManifestURLs = [];
```
dzianis-dashkevich, line 87: Based on my understanding of the code, it is better to use `Set` instead of `Array`

dzianis-dashkevich: `Set` uses hashes which makes the `includes` operation constant instead of O(n).

src/dash-playlist-loader.js
```diff
@@ -325,7 +325,10 @@ export default class DashPlaylistLoader extends EventTarget {
 
     // live playlist staleness timeout
     this.on('mediaupdatetimeout', () => {
-      this.refreshMedia_(this.media().id);
+      // We handle live content steering in the playlist controller
+      if (!this.media().attributes.serviceLocation) {
+        this.refreshMedia_(this.media().id);
```
dzianis-dashkevich, line 330: Should we check that `this.media()` exists?

wseymour15: I believe `this.media()` should always return something valid here, since in all of the places `mediaupdatetimeout` is triggered, the media is set. It definitely wouldn't hurt to have an extra check, but It looks like there was not a check previously either.

src/content-steering-controller.js
wseymour15: I believe we do this a few lines above with the check `if (!this.availablePathways_.size)`

src/content-steering-controller.js
wseymour15: This was meant so that if someone wanted to see the most recent error at any time, they could use this property on the player. However, it may make more sense to rename it `this.requestError` or create a getter, what do you think?

dzianis-dashkevich: Less code is better. I believe we should keep it only in case we already have a place to use it..

wseymour15: Fair enough, as we are logging that error anyways. Removed that in the most recent commit.

adrums86 (commented):
This looks great, Walter! Per @dzianis-dashkevich's comment, I think he is correct. We will probably want to use bandwidth instead of our internal throughput value.

wseymour15:
> Do we need `segmentLoader` as a dependency? As far as I see, we are using `throughput ` and `xhr`; both entities can be provided separately in the constructor.

@dzianis-dashkevich I believe we need to pass the entire segment loader here because `throughput` will be changing. If we just pass `throughput`, it will be a static value. Would you agree with that @adrums86?

adrums86:
@wseymour15 Thanks for addressing this, yeah that is correct. We need a reference to the actual `bandwidth` property on the segmentLoader as it's updated during segment requests. Since the segmentLoader also has an xhr reference we can get both by just passing in the segmentLoader.

dzianis-dashkevich:
> @dzianis-dashkevich I believe we need to pass the entire segment loader here because throughput will be changing. If we just pass throughput, it will be a static value. Would you agree with that @adrums86?

@wseymour15, @adrums86 

Yes, I understand that :)

Sorry, I didn't express my thoughts fully in my previous messages...
I was referring to a famous OOP quote: 

> You wanted a banana, but you got a gorilla holding the banana

My point was to provide a bare minimum interface required for a controller to work.

This is a common approach when designing a new entity, you are using a DI for injecting dependencies, but instead of relying on concrete implementation, you invert control by expecting some interface:
Check the Dependency Inversion Principle from SOLID: https://en.wikipedia.org/wiki/Dependency_inversion_principle

So, in this particular use case: 
Instead of providing the whole `segmentLoader`, we can provide:  `{ get bandwidth() { return segmentLoader.bandwith } }`
 
 or something like that.
 
 It is a minor thing, and I am ok to keep things as is.

adrums86:
@dzianis-dashkevich Yeah I see what you mean, and I 100% agree. The only reason it was done this way was because we needed `xhr` as well, sacrificing a bit of principle for less code. I'm cool with adding a getter to the segment loader and passing just that and `xhr` separately.

dzianis-dashkevich:
+1 from me when all checks will be successful

adrums86:
@dzianis-dashkevich we decided to remove the segmentLoader parameter and just pass xhr and a bandwidth function, I think it's much better. Thanks for the suggestion.

src/media-groups.js
dzianis-dashkevich: Can we add TODO, so we will not forget about it?

dzianis-dashkevich (approved):

src/playlist-controller.js
wseymour15: Correct. Dispose assumes we no longer want to use content steering functionality or that instance of a `contentSteeringController`.

src/media-groups.js
wseymour15: So I removed `stopLoaders` from the other media group types in the `onError` function. It is still used in `onTrackChanged` and `onGroupChanged` where I believe it is still needed.

adrums86 (approved):

PR #1428, opened 2023-09-26
fix: wrap onwarn values in an object as expected by the playlist-loader

ianjaku:
## Description
Hey, apologies for not first opening an issue, but it's a very small fix.
When debugging an issue in one of my projects I noticed the debug log: `VIDEOJS: DEBUG: VHS: PlaylistLoader > m3u8-parser warn for [URL]: undefined`.

This is caused by PlaylistLoader#parseManifest_ expecting onwarn to be called with an object that has a message parameter. In manifest.js#parseManifest the onwarn method is sometimes called with a string. It's also called with a message object inside the m3u8 parser.

## Specific Changes proposed
We could either change onwarn to always return a string, or return a message object from inside manifest.js#parseManifest.
In this fix I went with the latter as to not break any usage of parseManifest.

Please let me know if anything is unclear, I'll be happy to append the PR.

## Requirements Checklist
- [ ] Bug fixed

### DIFF
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -93,7 +93,7 @@ export const parseManifest = ({
     }
 
     if (onwarn) {
-      onwarn(`manifest has no targetDuration defaulting to ${targetDuration}`);
+      onwarn({ message: `manifest has no targetDuration defaulting to ${targetDuration}` });
     }
     manifest.targetDuration = targetDuration;
   }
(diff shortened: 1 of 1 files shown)


mister-ben (approved):
Thanks!

adrums86 (approved):
:shipit:

ianjaku:
Is there a way to get this merged without writing additional tests?

The change is extremely simple, and there is no regression as this was untested behaviour before the change.

PR #1430, opened 2023-10-03
fix: content steering bug fixes and tests

adrums86:
## Description
This PR addresses a few content steering bugs and ads some additional testing around live DASH.

## Specific Changes proposed
Refactoring the content steering initialization to attach listeners separately, change some `queryBeforeStart` behavior and an issue with the initial `ttl` value.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/content-steering-controller.js
+++ b/src/content-steering-controller.js
@@ -430,9 +430,9 @@ export default class ContentSteeringController extends videojs.EventTarget {
    */
   didDASHTagChange(baseURL, newTag) {
     return !newTag && this.steeringManifest.reloadUri ||
-      resolveUrl(baseURL, newTag.serverURL) !== this.steeringManifest.reloadUri ||
+      newTag && (resolveUrl(baseURL, newTag.serverURL) !== this.steeringManifest.reloadUri ||
       newTag.defaultServiceLocation !== this.defaultPathway ||
       newTag.queryBeforeStart !== this.queryBeforeStart ||
-      newTag.proxyServerURL !== this.proxyServerUrl_;
+      newTag.proxyServerURL !== this.proxyServerUrl_);
   }
 }
--- a/src/content-steering-controller.js
+++ b/src/content-steering-controller.js
@@ -133,10 +133,9 @@ export default class ContentSteeringController extends videojs.EventTarget {
    *
    * @param {string} initialUri The optional uri to make the request with.
    *    If set, the request should be made with exactly what is passed in this variable.
-   *    This scenario is specific to DASH when the queryBeforeStart parameter is true.
    *    This scenario should only happen once on initalization.
    */
-  requestSteeringManifest() {
+  requestSteeringManifest(initial) {
     const reloadUri = this.steeringManifest.reloadUri;
 
     if (!reloadUri) {
@@ -147,7 +146,7 @@ export default class ContentSteeringController extends videojs.EventTarget {
     // ExtUrlQueryInfo tag support. See the DASH content steering spec section 8.1.
 
     // This request URI accounts for manifest URIs that have been excluded.
-    const uri = this.getRequestURI(reloadUri);
+    const uri = initial ? reloadUri : this.getRequestURI(reloadUri);
 
     // If there are no valid manifest URIs, we should stop content steering.
     if (!uri) {
@@ -381,6 +380,8 @@ export default class ContentSteeringController extends videojs.EventTarget {
    * aborts steering requests clears the ttl timeout and resets all properties.
    */
   dispose() {
+    this.off('content-steering');
+    this.off('error');
     this.abort();
     this.clearTTLTimeout_();
     this.currentPathway = null;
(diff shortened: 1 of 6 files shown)


src/playlist-controller.js
```diff
@@ -2104,51 +2106,62 @@ export class PlaylistController extends videojs.EventTarget {
...
+      this.mainPlaylistLoader_.on('loadedplaylist', () => {
+        const main = this.main();
+        const didDashTagChange = this.contentSteeringController_.didDASHTagChange(main.uri, main.contentSteering);
 
-    // Do this at startup only, after that the steering requests are managed by the Content Steering class.
-    // DASH queryBeforeStart scenarios will be handled by the Content Steering class.
-    if (!this.contentSteeringController_.queryBeforeStart) {
-      this.tech_.one('canplay', () => {
-        this.contentSteeringController_.requestSteeringManifest();
+        if (didDashTagChange) {
```
wseymour15, line 2162: This is awesome, a good way to limit how often we have to reset the content steering controller.

src/content-steering-controller.js
```diff
@@ -413,13 +408,31 @@ export default class ContentSteeringController extends videojs.EventTarget {
...
+
+  /**
+   * Checks the refreshed DASH manifest content steering tag for changes.
+   *
+   * @param {string} baseURL new steering tag on DASH manifest refresh
+   * @param {Object} newTag the new tag to check for changes
+   * @return a true or false whether the new tag has different values
+   */
+  didDASHTagChange(baseURL, newTag) {
+    return !newTag && this.steeringManifest.reloadUri ||
```
wseymour15, line 432: Out of curiosity, why are we returning true if there is no new tag?

src/content-steering-controller.js
```diff
@@ -139,18 +136,18 @@ export default class ContentSteeringController extends videojs.EventTarget {
    *    This scenario is specific to DASH when the queryBeforeStart parameter is true.
    *    This scenario should only happen once on initalization.
    */
-  requestSteeringManifest(initialUri) {
+  requestSteeringManifest() {
```
wseymour15, line 139: I think my reasoning for including this `initialUri` logic is due to the spec saying that in the cases of `queryBeforeStart = true`, we are supposed to omit the `_DASH_pathway` and `_DASH_throughput` values on the initial request to the steering server. Is that still handled?

If we do want to remove this, I think we want to make sure we remove that param from the documentation as well.

src/playlist-controller.js
wseymour15, line 2162: Is it a possible scenario that the content steering tag stays the same, but the BaseURLs/playlists change and we need to update?

adrums86: Good call, definitely a valid case. Will address that.

src/content-steering-controller.js
adrums86: I was thinking we just send them if we have them, but I looked at the DASH spec again and you're correct. My mistake, fix incoming!

src/content-steering-controller.js
adrums86: Only if there's no new tag _and_ the `reloadUri` is set. So the case where we go from having any content steering tag (since `reloadUri` is required), to none.

src/playlist-controller.js
```diff
@@ -2104,51 +2106,83 @@ export class PlaylistController extends videojs.EventTarget {
...
+            const serviceLocation = playlist.attributes.serviceLocation;
+
+            if (serviceLocation) {
+              newPathways.push(serviceLocation);
+              if (!availablePathways.has(serviceLocation)) {
+                return true;
+              }
+            }
+            if (!newPathways.length && availablePathways.size) {
+              return true;
```
wseymour15, line 2177: I think I am having a tough time understanding what this if statement is doing.. This will always return false as long as the first playlist has a `serviceLocation` property, right? Do we need to check when `availablePathways` has a value that `newPathways` doesn't, AKA the removal of a pathway (I think).

adrums86: My mistake. This is supposed to be outside of the loop. This would be the case where a manifest refreshes with no `serviceLocation`s, represented by the `newPathways` collection. If the `.mpd` refreshes with no `serviceLocations` however the steering controller has available pathways, we want to return true and signal a change.

wseymour15: I think that makes more sense. I might be overthinking this, but won't `availablePathways.size` almost always return true? I think in this case we would want to return false.

Example scenario: In a case where nothing changes, newPathways length = 0, and `availablePathways` stays the same, but has a length. I think in this case we should be returning false, but is returning true currently.

Let me know if I am off here, as I definitely could be.

adrums86: Since we add a `serviceLocation` to `newPathways` regardless of `availablePathways` having the pathway or not (only if it exists), the example scenario would actually indicate changes and you would want to return `true` I think. If `newPathways.length === 0` then the main manifest/.mpd update has no `serviceLocation`s, so if the content steering controller has `availablePathways` and no `newPathway`s exist in the main manifest after a refresh, we would be changing state and need to disable content steering as the current main manifest has no `serviceLocation`s to steer with.

wseymour15: Thanks for the explanation, that makes more sense!

wseymour15 (approved):
Great work!

PR #1432, opened 2023-10-06
feat: Content Steering HLS Pathway Cloning

wseymour15:
## Description
Implementation for HLS Pathway cloning for content steering. See the spec here: https://datatracker.ietf.org/doc/draft-pantos-hls-rfc8216bis/

Whenever we get a response from the content server with a `PATHWAY-CLONES` array, we need to add, update, or delete playlists and media based on the data we receive. A pathway clone object tells us information about a new pathway that should be created/updated based off of another pathway.

## Specific Changes proposed
- Store the pathway cloning data retrieved from the server reponse in the content steering controlelr.
- Add a new pathway with playlist and media data when the pathway clone did not exist previously.
- Update playlist and media data from a previously existing pathway clone when we get new data from the pathway clone object.
- Delete playlist and media data from a pathway clone that is no longer in the content steering server response.
- Unit tests for this functionality.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/content-steering-controller.js
+++ b/src/content-steering-controller.js
@@ -11,10 +11,12 @@ import videojs from 'video.js';
  * TTL: number in seconds (optional) until the next content steering manifest reload.
  * RELOAD-URI: string (optional) uri to fetch the next content steering manifest.
  * SERVICE-LOCATION-PRIORITY or PATHWAY-PRIORITY a non empty array of unique string values.
+ * PATHWAY-CLONES: array (optional) (HLS only) pathway clone objects to copy from other playlists.
  */
 class SteeringManifest {
   constructor() {
     this.priority_ = [];
+    this.pathwayClones_ = [];
   }
 
   set version(number) {
@@ -43,6 +45,13 @@ class SteeringManifest {
     }
   }
 
+  set pathwayClones(array) {
+    // pathwayClones must be non-empty.
+    if (array && array.length) {
+      this.pathwayClones_ = array;
+    }
+  }
+
   get version() {
     return this.version_;
   }
@@ -57,6 +66,10 @@ class SteeringManifest {
 
   get priority() {
     return this.priority_;
+  }
+
+  get pathwayClones() {
+    return this.pathwayClones_;
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/playlist-loader.js
```diff
@@ -14,7 +14,9 @@ import {
   addPropertiesToMain,
   mainForMedia,
   setupMediaPlaylist,
-  forEachMediaGroup
+  forEachMediaGroup,
```
adrums86, line 17: Is this used anywhere?

src/content-steering-controller.js
```diff
@@ -397,7 +415,6 @@ export default class ContentSteeringController extends videojs.EventTarget {
     this.request_ = null;
     this.excludedSteeringManifestURLs = new Set();
     this.availablePathways_ = new Set();
-    this.excludedPathways_ = new Set();
```
adrums86, line 400: do we want to remove this here?

adrums86 (commented):
Looks great, just a couple of questions.

src/playlist-loader.js
wseymour15: Yeah, just [not in my changes](https://github.com/videojs/http-streaming/blob/5534ec726ab626fafb8ea2b95882a9a9b20a4c17/src/playlist-loader.js#L335) 👍

adrums86: Oh nice, good catch!

src/playlist-controller.js
```diff
@@ -2211,6 +2220,99 @@ export class PlaylistController extends videojs.EventTarget {
...
+
+    currentPathwayClones.forEach((clone) => {
+      const newClone = nextPathwayClones.find(c => c.ID === clone.ID);
+
+      // Delete the old pathway clone.
+      if (!newClone) {
+        this.mainPlaylistLoader_.updateOrDeleteClone(clone);
+        this.contentSteeringController_.excludePathway(clone.ID);
+      }
+    });
```
dzianis-dashkevich, line 2247: I see that `PathwayClone` entity has an ID, I would suggest using Map instead of Array. It will be easier to manipulate

wseymour15: Good call, I updated this to use `Map`s in my most recent commit

src/playlist-controller.js
```diff
@@ -2162,6 +2162,15 @@ export class PlaylistController extends videojs.EventTarget {
     if (!currentPathway) {
       return;
     }
+
+    const pastClones = this.contentSteeringController_.currentPathwayClones;
```
adrums86, line 2166: I'm wondering if it encapsulate things a bit better to move this into the `handlePathwayClones_` function and do something like `if (!hasClones) return` and just call `handlePathwayClones_` regardless?

adrums86 (commented):
This looks great overall, 1 minor style question.

src/playlist-controller.js
wseymour15: good call, moved that to the handlePathwayClones_ function in my most recent commit

adrums86 (approved):

PR #1433, opened 2023-10-11
Clamp seekable end

dzianis-dashkevich:
## Description

We have default `liveEdgePadding` as 3 * targetDuration. 
Assume we have a stream with a maximum of three segments. 
Each segment's actual duration is less than a `targetDuration`. 
In this scenario, we may have a `seekable.end` less than a `seekable.start`. 
This may lead to unexpected behavior.

## Specific Changes proposed
Clamp `seekable.end` to make sure that it is more or equal to `seekable start`.

## Requirements Checklist
  - [ ] Docs/guides updated


misteroneill (approved):
Seems eminently reasonable to me!

alex-barstow (approved):

adrums86 (approved):
Great find! :shipit:

wseymour15 (approved):

PR #1448, opened 2023-11-20
fix: use startTime instead of 0 for finiteDuration

adrums86:
## Description
We shouldn't set endTime to 0, instead we should set it to the startTime if we don't have a duration.

## Specific Changes proposed
replace 0 with startTime

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/util/text-tracks.js
+++ b/src/util/text-tracks.js
@@ -234,7 +234,7 @@ export const addMetadata = ({
   // Map each cue group's endTime to the next group's startTime
   sortedStartTimes.forEach((startTime, idx) => {
     const cueGroup = cuesGroupedByStartTime[startTime];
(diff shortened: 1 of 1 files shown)


adrums86:
@dzianis-dashkevich a fix from a while ago that you brought to my attention re #1411.

dzianis-dashkevich (approved):

robert-schurig-mi:
With "@videojs/http-streaming": "3.0.2" the cues worked as expected and they get active because the endTime was set to Infinity.

Then the endTime of a cue was set to "0" and the stoped working, they don't get active and not event dispatch.

Now in the latest version "@videojs/http-streaming": "3.8.0" the endTime is set to the startTime... okay. sounds for me like an improvement... but the cue doesn't get active too. 
So what does it fix ?

PR #1453, opened 2023-11-29
feat: Add experimental support for ManagedMediaSource

mister-ben:
Adds basic support for ManagedMediaSource. Must be enabled with the `useManagedMediaSource` VHS option.

Does not implement an alternate AirPlay source - this requires a more significant change, to add two source els. This means remote playback has to be disabled on the video el when using MMS.

Event listeners for advanced control are not yet implemented - `startstreaming`, `endstreaming`, `qualitychange`


gkatsev (approved):

mister-ben:
Adding as `<source>` children instead of the `src` attribute could be done [here](https://github.com/videojs/http-streaming/blob/99ab42ad3eb0c605aac9e8499a3616dc1b50eefb/src/videojs-http-streaming.js#L1061). We would need to ensure both the attribute and `<source>` children are removed on source changes, including to or from non-VHS sources. Also, will anything else expect the `src` attribute to be present?

snowinszu:
Hi, any news?

dzianis-dashkevich (approved):

PR #1456, opened 2023-11-29
fix: always use vod sync-point for vod

dzianis-dashkevich:


src/sync-controller.js
```diff
@@ -212,6 +212,13 @@ export default class SyncController extends videojs.EventTarget {
    *          A sync-point object
    */
   getSyncPoint(playlist, duration, currentTimeline, currentTime) {
+    // Always use VOD sync point for VOD
+    if (duration !== Infinity) {
+      const vodSyncPointStrategy = syncPointStrategies.find(({ name }) => name === 'VOD');
```
wseymour15, line 217: I'm no expert on this, but will this sync point always exist? Would it be safer to only return here if we do get this value?

If we are guaranteed to retrieve this value, then this can be ignored :)

wseymour15 (approved):
+1 Once my question is answered!

src/sync-controller.js
dzianis-dashkevich: It always exists, syncPointStrategies is static array

PR #1458, opened 2023-11-30
Sync controller media sequence strategy

dzianis-dashkevich:
- implement a new sync strategy based on the media sequence
- implement exact match for sync controller
- fix fast quality switch race condition
- implement force timestamp offset after resync loader

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -567,6 +567,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.checkBufferTimeout_ = null;
     this.error_ = void 0;
     this.currentTimeline_ = -1;
+    this.shouldForceTimestampOffsetAfterResync_ = false;
     this.pendingSegment_ = null;
     this.xhrOptions_ = null;
     this.pendingSegments_ = [];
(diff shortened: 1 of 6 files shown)


src/segment-loader.js
```diff
@@ -1433,6 +1437,7 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     if (next.isSyncRequest) {
       next.mediaIndex = getSyncSegmentCandidate(this.currentTimeline_, segments, bufferedEnd);
+      this.logger_(`choose next request. Cant find sync point. Fallback to media Index: ${next.mediaIndex}`);
```
wseymour15, line 1440: nit sp: `Can't` or 
```suggestion
      this.logger_(`choose next request. Can not find sync point. Fallback to media Index: ${next.mediaIndex}`);
```

wseymour15 (approved):
These changes make sense to me! Approving under the assumption that we are coming back to this and cleaning up the skipped tests

adrums86 (approved):
LGTM. I think the skips are okay here as these tests seem to be dependent on the old sync strategies throwing away a request. I do however think we should add some new sync-controller tests for the new strategy, either in this PR or a follow up.

dzianis-dashkevich:
@adrums86, I will add it in the next PRs

PR #1460, opened 2023-12-08
feat: enable playlists with 'usable' keystatus

adrums86:
## Description
Only enable playlists with a 'usable' keystatus when playing DASH with DRM.

## Specific Changes proposed
Reference the playlist keyId after getting a `keystatuschange` event from contrib-eme and only enable encrypted playlists that we have a matching usable keyId for.

Note: This PR includes changes from #1461

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -925,21 +925,6 @@ export default class DashPlaylistLoader extends EventTarget {
     }
   }
 
-  // /**
-  //  * Returns the default_KID from a DASH playlist.
-  //  *
-  //  * @param {playlist} playlist to fetch the default_KID from.
-  //  * @return a 32 digit hex string that represents the keyId of the encrypted playlist.
-  //  */
-  // getKID(playlist) {
-  //   if (playlist.contentProtection &&
-  //     playlist.contentProtection.mp4protection &&
-  //     playlist.contentProtection.mp4protection.attributes['cenc:default_KID']) {
-  //     // default KID is a 32 digit hext string separated by '-'.
-  //     return playlist.contentProtection.mp4protection.attributes['cenc:default_KID'].replace(/-/g, '');
-  //   }
-  // }
-
   /**
    * Returns the key ID set from a playlist
    *
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -930,6 +930,10 @@ export default class DashPlaylistLoader extends EventTarget {
 
   excludeNonUsablePlaylists() {
     this.mainPlaylistLoader_.main.playlists.forEach((playlist) => {
+      if (!playlist.contentProtection.mp4protection ||
+        !playlist.contentProtection.mp4protection.attributes['cenc:default_KID']) {
+        return;
+      }
       const playlistKID = playlist.contentProtection.mp4protection.attributes['cenc:default_KID'].replace('-', '');
       const hasUsableKeystatus = this.keyStatusMap.has(playlistKID) && this.keyStatusMap.get(playlistKID) === 'usable';
 
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -314,7 +314,7 @@ export default class DashPlaylistLoader extends EventTarget {
     this.vhs_ = vhs;
     this.withCredentials = withCredentials;
     this.addMetadataToTextTrack = options.addMetadataToTextTrack;
-    this.keyStatusMap = new Map();
+    this.keyStatusMap_ = new Map();
 
     if (!srcUrlOrPlaylist) {
       throw new Error('A non-empty playlist URL or object is required');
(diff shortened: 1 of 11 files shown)


src/videojs-http-streaming.js
```diff
@@ -1127,6 +1127,7 @@ class VhsHandler extends Component {
 
     this.player_.tech_.on('keystatuschange', (e) => {
       if (e.status !== 'output-restricted') {
+        this.playlistController_.updatePlaylistByKeyStatus(e.keyId, e.status);
```
dzianis-dashkevich, line 1130: Do we want to keep all the code listed further for cases when we receive `output-restricted`?

src/dash-playlist-loader.js
```diff
@@ -924,4 +924,41 @@ export default class DashPlaylistLoader extends EventTarget {
...
+  //  * @return a 32 digit hex string that represents the keyId of the encrypted playlist.
+  //  */
+  // getKID(playlist) {
+  //   if (playlist.contentProtection &&
+  //     playlist.contentProtection.mp4protection &&
+  //     playlist.contentProtection.mp4protection.attributes['cenc:default_KID']) {
+  //     // default KID is a 32 digit hext string separated by '-'.
+  //     return playlist.contentProtection.mp4protection.attributes['cenc:default_KID'].replace(/-/g, '');
+  //   }
+  // }
```
dzianis-dashkevich, line 941: commented code

wseymour15 (commented):
Looks great! I left a few nit comments, as well as maybe including an additional test

src/playlist-controller.js
```diff
@@ -2380,4 +2383,72 @@ export class PlaylistController extends videojs.EventTarget {
...
+        const hasUsableKeyStatus = this.keyStatusMap_.has(key) && this.keyStatusMap_.get(key) === USABLE;
+        const nonUsableExclusion = playlist.lastExcludeReason_ === NON_USABLE && playlist.excludeUntil === Infinity;
+
+        if (!hasUsableKeyStatus) {
+          playlist.excludeUntil = Infinity;
+          playlist.lastExcludeReason_ = NON_USABLE;
+        } else if (hasUsableKeyStatus && nonUsableExclusion) {
+          delete playlist.excludeUntil;
+          delete playlist.lastExcludeReason_;
+        }
```
dzianis-dashkevich, line 2417: can we add logs for cases when we exclude it and when we restore it back

src/videojs-http-streaming.js
dzianis-dashkevich: I guess it is not longer valid

src/playlist-controller.js
```diff
@@ -2380,4 +2383,72 @@ export class PlaylistController extends videojs.EventTarget {
...
+        const NON_USABLE = 'non-usable';
+        const hasUsableKeyStatus = this.keyStatusMap_.has(key) && this.keyStatusMap_.get(key) === USABLE;
+        const nonUsableExclusion = playlist.lastExcludeReason_ === NON_USABLE && playlist.excludeUntil === Infinity;
+
+        if (!hasUsableKeyStatus) {
+          playlist.excludeUntil = Infinity;
+          playlist.lastExcludeReason_ = NON_USABLE;
+        } else if (hasUsableKeyStatus && nonUsableExclusion) {
+          delete playlist.excludeUntil;
+          delete playlist.lastExcludeReason_;
```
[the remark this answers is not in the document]
adrums86: Yeah good call. Adding that test.

src/videojs-http-streaming.js
adrums86: Yeah I wasn't sure the best approach here. Leaving the code below, we'll maintain the same behavior as before where we exclude all the HD renditions on a single `output-restricted` status. I was also originally considering having this behind an options flag? Definitely open to some discussion on what we ultimately should have as the default behavior here.

dzianis-dashkevich: as far as I understand it conflicts with the proposed, since when we receive `output-restricted` we will exclude every HD+ rendition.

dzianis-dashkevich: I guess we should just remove it

dzianis-dashkevich: I mean the change is straightforward and should just work, do we need a flag for this?

adrums86: Yeah, I don't really see much downside in keeping this as is without a flag as the playlists with a keyId and no usable keystatus would not be playable anyway.  Also agree, I think the HD rendition filtering is safe to remove here as the same thing would be accomplished through the new logic. Removing the rest below and simplifying this listener.

adrums86:
@wseymour15 and @dzianis-dashkevich Thanks for taking a look at this! I just pushed some updates, another look would be appreciated when you have time.

dzianis-dashkevich (approved):

PR #1463, opened 2023-12-14
fix: check if change to the provided type is supported

rtkac:
## Description
This change will fix an error on change to the provided codec which is not supported.
Just warn about it and do not freeze the video.

### DIFF
--- a/src/source-updater.js
+++ b/src/source-updater.js
@@ -280,8 +280,13 @@ const actions = {
 
     sourceUpdater.logger_(`changing ${type}Buffer codec from ${sourceUpdater.codecs[type]} to ${codec}`);
 
-    sourceBuffer.changeType(mime);
-    sourceUpdater.codecs[type] = codec;
+    // check if change to the provided type is supported
+    try {
+      sourceBuffer.changeType(mime);
+      sourceUpdater.codecs[type] = codec;
+    } catch (e) {
+      videojs.log.warn(`Failed to changeType on ${type}Buffer`, e);
+    }
   }
 };
 


adrums86:
Thank you for submitting this PR! The Edge `changeType` issue has been on our radar and we're also thinking about solutions. We have some ideas, but I think we may need to handle this by re-creating the source buffer instead of calling change type. There's some more information in the [W3C spec](https://www.w3.org/TR/media-source-2/#sourcebuffer-init-segment-received) regarding `changeType` throwing an error in certain cases where `isTypeSupported` is returning `true` for that same codec.

rtkac:
@adrums86 thanks for your comment. 

Do you mean that there will be more complex solution than the fix in the PR? Do you know when we would expect an update and release of the http-streaming library? 
FYI: we are using video.js v7.x which is using http-streaming v2.x. So I believe v2 version will be updated as well.

For more context:
Edge already uses the MediaFoundation renderer, which doesn't support `changeType` [https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer/changeType](https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer/changeType). According to `chrome://flags/#enable-media-foundation-clear` flag, there will be a transition to MediaFoundation for non DRM content. So this is where the problem begins.

Thanks a lot again.

adrums86:
@rtkac Yeah unfortunately I think the solution might be a bit more complex. We will likely have to do some re-working of the logic above `changeType` to create s new source buffer instead of calling `changeType` here for Edge. I can't provide a solid date unfortunately, but I can say that this is _very_ high on our priority list. I am personally looking at this issue, so I can provide updates as soon as their available. Your effort and feedback are super helpful so please feel free to reach out on videojs slack or here on github if you'd like to collaborate further on the fix.

gkatsev:
Of note, are we still maintaining the 2.x branch? Last commit was in February.

However, maybe worth having a change like this in main if it improves things and doesn't cause issues while the proper solution is being worked on?

rtkac:
@adrums86 understand, thanks for your reply.
@gkatsev I hope so :D we would really appreciate it if the changes also affects 2.x branch. It is super important for us right now because video.js 7.x still uses 2.x http-streaming. Thanks

adrums86:
I'm not opposed to using this as a bandaid until we can get the proper fix implemented around `changeType` for Edge. Yeah 2.x is quite stale and not actively maintained 😅 .

dzianis-dashkevich:
> Edge already uses the MediaFoundation renderer, which doesn't support changeType https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer/changeType.

Is there any open documentation regarding this?

I am trying to understand how this fix helps avoid playback failure.

I would assume it should fail if we didn't actually `changeType` on the source buffer and then call `appendBuffer` with data encoded using different codecs.
 
We exclude renditions with unsupported codecs using [MediaSource.isTypeSupported](https://www.w3.org/TR/media-source-2/#istypesupported-method) here https://github.com/videojs/http-streaming/blob/main/src/playlist-controller.js#L1784

In this case, if `changeType` fails for some reason we should probably attempt to re-initialize sourceBuffer with provided codecs. My expectation is that this should work fine since we should have only supported codecs at this moment.

src/source-updater.js
```diff
@@ -280,7 +280,13 @@ const actions = {
 
     sourceUpdater.logger_(`changing ${type}Buffer codec from ${sourceUpdater.codecs[type]} to ${codec}`);
 
-    sourceBuffer.changeType(mime);
+    // check if change to the provided type is supported
+    try {
+      sourceBuffer.changeType(mime);
+    } catch (e) {
+      videojs.log.warn(e);
```
dzianis-dashkevich, line 287:
```suggestion
      videojs.log.warn(`Failed to changeType on ${type}Buffer`, e);
```

src/source-updater.js
```diff
@@ -280,7 +280,13 @@ const actions = {
...
+    }
+
     sourceUpdater.codecs[type] = codec;
```
dzianis-dashkevich, line 290:
```suggestion
    // check if change to the provided type is supported
    try {
      sourceBuffer.changeType(mime);
      sourceUpdater.codecs[type] = codec;
    } catch (e) {
      videojs.log.warn(`Failed to changeType on ${type}Buffer`, e);
    }
```

rtkac: Hi @dzianis-dashkevich

thank you for your reply and suggestion.

Have you tried your changes? Is it working now as expected? I definitely gonna try this to see the behaviour. 

Thanks again!

rtkac:
@adrums86 I understand that the 2.x is not maintained, but this fix would be real helpful for us because we depend on v2 right now. What do you think, is it possible to merge it into v2 and release a new version?
Thanks!

adrums86:
@rtkac It _might_ be possible to get this into 2.x and get a quick release out. I do think we should get the change @dzianis-dashkevich suggested into the main branch while we work on the proper fix however.

rtkac:
@adrums86 @dzianis-dashkevich PR updated

dzianis-dashkevich (approved):

rtkac:
@dzianis-dashkevich thank you 👍

adrums86:
@rtkac I forgot to mention I released 2.x with this fix https://github.com/videojs/http-streaming/releases/tag/v2.16.3

MCab93:
@adrums86 @dzianis-dashkevich I am checking release of branch v2.16.3 https://github.com/videojs/http-streaming/actions/runs/7592761058/job/20682472936 - this build failed. When I use http-streaming v2.16.3 in my project the changes of this PR are not there. Can you please double check and make new release of 2x branch ?

PR #1466, opened 2023-12-15
fix: toLowerCase keyIds from manifest and use fastQualityChange

adrums86:
## Description
Manifests with uppercase keyIds were not matching with keyIds coming from the MediaKeyStatusMap. Also occasionally we would buffer encrypted content that we may not have a key for, we need to call fastQualityChange to clear the buffer before switching playlists.

## Specific Changes proposed
Add `toLowerCase` to the keyId from the manifest when passing the keyIdSet for playlist filtering and call `fastQualityChange` instead of `switchMedia` then fix tests.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/dash-playlist-loader.js
+++ b/src/dash-playlist-loader.js
@@ -940,7 +940,7 @@ export default class DashPlaylistLoader extends EventTarget {
 
         if (defaultKID) {
           // DASH keyIds are separated by dashes.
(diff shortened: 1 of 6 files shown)


alex-barstow (approved):
lgtm

dzianis-dashkevich (approved):

wseymour15 (approved):

PR #1470, opened 2023-12-26
fix: account for difference between duration info in the playlist and the actual duration

dzianis-dashkevich:
## Description
we should account for the difference between the duration info in the manifest and the actual duration of the segment when selecting next segment after quality change.

Assume we have the following playlist:

```shell
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-TARGETDURATION:10
#EXTINF:10.000,
segment-1.ts
#EXTINF:10.000,
segment-2.ts
#EXTINF:10.000,
segment-3.ts
#EXTINF:10.000,
segment-4.ts
#EXT-X-ENDLIST
```

Assume we already appended two segments, so `bufferend.end` should be `20`, but we don't have an exact duration match between the playlist and the actual segment duration, so we receive it as `19.99999..`. 
Currently, we select segment-2.ts instead of segment-3.ts because we do not account for this timestamp fluctuation.

## Specific Changes proposed
Check whether we hit timestamp fluctuation by adding `TIME_FUDGE_FACTOR` (1 / 30).

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -511,8 +511,16 @@ export const getMediaInfoForTime = function({
 
     time -= partAndSegment.duration;
 
-    if (time === 0) {
-      // we are exactly at the end of the current segment
+    if (time === 0 || (time + TIME_FUDGE_FACTOR) >= 0) {
+      // 1) We are exactly at the end of the current segment.
+      // 2) We are extremely close to the end of the current segment (The difference is less than  1 / 30).
+      //    We may encounter this situation when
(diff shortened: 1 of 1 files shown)


adrums86 (approved):
Makes sense to me. :shipit:

PR #1472, opened 2024-01-11
feat: handle rollover for VTT cues

dzianis-dashkevich:
## Description
Handle left/right rollover for VTT cues

### DIFF
--- a/src/vtt-segment-loader.js
+++ b/src/vtt-segment-loader.js
@@ -477,13 +477,17 @@ export default class VTTSegmentLoader extends SegmentLoader {
       return;
     }
 
-    const timestampmap = segmentInfo.timestampmap;
-    const diff = (timestampmap.MPEGTS / ONE_SECOND_IN_TS) - timestampmap.LOCAL + mappingObj.mapping;
+    const { MPEGTS, LOCAL } = segmentInfo.timestampmap;
+    const diff = (MPEGTS / ONE_SECOND_IN_TS) - LOCAL + mappingObj.mapping;
 
     segmentInfo.cues.forEach((cue) => {
-      // First convert cue time to TS time using the timestamp-map provided within the vtt
-      cue.startTime += diff;
-      cue.endTime += diff;
+      const duration = cue.endTime - cue.startTime;
+      const startTime = this.sourceType_ === 'dash' ?
+        cue.startTime + diff :
+        this.handleRollover_(cue.startTime + diff, mappingObj.time);
+
+      cue.startTime = Math.max(startTime, 0);
+      cue.endTime = Math.max(startTime + duration, 0);
     });
 
     if (!playlist.syncInfo) {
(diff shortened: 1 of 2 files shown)


src/vtt-segment-loader.js
```diff
@@ -496,4 +500,30 @@ export default class VTTSegmentLoader extends SegmentLoader {
       };
     }
   }
+
+  handleRollover_(value, reference) {
```
adrums86, line 504: I wonder if it's worth making the rollover mechanic more generic and exposing it from mux, since we have almost the same thing [here](https://github.com/videojs/mux.js/blob/c10bb8cc484fd79340112ed86831a824f6e4c93c/lib/m2ts/timestamp-rollover-stream.js#L22)?

adrums86 (commented):
Looks great, just a question regarding `handleRollover` and the possibility of using a similar function from mux.

src/vtt-segment-loader.js
wseymour15: On top of this, I feel like some documentation on what this function is doing, as well as the reasoning behind it, could be beneficial.

src/vtt-segment-loader.js
```diff
@@ -477,13 +477,17 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
-    const diff = (timestampmap.MPEGTS / ONE_SECOND_IN_TS) - timestampmap.LOCAL + mappingObj.mapping;
+    const { MPEGTS, LOCAL } = segmentInfo.timestampmap;
+    const diff = (MPEGTS / ONE_SECOND_IN_TS) - LOCAL + mappingObj.mapping;
 
     segmentInfo.cues.forEach((cue) => {
-      // First convert cue time to TS time using the timestamp-map provided within the vtt
-      cue.startTime += diff;
-      cue.endTime += diff;
+      const duration = cue.endTime - cue.startTime;
+      const startTime = this.sourceType_ === 'dash' ?
```
wseymour15, line 485: For my own understanding, why doesn't DASH need to handle rollover?

src/vtt-segment-loader.js
dzianis-dashkevich: @adrums86 

Yes, I was considering re-using mux.js, but I realized it should not be a public API from mux.js.

Ideally, it should be moved to video-utils and exported in both VHS and mux.js
Then I realized there would be no other places besides mux.js and VTT loader, and it would not be changed, so I added it here.

Do you think we could benefit from moving it to videojs-utils?

@wseymour15 this is basically 100% the same functionality we have in Mux.js in RolloverStream. But I can def add some details here as well.

src/vtt-segment-loader.js
dzianis-dashkevich: Actually, this is an interesting question... 
MPEGTS has a limitation for timestamps as 2^33. If the timestamp is more than 2^33, it will roll over to 0. It will continue to increase monotonically until it reaches 2^33 and rolls over once again, etc...

Initially, I decided to keep it isolated to HLS, but technically, it is possible to use MPEGTS with a DASH (I haven't seen it, though).
So I would assume that it might be better to check that MPEGTS timestamp map > 0 rather than isolate it to HLS only

src/vtt-segment-loader.js
adrums86: Yeah I mean, I think in an ideal world we would move this out into video-utils. In this specific case due to multiple factors I'm not sure it's worth the refactoring across multiple repositories. I'd say we leave it as is.

dzianis-dashkevich: @wseymour15 

> I feel like some documentation on what this function is doing, as well as the reasoning behind it, could be beneficial.

I added documentation

src/vtt-segment-loader.js
dzianis-dashkevich: I updated this check

wseymour15 (approved):

adrums86 (approved):

PR #1474, opened 2024-01-16
fix: changeType on full codec change only

adrums86:
## Description
Call changeType only when the codec base has changed.

## Specific Changes proposed
Parse the codec base from the new codec and the existing codec for the given track and compare that to determine if we need to call changeType on the source buffer.

## Requirements Checklist
  - [ ] Docs/guides updated


misteroneill (approved):
Seems reasonable to me!

PR #1478, opened 2024-01-19
feat: error type enhancement

wseymour15:
## Description

NOTE: We will need to pull in the newest version of video.js before we can merge this, as there is a dependency on the newly created `videojs.Errors`: https://github.com/videojs/video.js/pull/8564

The goal of this change is to pass error metadata into MediaErrors in the form of the `metadata` property so we can gain more details on errors.

## Specific Changes proposed

- Add error types and other metadata when we create errors, so we can more specificity on what type of error is occurring.

## Requirements Checklist
  - [ ] Docs/guides updated


src/dash-playlist-loader.js
```diff
@@ -431,7 +433,8 @@ export default class DashPlaylistLoader extends EventTarget {
           internal: true,
           playlistExclusionDuration: Infinity,
           // MEDIA_ERR_NETWORK
-          code: 2
+          code: 2,
+          errorType: videojs.Errors.DashManifestSidxContainerError,
```
dzianis-dashkevich, line 437: I would name it as UnsupportedSidxContainer, and probably pass received container as additional metadata.

src/media-segment-request.js
```diff
@@ -163,7 +163,8 @@ const parseInitSegment = (segment, callback) => {
     return callback({
       internal: true,
       message: `Found unsupported ${type || 'unknown'} container for initialization segment at URL: ${uri}`,
-      code: REQUEST_ERRORS.FAILURE
+      code: REQUEST_ERRORS.FAILURE,
+      errorType: videojs.Errors.SegmentUnsupportedContainer
```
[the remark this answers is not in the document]
dzianis-dashkevich: And I would expect the received type to be a part of the error's metadata

src/segment-loader.js
```diff
@@ -2277,7 +2277,8 @@ export default class SegmentLoader extends videojs.EventTarget {
         `video buffer: ${timeRangesToArray(videoBuffered).join(', ')}, `);
       this.error({
         message: 'Quota exceeded error with append of a single segment of content',
-        excludeUntil: Infinity
+        excludeUntil: Infinity,
+        errorType: videojs.Errors.SegmentExceedsBufferQuota
```
dzianis-dashkevich, line 2281: `SegmentExceedsSourceBufferQuota`

dzianis-dashkevich: I noticed that we always use the following pattern:

      this.error(...);
      this.trigger('error');
      
      maybe we should move trigger to error method?

wseymour15: I think the only issue for moving all of this into the error method would be that occasionally `error` is triggered, and other times `appendError` is triggered. We could create a separate `this.appendError` function that calls both `this.error(..)` and `this.trigger('appendError')` and clean this up a bit.

dzianis-dashkevich (approved):

dzianis-dashkevich:
I think videojs as peer dependency should be updated to a specific version since we are using `videojs.Error`

wseymour15:
> I think videojs as peer dependency should be updated to a specific version since we are using `videojs.Error`

Will this be problematic? I feel like the peer dependencies are set up that way so we can depend on both versions of video.js, for whatever reason. I did not change package.json for now, but I did update the peer dependency in the lock file

src/media-segment-request.js
```diff
@@ -159,11 +160,16 @@ const parseInitSegment = (segment, callback) => {
...
+    const mediaType = type || 'unknown';
 
     return callback({
       internal: true,
-      message: `Found unsupported ${type || 'unknown'} container for initialization segment at URL: ${uri}`,
-      code: REQUEST_ERRORS.FAILURE
+      message: `Found unsupported ${mediaType} container for initialization segment at URL: ${uri}`,
+      code: REQUEST_ERRORS.FAILURE,
+      metadata: {
+        errorType: videojs.UnsupportedMediaInitialization,
```
dzianis-dashkevich, line 170: `videojs.Error`

PR #1504, opened 2024-04-12
fix: lockfile integrity

adrums86:
## Description
`npm install` 

## Specific Changes proposed
ran install and the lockfile needed an update

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1868,13 +1868,13 @@
     "JSV": {
       "version": "4.0.2",
       "resolved": "https://registry.npmjs.org/JSV/-/JSV-4.0.2.tgz",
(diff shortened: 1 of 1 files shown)


wseymour15:
Isn't `SHA512` the more secure algorithm? I wonder why these dependencies are changing to a less secure hash.

gkatsev:
I assume it's a npm version mismatch between the one that was last used to generate, and then the one used here. We've historically only used the npm version that ships with the node version specified in the nvmrc. If a newer version of npm was used, everyone else should update to use that version of npm because otherwise there'll be lots of thrashing of the package-lock file.

adrums86:
Yeah this was the result of an `npm install` with node v14.21.3 I noticed during a previous release, the nvmrc has `14` as well 🤷‍♂️ . Closing for now, we can sort this out later.

PR #1505, opened 2024-04-22
fix: use paren media sequence sync for audio and vtt, since they are opt-in features and can be enabled after main init.

dzianis-dashkevich:
## Description
`audio` and `vtt` segment loaders are opt-in features and can be enabled after the `main` start. (eg: via UI)
We should align the media sequence start time with the `main` for them.

## Specific Changes proposed
Use `parent` for `vtt` and `audio` media sequence storages.
`parent` should be `main`.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -5,7 +5,7 @@
 import {sumDurations, getPartsAndSegments} from './playlist';
 import videojs from 'video.js';
 import logger from './util/logger';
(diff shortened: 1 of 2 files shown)


adrums86 (approved):
makes sense. LGTM!

misteroneill (approved):

PR #1525, opened 2024-07-10
fix: fastQualitySwitch stability

adrums86:
## Description
Currently, our fastQualitySwitch triggers a full video/main buffer clear twice. This is because we clear the buffer first with a `resetEverything` call on the mainSegmentLoader, then once that removal is complete we call `setCurrentTime` which calls `resetEverything` on all active segment loaders, clearing all the other buffers as well. This is causing some instability, specifically when a quality change occurs at or near a discontinuity.

## Specific Changes proposed
Instead of calling `setCurrentTime` after a fastQualitySwitch, lets just call `load` on the mainSegmentLoader.
Note: There were 2 browser bugs that seemed to require a `setCurrentTime` call according to the comments, but after checking both URLs, those bugs seem to no longer be relevant, so we should be safe removing this part and only calling load.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -1059,12 +1059,7 @@ export class PlaylistController extends videojs.EventTarget {
 
   runFastQualitySwitch_() {
     this.waitingForFastQualityPlaylistReceived_ = false;
-    // Delete all buffered data to allow an immediate quality switch, then seek to give
-    // the browser a kick to remove any cached frames from the previous rendtion (.04 seconds
-    // ahead was roughly the minimum that will accomplish this across a variety of content
-    // in IE and Edge, but seeking in place is sufficient on all other browsers)
-    // Edge/IE bug: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14600375/
-    // Chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=651904
+    // Delete all buffered data to allow an immediate quality switch.
     this.mainSegmentLoader_.pause();
     this.mainSegmentLoader_.resetEverything(() => {
       this.mainSegmentLoader_.load();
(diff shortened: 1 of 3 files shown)


src/playlist-controller.js
```diff
@@ -1059,15 +1059,10 @@ export class PlaylistController extends videojs.EventTarget {
...
-    // the browser a kick to remove any cached frames from the previous rendtion (.04 seconds
-    // ahead was roughly the minimum that will accomplish this across a variety of content
-    // in IE and Edge, but seeking in place is sufficient on all other browsers)
-    // Edge/IE bug: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14600375/
-    // Chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=651904
+    // Delete all buffered data to allow an immediate quality switch.
     this.mainSegmentLoader_.pause();
     this.mainSegmentLoader_.resetEverything(() => {
-      this.tech_.setCurrentTime(this.tech_.currentTime());
+      this.mainSegmentLoader_.load();
```
dzianis-dashkevich, line 1065: Do we want to load other loaders here?

dzianis-dashkevich (approved):

src/playlist-controller.js
adrums86: I don't think it's necessary, the only reason `load` was previously called on all the loaders was due to `setCurrentTime` calling `pause` then `resetEverything on _all_ the loaders, including the `mainSegmentLoader` a second time. https://github.com/videojs/http-streaming/blob/0b4da7ce70b49f0287600a0ef2f1468ec79e5e34/src/playlist-controller.js#L1535

wseymour15 (approved):

PR #1530, opened 2024-08-06
fix: audio segment on incorrect timeline

wseymour15:
## Description

Currently there is an edge case in DASH where certain points of audio segments and video segments do not always exist on the same timeline.

Example: We seek to `100` seconds into the video. There is an audio segment that is from 98-100.25 seconds, and the video segment is from 100-109.95. The audio segment is on a timeline from 90-99.99, and the video segment is on a timeline from 110-119.99. We fall into a loop in this scenario because VHS catches that we have bad timelines, but we do not do anything to fix it in this scenario.

This change allows us to check for this specific scenario. When we attempt to load an audio segment that is on a prior timeline to the video segment, we will now force the player forward to just past the faulty audio segment. This ensures that the audio and video segments will now be on the same timeline.

## Specific Changes proposed

- Moves the `fixBadTimeline` logic outside of the `hasEnoughInfoToLoad_` and `hasEnoughInfoToAppend_` functions so we do not have unexpected actions on a function that should just return a boolean.
- Create an event for when the audio segment is on a prior timeline to what the main video segment is on.
- Update the pending audio timeline change, and set the `currentTime` to just after the end of the faulty audio segment.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -929,6 +929,31 @@ export class PlaylistController extends videojs.EventTarget {
       this.onEndOfStream();
     });
 
+    // In DASH, there is the possibility of the video segment and the audio segment
+    // at a current time to be on different timelines. When this occurs, the player
+    // forwards playback to a point where these two segment types are back on the same
+    // timeline. This time will be just after the end of the audio segment that is on
+    // a previous timeline.
+    this.timelineChangeController_.on('audioTimelineBehind', () => {
+      const mainPendingTimelineChange = this.timelineChangeController_.pendingTimelineChange({ type: 'main' });
+
+      // Adjust the pending audio timeline to match the main timeline
+      this.timelineChangeController_.pendingTimelineChange({ to: mainPendingTimelineChange.to, from: mainPendingTimelineChange.from, type: 'audio' });
+
+      const segmentInfo = this.audioSegmentLoader_.pendingSegment_;
+
+      if (!segmentInfo || !segmentInfo.segment || !segmentInfo.segment.syncInfo) {
+        return;
+      }
+
+      // Update the current time to just after the faulty audio segment.
+      // This moves playback to a spot where both audio and video segments
+      // are on the same timeline.
+      const newTime = segmentInfo.segment.syncInfo.end + 0.01;
+
+      this.tech_.setCurrentTime(newTime);
+    });
+
     this.mainSegmentLoader_.on('earlyabort', (event) => {
       // never try to early abort with the new ABR algorithm
       if (this.bufferBasedABR) {
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -934,25 +934,27 @@ export class PlaylistController extends videojs.EventTarget {
     // forwards playback to a point where these two segment types are back on the same
     // timeline. This time will be just after the end of the audio segment that is on
     // a previous timeline.
-    this.timelineChangeController_.on('audioTimelineBehind', () => {
-      const mainPendingTimelineChange = this.timelineChangeController_.pendingTimelineChange({ type: 'main' });
+    if (this.sourceType_ === 'dash') {
+      this.timelineChangeController_.on('audioTimelineBehind', () => {
+        const mainPendingTimelineChange = this.timelineChangeController_.pendingTimelineChange({ type: 'main' });
 
-      // Adjust the pending audio timeline to match the main timeline
-      this.timelineChangeController_.pendingTimelineChange({ to: mainPendingTimelineChange.to, from: mainPendingTimelineChange.from, type: 'audio' });
+        // Adjust the pending audio timeline to match the main timeline
+        this.timelineChangeController_.pendingTimelineChange({ to: mainPendingTimelineChange.to, from: mainPendingTimelineChange.from, type: 'audio' });
 
-      const segmentInfo = this.audioSegmentLoader_.pendingSegment_;
+        const segmentInfo = this.audioSegmentLoader_.pendingSegment_;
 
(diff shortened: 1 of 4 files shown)


src/segment-loader.js
```diff
@@ -413,6 +413,11 @@ export const shouldFixBadTimelineChanges = (timelineChangeController) => {
   const pendingAudioTimelineChange = timelineChangeController.pendingTimelineChange({ type: 'audio' });
   const pendingMainTimelineChange = timelineChangeController.pendingTimelineChange({ type: 'main' });
   const hasPendingTimelineChanges = pendingAudioTimelineChange && pendingMainTimelineChange;
+
+  if (hasPendingTimelineChanges && pendingAudioTimelineChange.to < pendingMainTimelineChange.to) {
+    timelineChangeController.trigger('audioTimelineBehind');
```
dzianis-dashkevich, line 418: Should it be inside `should fix bad timeline changes`?
this looks like a side effect, and should be moved outside

src/playlist-controller.js
```diff
@@ -929,6 +929,31 @@ export class PlaylistController extends videojs.EventTarget {
       this.onEndOfStream();
     });
 
+    // In DASH, there is the possibility of the video segment and the audio segment
+    // at a current time to be on different timelines. When this occurs, the player
+    // forwards playback to a point where these two segment types are back on the same
+    // timeline. This time will be just after the end of the audio segment that is on
+    // a previous timeline.
+    this.timelineChangeController_.on('audioTimelineBehind', () => {
```
dzianis-dashkevich, line 937: sounds like we can listen to this event only when dash

src/playlist-controller.js
```diff
@@ -929,6 +929,31 @@ export class PlaylistController extends videojs.EventTarget {
...
+      if (!segmentInfo || !segmentInfo.segment || !segmentInfo.segment.syncInfo) {
+        return;
+      }
+
+      // Update the current time to just after the faulty audio segment.
+      // This moves playback to a spot where both audio and video segments
+      // are on the same timeline.
+      const newTime = segmentInfo.segment.syncInfo.end + 0.01;
+
+      this.tech_.setCurrentTime(newTime);
```
dzianis-dashkevich, line 954: We buffer segments ahead of time. This will jump ahead of what a user sees on the screen at the moment of the jump. Is that expected?

src/playlist-controller.js
wseymour15: I guess leaving that comment was to point out that this scenario should only happen when DASH.. but I am not sure if it hurts to handle it in any case.

I can probably make that more clear

src/playlist-controller.js
wseymour15: In this scenario, I believe what the user will see at the moment of the jump is a loading screen. This is because the segments that _should_ be playing are not loaded to due this timeline inconsistency.

In this case, I think the jump is OK, but let me know if I am off there.

src/playlist-controller.js
dzianis-dashkevich: I mean, if this code is explicitly intended to handle a dash-specific case, why should we run it for HLS (and potentially impact HLS playback)

I think we should guard it with if statement

src/segment-loader.js
wseymour15: I agree that this can be moved elsewhere.. but I think there are a few different places it could be placed.

Do you think it would make sense to just check after an [audio timeline change](https://github.com/videojs/http-streaming/blob/main/src/segment-loader.js#L723)?

src/playlist-controller.js
dzianis-dashkevich: > This is because the segments that should be playing are not loaded to due this timeline inconsistency.

Yeah, that is what I am trying to say; the segment you are loading is around `bufferedEnd`, while what users see is around `currentTime`.  We do not load segments that should be playing right now, in most cases, we are loading segments future segments (ahead of time).

we load segments around currentTime only when we are switching rendition.

dzianis-dashkevich: I guess we should probably check whether currentTime === bufferedEnd, and we are really stuck so we can jump ahead. I suspect we should already have this logic in the playback watcher.

src/segment-loader.js
dzianis-dashkevich: probably yes

src/segment-loader.js
```diff
@@ -423,15 +423,71 @@ export const shouldFixBadTimelineChanges = (timelineChangeController) => {
...
+const isAudioTimelineBehind = (segmentLoader) => {
+  const pendingAudioTimelineChange = segmentLoader.timelineChangeController_.pendingTimelineChange({ type: 'audio' });
+  const pendingMainTimelineChange = segmentLoader.timelineChangeController_.pendingTimelineChange({ type: 'main' });
+  const hasPendingTimelineChanges = pendingAudioTimelineChange && pendingMainTimelineChange;
+
+  if (hasPendingTimelineChanges && pendingAudioTimelineChange.to < pendingMainTimelineChange.to) {
+    return true;
+  }
+
+  return false;
```
dzianis-dashkevich, line 457: you can simply:
  `return hasPendingTimelineChanges && pendingAudioTimelineChange.to < pendingMainTimelineChange.to;`

adrums86 (approved):
I really like how this ended up, great suggestions @dzianis-dashkevich and fantastic execution @wseymour15 , especially consolidating the logic around `checkAndFixTimelines`, that leaves us a cleaner way to add other "corrections" for edge cases here if necessary.

dzianis-dashkevich (approved):

PR #1542, opened 2024-10-02
feat: Add support for ManagedMediaSource 'startstreaming' and 'endstream' event handling

alex-barstow:
## Description
This PR adds onto the experimental support for `ManagedMediaSource`. When the user agent determines that a ManagedMediaSource has enough data buffered to ensure uninterrupted playback, or does not have enough to ensure it, it will fire a `endstreaming` or `startstreaming` event, indicating that the player should stop loading new segments or start loading new segments, respectively.

## Specific Changes proposed
Add 2 new event handlers for `startstreaming` and `endstreaming`. The former will call `playlistController.load()`, which was a preexisting method that calls `load()` on all active segment loaders. The latter calls `playlistController.pause()`, which is a new method that calls `pause()` on all active segment loaders.

## Requirements Checklist
  - [ ] Docs/guides updated

## References
https://www.w3.org/TR/media-source-2/#dom-managedmediasource-streaming
https://github.com/w3c/media-source/issues/320

### DIFF
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -192,7 +192,6 @@ export class PlaylistController extends videojs.EventTarget {
     this.playlistExclusionDuration = playlistExclusionDuration;
     this.maxPlaylistRetries = maxPlaylistRetries;
     this.enableLowInitialPlaylist = enableLowInitialPlaylist;
-    this.hasManagedMediaSource_ = false;
 
     if (this.useCueTags_) {
       this.cueTagsTrack_ = this.tech_.addTextTrack(
@@ -216,7 +215,6 @@ export class PlaylistController extends videojs.EventTarget {
       // Airplay source not yet implemented. Remote playback must be disabled.
       this.tech_.el_.disableRemotePlayback = true;
       this.mediaSource = new window.ManagedMediaSource();
-      this.hasManagedMediaSource_ = true;
 
       videojs.log('Using ManagedMediaSource');
     } else if (window.MediaSource) {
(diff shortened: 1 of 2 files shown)


src/playlist-controller.js
```diff
@@ -192,6 +192,7 @@ export class PlaylistController extends videojs.EventTarget {
     this.playlistExclusionDuration = playlistExclusionDuration;
     this.maxPlaylistRetries = maxPlaylistRetries;
     this.enableLowInitialPlaylist = enableLowInitialPlaylist;
+    this.hasManagedMediaSource_ = false;
```
wseymour15, line 195: nit: To save a few lines of code, you could change this to:
```
this.hasManagedMediaSource_ = experimentalUseMMS && window.ManagedMediaSource;
```

src/playlist-controller.js
```diff
@@ -215,6 +216,8 @@ export class PlaylistController extends videojs.EventTarget {
       // Airplay source not yet implemented. Remote playback must be disabled.
```
wseymour15, line 216: You can change the above if statement to:
```
if (this.hasManagedMediaSource_) {
```

src/playlist-controller.js
```diff
@@ -232,6 +235,14 @@ export class PlaylistController extends videojs.EventTarget {
     // we don't have to handle sourceclose since dispose will handle termination of
     // everything, and the MediaSource should not be detached without a proper disposal
 
+    if (this.hasManagedMediaSource_) {
+      this.handleStartStreaming_ = this.handleStartStreaming_.bind(this);
```
wseymour15, line 239: You can add the content of this if statement to the above pre-existing if statement. All nits, but saves a few lines

src/playlist-controller.js
alex-barstow: The intent of the flag was to signal that the managedMediaSource has been created and currently exists. If we set it here we should rename the flag to `useManagedMediaSource_` or `createManagedMediaSource_`, as it hasn't been created yet.

src/playlist-controller.js
alex-barstow: You mean to the `if (experimentalUseMMS && window.ManagedMediaSource) {...}` block? It is more concise but  I opted against it because to my eye it made the code slightly less organized by putting the new `addEventListener` calls inside the block where the MediaSource is set rather than in the subsequent step/section where the other MediaSource event handlers are added. Minor issue, I'm happy to change it

wseymour15: Yeah, all my comments were basically to save some lines of code by moving this code to where you set the `hasManagedMediaSource` flag and making the flag unnecessary, but fine with me if it keeps it cleaner in your eyes. 👍

wseymour15 (approved):
Good work! Looks good to me

src/playlist-controller.js
```diff
@@ -232,6 +235,14 @@ export class PlaylistController extends videojs.EventTarget {
...
+      this.handleEndStreaming_ = this.handleEndStreaming_.bind(this);
+
+      this.mediaSource.addEventListener('startstreaming', this.handleStartStreaming_);
+      this.mediaSource.addEventListener('endstreaming', this.handleEndStreaming_);
```
dzianis-dashkevich, line 243: I guess we can remove the `hasManagedMediaSource_ ` flag and simply add these listeners alongside other listeners; we should not care about the type of the media source

src/playlist-controller.js
```diff
@@ -1223,6 +1251,26 @@ export class PlaylistController extends videojs.EventTarget {
...
+  /**
+   * Handle the endstreaming event on the ManagedMediaSource.
+   * This event indicates that segment loading should be paused.
+   *
+   * @private
+   */
+  handleEndStreaming_() {
+    this.pause();
+  }
+
```
dzianis-dashkevich, line 1273: Do we really need these wrappers for load and pause?

alex-barstow: We don't. If the logic ever becomes more sophisticated that just load/pause in the future, then we would need them. We can cross that bridge if we get there, I'm fine removing them now

dzianis-dashkevich: yeah, at this point seems redundant

dzianis-dashkevich (approved):

PR #1543, opened 2024-10-08
feat: Add Airplay support when overriding native HLS in Safari/iOS

alex-barstow:
## Description
**Note:** This depends on https://github.com/videojs/video.js/pull/8886. Before we merge this, we will need a new Video.js release and then have to bump the peer dependency version in this package.

This PR adds Airplay support when overriding native HLS in Safari and/or iOS by using two `source` elements, one for the MediaSource URL object and one for the Airplay-compatible manifest URL.

From the Webkit guide on [Airplay and MSE](https://webkit.org/blog/15036/how-to-use-media-source-extensions-with-airplay): 
> When Safari detects that an alternative source is available in addition to the MediaSource URL object, it will display the familiar AirPlay icon to the video player control. Should the user select AirPlay, it will switch over from MSE to the AirPlay-compatible URL.

## Specific Changes proposed
If we are playing HLS with MSE in Safari or iOS, set the source by adding two `<source>` elements-- one for the Media Source blob and one for the HLS manifest that will be used for Airplay. Preserve the existing behavior in all other cases.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -1079,7 +1079,14 @@ class VhsHandler extends Component {
 
     this.mediaSourceUrl_ = window.URL.createObjectURL(this.playlistController_.mediaSource);
 
-    this.tech_.src(this.mediaSourceUrl_);
+    // If we are playing HLS with MSE in Safari, add source elements for both the blob and manifest URLs.
+    // The latter will enable Airplay playback on receiver devices.
+    if ((videojs.browser.IS_ANY_SAFARI || videojs.browser.IS_IOS) && this.options_.overrideNative && this.options_.sourceType === 'hls') {
+      this.tech_.addSourceElement(this.mediaSourceUrl_);
+      this.tech_.addSourceElement(this.source_.src);
+    } else {
+      this.tech_.src(this.mediaSourceUrl_);
+    }
   }
 
   createKeySessions_() {
--- a/src/videojs-http-streaming.js
+++ b/src/videojs-http-streaming.js
@@ -1081,7 +1081,12 @@ class VhsHandler extends Component {
 
     // If we are playing HLS with MSE in Safari, add source elements for both the blob and manifest URLs.
     // The latter will enable Airplay playback on receiver devices.
-    if ((videojs.browser.IS_ANY_SAFARI || videojs.browser.IS_IOS) && this.options_.overrideNative && this.options_.sourceType === 'hls') {
+    if ((
+      videojs.browser.IS_ANY_SAFARI || videojs.browser.IS_IOS) &&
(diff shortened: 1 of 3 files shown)


src/videojs-http-streaming.js
```diff
@@ -1079,7 +1079,14 @@ class VhsHandler extends Component {
 
     this.mediaSourceUrl_ = window.URL.createObjectURL(this.playlistController_.mediaSource);
 
-    this.tech_.src(this.mediaSourceUrl_);
+    // If we are playing HLS with MSE in Safari, add source elements for both the blob and manifest URLs.
+    // The latter will enable Airplay playback on receiver devices.
+    if ((videojs.browser.IS_ANY_SAFARI || videojs.browser.IS_IOS) && this.options_.overrideNative && this.options_.sourceType === 'hls') {
+      this.tech_.addSourceElement(this.mediaSourceUrl_);
```
wseymour15, line 1085: Is it possible that `addSourceElement` will not exist on some techs? Do we want to check for it's existence first?

src/videojs-http-streaming.js
```diff
@@ -1079,7 +1079,14 @@ class VhsHandler extends Component {
...
+      this.tech_.addSourceElement(this.source_.src);
+    } else {
+      this.tech_.src(this.mediaSourceUrl_);
```
wseymour15, line 1088: Is there any difference between `this.tech_.addSourceElement(this.mediaSourceUrl_)` and `this.tech_.src(this.mediaSourceUrl_)`?

Could we just do one in every case, and conditionally add the manifest URL source?

alex-barstow: `this.tech_.src()` sets the source by setting the `src` attribute on the `<video>` element, while `this.tech_.addSourceElement()` adds the source as a separate `<source>` element that is a child of the `<video>` element. We definitely want both the Media Source and Airplay-compatible sources to be `<source>` elements when using MSE in Safari, per the Webkit doc, and my inclination was to preserve the existing behavior in all other cases (setting the `<video>`'s `src` attribute)

src/videojs-http-streaming.js
alex-barstow: MSE is an extension of HTML5 so we should be able to safely assume that all `tech` references are for the HTML5 tech.

gkatsev: Though unlikely, doing a check here adds safety in case someone is using a later version of vhs with the current version of videojs.

wseymour15 (approved):
+1 once the changes from video.js are pulled in

dzianis-dashkevich (approved):

PR #1545, opened 2024-10-11
feat: parse mp4 webvtt segments

adrums86:
### Feature
This pull request adds support for parsing mp4 encapsulated webvtt segments by modifying the existing `vtt-segment-loader` and associated functions to interface with the transmuxer to parse fmp4 webvtt initialization and content segments.

## Specific Changes proposed
- Update mux.js to include the WebVttParser for use in the transmuxer worker.
- Modify the `vtt-segment-loader` to be able to handle fragmented mp4 segments.
- Parse `wvtt` initialization segments and content segments with the help of the transmuxer.
- Parse the cues provided by the transmuxer into `VTTCue` objects and add them to the subtitles text tracks.

## Demo
See the deployed demo page [here](https://deploy-preview-1545--videojs-http-streaming.netlify.app/?debug=true&autoplay=false&muted=false&fluid=false&minified=false&sync-workers=false&liveui=true&llhls=true&url=https%3A%2F%2Flivesim2.dashif.org%2Flivesim2%2Ftimesubswvtt_en%2Ftestpic_2s%2FManifest.mpd&type=application%2Fdash%2Bxml&keysystems=&buffer-water=false&exact-manifest-timings=false&pixel-diff-selector=false&network-info=false&dts-offset=false&override-native=true&use-mms=true&preload=auto&mirror-source=true&forced-subtitles=false)
with the simulated live DASH manifest from DASH-IF with mp4 encapsulated webvtt: https://livesim2.dashif.org/livesim2/timesubswvtt_en/testpic_2s/Manifest.mpd

## Requirements Checklist
  - [ ] Docs/guides updated


src/vtt-segment-loader.js
```diff
@@ -282,6 +277,11 @@ export default class VTTSegmentLoader extends SegmentLoader {
     }
 
     const segmentInfo = this.pendingSegment_;
+    const isMp4WebVttSegment = result.mp4VttCues && result.mp4VttCues.length;
```
alex-barstow, line 280: minor but would `isMp4WebVttSegmentWithCues` be a more accurate name?

src/vtt-segment-loader.js
```diff
@@ -380,14 +382,47 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
+   */
+  parseMp4VttCues_(segmentInfo) {
+    const timestampOffset = this.sourceUpdater_.videoTimestampOffset() === null ?
+      this.sourceUpdater_.audioTimestampOffset() :
+      this.sourceUpdater_.videoTimestampOffset();
+
+    segmentInfo.mp4VttCues.forEach((cue) => {
+      const start = cue.start + timestampOffset;
+      const end = cue.end + timestampOffset;
+      const vttCue = new window.VTTCue(start, end, cue.cueText);
```
alex-barstow, line 411: is `cue.cueText` guaranteed to exist/be a string?

src/vtt-segment-loader.js
```diff
@@ -380,14 +382,47 @@ export default class VTTSegmentLoader extends SegmentLoader {
     this.handleAppendsDone_();
   }
 
-  handleData_() {
-    // noop as we shouldn't be getting video/audio data captions
-    // that we do not support here.
+  handleData_(simpleSegment, result) {
+    const isFmp4VttSegment = simpleSegment.type === 'vtt' && result.type === 'text';
```
wseymour15, line 386: Could errors possibly be thrown here if `simpleSegment` or `result` are undefined? It might be worth to have a check here

src/vtt-segment-loader.js
```diff
@@ -334,11 +334,13 @@ export default class VTTSegmentLoader extends SegmentLoader {
       return;
     }
 
-    this.updateTimeMapping_(
-      segmentInfo,
-      this.syncController_.timelines[segmentInfo.timeline],
-      this.playlist_
-    );
+    if (!isMp4WebVttSegment) {
```
wseymour15, line 337: I'm sure there is a good reason, but why do we not want to update the time mapping in this case?

src/transmuxer-worker.js
```diff
@@ -207,6 +208,41 @@ class MessageHandlers {
...
+   * @param {Uint8Array} data mp4 boxed WebVTT init segment data
+   */
+  initMp4WebVttParser(data) {
+    if (!this.webVttParser) {
+      this.webVttParser = new WebVttParser();
+    }
+    const segment = new Uint8Array(data.data, data.byteOffset, data.byteLength);
+
+    // set the timescale for the parser.
+    this.webVttParser.init(segment);
```
wseymour15, line 223: I'm note exactly sure how the `init` function works on the parser, but is it possible/valid to call `init` when the parser already exists and `init` has already been called?

src/transmuxer-worker.js
```diff
@@ -207,6 +208,41 @@ class MessageHandlers {
...
+  }
+
+  /**
+   * Parse an mp4 encapsulated WebVTT segment and return an array of cues.
+   *
+   * @param {Uint8Array} data a text/webvtt segment
+   * @return {Object[]} an array of parsed cue objects
+   */
+  getMp4WebVttText(data) {
+    if (!this.webVttParser) {
```
wseymour15, line 233: Is this check necessary? Mainly asking because I am unclear if the order of the `postMessage` calls will guarantee the parser exists

src/media-segment-request.js
```diff
@@ -164,6 +166,43 @@ const handleKeyResponse = (segment, objects, finishProcessingFn, triggerSegmentE
...
 };
 
+/**
+ * Processes an mp4 init segment depending on the codec through the transmuxer.
+ *
+ * @param {Object} segment init segment to process
+ * @param {string} codec the codec of the text segments
+ */
+const initMp4Text = (segment, codec) => {
+  if (codec === WEB_VTT_CODEC) {
```
wseymour15, line 176: nit: For both these functions, I would move the codec check on the call of the function, so it is not even called if it is not the correct codec.

adrums86: I think the idea for having these functions as generic `mp4Text` init/segments is to leave a low resistance path for other text codecs to be initialized/parsed, such as ttml. So we could do a `else if (codec === TTML_CODEC)` later. I'm not totally sold on that being necessary though, if you think it makes more sense to codec check outside we can remove these `mp4Text` functions altogether.

src/transmuxer-worker.js
adrums86: So, the init function for the `WebVttParser` only passes the segment data to extract a timescale. So it actually is valid to call this multiple times, especially if the timescale changes over a discontinuity. I'll add a bit to the comment here that init is ok to call repeatedly. https://github.com/videojs/mux.js/blob/e1b4d161f51782e4712d020b66d5087bd6febb3d/lib/mp4/webvtt-parser.js#L18

src/transmuxer-worker.js
adrums86: Checking here is only precautionary. Ideally any series of fmp4 text segments should be preceded by an init segment which would initialize the parser and set the timescale. I'm only inclined to keep it here to improve resiliency, maybe logging here would be helpful?

src/vtt-segment-loader.js
adrums86: The `updateTimeMapping_` function being called here is specific to MPEGTS/vtt time mapping. Since these are mp4 segments we use the parent segment loader timing.

src/vtt-segment-loader.js
adrums86: Good call, will make this check a bit more robust.

src/vtt-segment-loader.js
adrums86: The sample duration and `cueText` must be defined coming from the transmuxer, so I think we're guaranteed these values will be correct with regards to _type_ at this point. https://github.com/videojs/mux.js/blob/e1b4d161f51782e4712d020b66d5087bd6febb3d/lib/mp4/webvtt-parser.js#L107

src/transmuxer-worker.js
wseymour15: Yeah, a warning that explaining that the VTT parser is being init in that function could be useful

src/media-segment-request.js
wseymour15: I think it's fine as is if the expectation is that other checks will happen in this function

dzianis-dashkevich (approved):

src/transmuxer-worker.js
adrums86: Actually, looking at this again I'm not sure we want a log here, since this is in the transmuxer worker we wouldn't be able to easily send this through the logger. I'll add a comment instead as clarification. The parser will fallback to the default 90k timescale if not initialized by an init segment anyway, so nothing would break, but the cue timing could potentially be off.

wseymour15 (approved):

PR #1550, opened 2024-10-30
fix: Ensure native segment metadata track has mode hidden

mister-ben:
## Description
When using native text tracks, the segment metadata track was being created with mode `disabled`. This meant it wouldn't trigger `cuechange` events, but also Chrome (at least) does not add cues to disabled tracks. This change makes the track start out `hidden`.

## Specific Changes proposed
- Set mode to `hidden`
- Adds a simple test for track mode. Testing whether the cues are added isn't simple as `addSegmentMetadataCue_` is made noop in the tests.
-  Adds native text tracks as an option to the text page for manual verification.

## Requirements Checklist
  - [ ] Docs/guides updated


phloxic (commented):
typo in commit message: `s/mdoe hidden/mode hidden/`

dzianis-dashkevich (approved):

src/playlist-controller.js
```diff
@@ -254,6 +254,8 @@ export class PlaylistController extends videojs.EventTarget {
       label: 'segment-metadata'
     }, false).track;
 
+    this.segmentMetadataTrack_.mode = 'hidden';
```
dzianis-dashkevich, line 257: I thought it is possible to pass `mode` as an option to add remote text track

PR #1565, opened 2025-02-04
Audio only exclude

dzianis-dashkevich:
## Description
We might encounter situations when we have a main playlist with audio-only and video-only or muxed renditions. VHS does not support switching between audio-only and video-only or muxed, so we want to exclude audio-only renditions if we have renditions with video.

## Specific Changes proposed
- Exclude audio-only renditions if we detected renditions with video.
- Enable use of networkInformation API by default.

## Requirements Checklist
  - [ ] Docs/guides updated


adrums86 (approved):
Good catch. We definitely should not be including demuxed audio with video playlists.

PR #1585, opened 2025-07-23
chore: get ci passing and update node

Essk:
It seems ci test runs have been failing for a while and after a bit of investigation there seem to be two main culprits:

1) We need to pin Firefox to v64, same as for [VJS](https://github.com/videojs/video.js/pull/5793)
2) Looks like there may be a regression in playback for the source used for "DRM Dash" playback test. This asset plays in the DASH reference player but not the VHS demo.
  - We recently improved support for multi-DRM in contrib-eme, so I've bumped to the latest version but sadly this doesn't seem to fix things for this asset. Further investigation will be needed.
  
For good measure, I've also updated Node to v22.

### DIFF
--- a/test/playback.test.js
+++ b/test/playback.test.js
@@ -250,32 +250,6 @@ QUnit.test('Live DASH', function(assert) {
   player.play();
 });
 
-QUnit.test('Multiperiod dash works and can end', function(assert) {
-  const done = assert.async();
-
-  assert.expect(2);
-  const player = this.player;
-
-  playFor(player, 2, function() {
-    assert.ok(true, 'played for at least two seconds');
-    assert.equal(player.error(), null, 'has no player errors');
-
-    player.one('ended', () => {
-      assert.ok(true, 'triggered ended event');
-      done();
-    });
-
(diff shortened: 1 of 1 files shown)


scripts/karma.conf.js
```diff
@@ -27,6 +27,11 @@ module.exports = function(config) {
       if (CI_TEST_TYPE === 'coverage') {
         defaults = {};
       }
+      // pin Browserstack Firefox version to 64
+      if (defaults.bsFirefox) {
+        // eslint-disable-next-line camelcase
+        defaults.bsFirefox.browser_version = '64.0';
```
gkatsev: this firefox is pretty old at this point (current is 147), even older than firefox ESR (their LTS release). Would be nice to update, but not a blocker for this PR.

gkatsev (approved):

scripts/karma.conf.js
gkatsev: ah, I see the comment about firefox. Theoretically, a karma bump should fix it?

PR #1601, opened 2026-03-12
fix: Safari playback stall with overrideNative and experimentalUseMMS

### LINKED ISSUE #1600: Safari: media playback does not start when using overrideNative + experimentalUseMMS
Hello there!

**Description**

I encountered a similar issue to [this one](https://github.com/videojs/video.js/issues/9117) however I have not faced the `MEDIA_ERR_SRC_NOT_SUPPORTED` error

I noticed that when using Safari with the flags `overrideNative` and `experimentalUseMMS` set to true, video playback does not start.

```
      html5: {
        vhs: {
          overrideNative: true,
          experimentalUseMMS: true
        }
```

I traced the issue back to the following commit:
https://github.com/videojs/http-streaming/commit/c1d31867b913e404d7e9bb9cd777d0ec2fa55af8

Prior to this change everything works as expected.

**Test case**

This sample reproduces the issue when tested in Safari.

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8" />
  <title>Example: overrideNative error on Safari</title>

  <!-- Local Video.js assets -->
  <!-- <link href="../video.js/dist/video-js.css" rel="stylesheet" type="text/css" />
  <script src="../video.js/dist/video.js"></script> -->
  <script src="https://vjs.zencdn.net/8.23.4/video.js"></script>
  <link href="https://vjs.zencdn.net/8.23.4/video-js.css" rel="stylesheet" type="text/css" />
</head>
<body>

  <h1>Example: overrideNative error on Safari</h1>

  <video
    id="player"
    class="video-js vjs-default-skin vjs-big-play-centered"
    controls
    preload="auto"
    autoplay
    muted
    width="640"
    height="360"
    data-setup="{}"
  >
    <p class="vjs-no-js">
      To view this video please enable JavaScript, and consider upgrading to a
      web browser that
      <a href="https://videojs.com/html5-video-support/" target="_blank"
        >supports HTML5 video</a
      >.
    </p>
  </video>

  <script>
    const player = videojs('player', {
      html5: {
        vhs: {
          overrideNative: true,
          experimentalUseMMS: true
        },
      }
    });
    player.src({
      src: 'https://stream.mux.com/v69RSHhFelSm4701snP22dYz2jICy4E4FUyk02rW4gxRM.m3u8',
      type: 'application/x-mpegURL'
    });
    player.ready(function () {
      console.log('Video.js player ready');
    });
  </script>

</body>
</html>
``` 

**Possible fix**

After reviewing the code further, I noticed that playback breaks when the following line was introduced:
```javascript
this.segmentMetadataTrack_.mode = 'hidden';
```
Which was introduced in the mentioned commit

Changing the file `playlist-controller.js` with this
```javascript
    // Only disable segment metadata track in Safari/iOS when using overrideNative HLS
    // (same scenario where we add both blob and manifest source elements for AirPlay)
    if ((videojs.browser.IS_ANY_SAFARI || videojs.browser.IS_IOS) &&
      options.overrideNative &&
      options.experimentalUseMMS &&
      this.sourceType_ === 'hls') {
        this.segmentMetadataTrack_.mode = 'disabled';
    } else {
      this.segmentMetadataTrack_.mode = 'hidden';
    }
```
This change appears to fix the issue.

I will open a PR with the proposed fix so it can be reviewed.

**What version of Video.js are you using?**

8.23.4

**Video.js plugins used**

No response

**What browser(s) including version(s) does this occur with?**

Safari 26.3

**What OS(es) and version(s) does this occur with?**

MacOS Tahoe 26.3

Demenus:
## Description
Safari playback does not start when using overrideNative together with experimentalUseMMS.

This appears to be caused by setting the segment metadata track mode to hidden, which can stall playback in Safari when using Managed Media Source.

Disabling the metadata track in this specific scenario prevents Safari from processing it and allows playback to start normally.

Fixes #1600

This behavior seems specific to Safari when using MMS. Other browsers (Chrome, Firefox) do not appear to be affected by this issue.

## Specific Changes proposed

- Detect Safari/iOS when using `overrideNative` + `experimentalUseMMS`
- Set `segmentMetadataTrack_.mode = 'disabled'` instead of `hidden` in this specific scenario
- Keep the existing behavior (`hidden`) for all other browsers and configurations

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-controller.js
+++ b/src/playlist-controller.js
@@ -254,7 +254,13 @@ export class PlaylistController extends videojs.EventTarget {
       label: 'segment-metadata'
     }, false).track;
 
-    this.segmentMetadataTrack_.mode = 'hidden';
+    // Only disable segment metadata track in Safari/iOS when using overrideNative HLS
+    if ((videojs.browser.IS_ANY_SAFARI || videojs.browser.IS_IOS) &&
+      this.sourceType_ === 'hls') {
+      this.segmentMetadataTrack_.mode = 'disabled';
+    } else {
+      this.segmentMetadataTrack_.mode = 'hidden';
+    }
 
     this.decrypter_ = new Decrypter();
     this.sourceUpdater_ = new SourceUpdater(this.mediaSource);


src/playlist-controller.js
```diff
@@ -254,7 +254,15 @@ export class PlaylistController extends videojs.EventTarget {
       label: 'segment-metadata'
     }, false).track;
 
-    this.segmentMetadataTrack_.mode = 'hidden';
+    // Only disable segment metadata track in Safari/iOS when using overrideNative HLS
+    if ((videojs.browser.IS_ANY_SAFARI || videojs.browser.IS_IOS) &&
+      options.overrideNative &&
+      options.experimentalUseMMS &&
```
mister-ben:
```suggestion
```
I think these can be removed. If this code is reached, it's not native playback. And without the MMS check this seems to fix videojs/video.js#9117 too.

Demenus: You are right — at that point in the code we are no longer using native playback.

I have removed those flags from the condition.

Thank you for the fast review 👍🏻

hannseman:
Is there any way for me to help push this forward? Would love to see this issue fixed.

mister-ben:
I can no longer reproduce the issue in #1600 with the example there. I've updated to Mac OS/Safari 26.4 so perhaps there's been a change in Safari. @Demenus do you experience the same?

Unfortunately it was a false positive that this would also resolve videojs/video.js#9117. It was only appearing to do so as the version of video.js in this repo's package-lock.json is a bit outdated.

Demenus:
@mister-ben I can confirm the same behavior on my side, but it looks like this was actually a Safari regression that’s already been fixed.

Safari 26.3 → playback doesn’t start (reproducible)
Safari 26.4 → playback works fine without any code changes

Looking at the Safari 26.4 release notes, this might be related to:

“Fixed an issue where empty <track> elements prevented media from advancing its readyState and blocked play() calls.” (164125914)

https://developer.apple.com/documentation/safari-release-notes/safari-26_4-release-notes

Since VHS uses internal metadata tracks, this lines up pretty well with what we’re seeing.

So it seems the segmentMetadataTrack_.mode = 'hidden' change isn’t inherently broken, but rather exposed a Safari bug that’s now fixed.

The proposed workaround (forcing 'disabled' for Safari) might still make sense for older versions, but probably isn’t needed anymore for newer ones.

skrhlm:
experiencing the same issue here, i have a video that loads with poor quality/ keeps jumping between different resolutions in safari 26.3 and 26.4.
```
// default settings:
26.3  => ❌ poor quality, inconsistent resolutions picked.
26.4  => ❌ poor quality, inconsistent resolutions picked.

26.3 with html5.vhs.overrideNative = true => ❌ video does not start at all.
26.4 with html5.vhs.overrideNative = true =>  ✅  video works as expected, correct resolution picked.

File: src/playlist-controller.js [hunk 2]
```diff
@@ -254,7 +254,13 @@ export class PlaylistController extends videojs.EventTarget {
       label: 'segment-metadata'
     }, false).track;
 
-    this.segmentMetadataTrack_.mode = 'hidden';
+    // Only disable segment metadata track in Safari/iOS when using overrideNative HLS
+    if ((videojs.browser.IS_ANY_SAFARI || videojs.browser.IS_IOS) &&
```
skrhlm, line 258: I think the `IS_IOS` check here may be redundant, anything that falls under `IS_IOS` would also be matched by `IS_ANY_SAFARI`.

skrhlm:
Just confirming that this issue still persists in the new v10 beta.

PR #1607, opened 2026-06-17
fix(vtt): never transmux raw WebVTT as MPEG-TS

djonesey:
## Description

Fixes an intermittent WebVTT subtitle rendering failure where raw WebVTT subtitle segments can be misdetected as MPEG-TS and routed through the transmuxer. When this happens, the subtitle track loads but parses 0 cues, and the console logs:

> VIDEOJS: WARN: Error encountered when parsing cues: Malformed WebVTT signature.

The issue is content-dependent. Raw WebVTT text can coincidentally contain `0x47` (`G`) bytes exactly 188 bytes apart near the start of the segment. That byte pattern trips the fuzzy MPEG-TS container heuristic even though the file is valid WebVTT. The same player/config can therefore work for one subtitle segment and fail for another depending only on subtitle content and byte layout.

Reproducible test case:

https://output.jsbin.com/kahoqav

The repro serves a raw WebVTT subtitle segment whose cue text triggers the false-positive MPEG-TS detection. In the buggy path, `repro.vtt` is fetched successfully, the English subtitle track is `showing`, but the parser logs `Malformed WebVTT signature` and the track ends with 0 cues.

## Specific Changes proposed

- Add an early `segment.type === 'vtt'` guard in `src/media-segment-request.js` before container sniffing.
- Skip container detection and transmuxing for raw WebVTT subtitle segments.
- Preserve the existing fmp4 `wvtt` handling path.
- Add a regression test in `test/media-segment-request.test.js` that:
  - builds raw WebVTT bytes which `detectContainerForBytes()` misdetects as `ts`,
  - passes them through `mediaSegmentRequest()` with `segment.type === 'vtt'` and a truthy transmuxer,
  - asserts the transmuxer is not invoked,
  - asserts the transmux data path is not used,
  - asserts the original WebVTT bytes are preserved.
- Keep the existing fmp4 `wvtt` test as a regression guard for fmp4 text subtitles.
- fmp4 `wvtt` subtitles are unaffected by this change because they are handled earlier by the fmp4 media segment path.

## Requirements Checklist

  - [ ] Docs/guides updated
  - [ ] Example created 


djonesey:
Closing this. The cleaner fix belongs in `@videojs/vhs-utils`, by detecting `WEBVTT` before the TS heuristic runs. The defensive empty-buffer retry in `VTTSegmentLoader` also makes sense.

